PR #15, opened 2021-08-29
Support mTLS for gRPC channel

dmsolr:

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
@@ -135,6 +135,21 @@ public class Config {
          * Force open TLS for gRPC channel if true.
          */
         public static boolean FORCE_TLS = false;
+
+        /**
+         * SSL trusted ca file. If it exists, will enable TLS for gRPC channel.
+         */
+        public static String SSL_TRUSTED_CA_PATH = "ca" + Constants.PATH_SEPARATOR + "ca.crt";
+
+        /**
+         * Key cert chain file. If ssl_cert_chain and ssl_key exist, will enable mTLS for gRPC channel.
+         */
+        public static String SSL_CERT_CHAIN_PATH;
+
+        /**
+         * Private key file. If ssl_cert_chain and ssl_key exist, will enable mTLS for gRPC channel.
+         */
+        public static String SSL_KEY_PATH;
     }
 
     public static class OsInfo {
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
@@ -44,16 +44,17 @@ public class TLSChannelBuilder implements ChannelBuilder<NettyChannelBuilder> {
     public NettyChannelBuilder build(
         NettyChannelBuilder managedChannelBuilder) throws AgentPackageNotFoundException, IOException {
 
-        File caFile = new File(toAbsolutePath(Config.Agent.SSL_TRUSTED_CA_PATH));
-        if (Config.Agent.FORCE_TLS || caFile.isFile()) {
+        File caFile = new File(AgentPackagePath.getPath(), Config.Agent.SSL_TRUSTED_CA_PATH);
+        boolean isCAFileExist = caFile.exists() && caFile.isFile();
+        if (Config.Agent.FORCE_TLS || isCAFileExist) {
             SslContextBuilder builder = GrpcSslContexts.forClient();
 
-            if (caFile.isFile()) {
+            if (isCAFileExist) {
                 String certPath = Config.Agent.SSL_CERT_CHAIN_PATH;
                 String keyPath = Config.Agent.SSL_KEY_PATH;
                 if (StringUtil.isNotBlank(certPath) && StringUtil.isNotBlank(keyPath)) {
-                    File keyFile = new File(toAbsolutePath(keyPath));
-                    File certFile = new File(toAbsolutePath(certPath));
+                    File keyFile = new File(AgentPackagePath.getPath(), keyPath);
+                    File certFile = new File(AgentPackagePath.getPath(), certPath);
 
                     if (certFile.isFile() && keyFile.isFile()) {
                         try (InputStream cert = new FileInputStream(certFile);
@@ -70,14 +71,6 @@ public class TLSChannelBuilder implements ChannelBuilder<NettyChannelBuilder> {
             managedChannelBuilder.negotiationType(NegotiationType.TLS).sslContext(builder.build());
         }
         return managedChannelBuilder;
-    }
-
-    private static String toAbsolutePath(final String path) throws AgentPackageNotFoundException {
-        if (path.startsWith("/")) {
-            return AgentPackagePath.getPath() + path;
-        } else {
-            return AgentPackagePath.getPath() + "/" + path;
-        }
     }
 
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
@@ -60,8 +60,7 @@ public class TLSChannelBuilder implements ChannelBuilder<NettyChannelBuilder> {
                              InputStream key = PrivateKeyUtil.loadDecryptionKey(keyPath)) {
                             builder.keyManager(cert, key);
                         }
-                    }
-                    else if (!certFile.isFile() || !keyFile.isFile()) {
+                    } else if (!certFile.isFile() || !keyFile.isFile()) {
                         LOGGER.warn("Failed to enable mTLS caused by cert or key cannot be found.");
                     }
                 }
@@ -75,9 +74,7 @@ public class TLSChannelBuilder implements ChannelBuilder<NettyChannelBuilder> {
 
     private static String toAbsolutePath(final String path) throws AgentPackageNotFoundException {
         if (path.startsWith("/")) {
-            return path;
-        } else if (path.startsWith("./")) {
-            return AgentPackagePath.getPath() + path.substring(1);
+            return AgentPackagePath.getPath() + path;
         } else {
             return AgentPackagePath.getPath() + "/" + path;
         }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
@@ -23,31 +23,64 @@ import io.grpc.netty.NegotiationType;
 import io.grpc.netty.NettyChannelBuilder;
 import io.netty.handler.ssl.SslContextBuilder;
 import java.io.File;
-import javax.net.ssl.SSLException;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import org.apache.skywalking.apm.agent.core.boot.AgentPackageNotFoundException;
 import org.apache.skywalking.apm.agent.core.boot.AgentPackagePath;
 import org.apache.skywalking.apm.agent.core.conf.Config;
-import org.apache.skywalking.apm.agent.core.conf.Constants;
+import org.apache.skywalking.apm.agent.core.logging.api.ILog;
+import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
+import org.apache.skywalking.apm.agent.core.util.PrivateKeyUtil;
+import org.apache.skywalking.apm.util.StringUtil;
 
 /**
- * Detect the `/ca` folder in agent package, if `ca.crt` exists, start TLS (no mutual auth).
+ * If only ca.crt exists, start TLS. If cert, key and ca files exist, enable mTLS.
  */
 public class TLSChannelBuilder implements ChannelBuilder<NettyChannelBuilder> {
-    private static String CA_FILE_NAME = "ca" + Constants.PATH_SEPARATOR + "ca.crt";
+    private static final ILog LOGGER = LogManager.getLogger(TLSChannelBuilder.class);
 
     @Override
     public NettyChannelBuilder build(
-        NettyChannelBuilder managedChannelBuilder) throws AgentPackageNotFoundException, SSLException {
-        File caFile = new File(AgentPackagePath.getPath(), CA_FILE_NAME);
-        boolean isCAFileExist = caFile.exists() && caFile.isFile();
-        if (Config.Agent.FORCE_TLS || isCAFileExist) {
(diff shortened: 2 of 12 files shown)


wu-sheng:
Please fix dead link

docs/en/setup/service-agent/java-agent/TLS.md
```diff
@@ -17,10 +17,23 @@ Only support **no mutual auth**.
 ## Open and config TLS
 
 ### Agent config
-- Place `ca.crt` into `/ca` folder in agent package. Notice, `/ca` is not created in distribution, please create it by yourself.
```
wu-sheng, line 20: There was a sentence `Only support no mutual auth.` in this doc, please fix.

docs/en/setup/service-agent/java-agent/TLS.md
```diff
@@ -17,10 +17,23 @@ Only support **no mutual auth**.
...
 ```
-agent.force_tls=${SW_AGENT_FORCE_TLS:false}
+agent.force_tls=${SW_AGENT_FORCE_TLS:true}
+```
+
+## Enable mutual TLS
+
+- Sharing gRPC server must be started with enabled mTLS. More details see `receiver-sharing-server` section in `application.yaml`. Please refer to [gRPC SSL](../../backend/grpc-ssl.md)  
+- Configure Client-side SSL/TLS in `agent.conf`.
+- Change `SW_AGENT_COLLECTOR_BACKEND_SERVICES` to host and port of `receiver-sharing-server`.
```
wu-sheng, line 30:
```suggestion
- Change `SW_AGENT_COLLECTOR_BACKEND_SERVICES` targeting to host and port of `receiver-sharing-server`.
```

docs/en/setup/service-agent/java-agent/TLS.md
```diff
@@ -25,9 +24,9 @@ agent.force_tls=${SW_AGENT_FORCE_TLS:true}
 
 ## Enable mutual TLS
 
-- Sharing gRPC server must be started with enabled mTLS. More details see `receiver-sharing-server` section in `application.yaml`. Please refer to [gRPC SSL](../../backend/grpc-ssl.md)  
+- Sharing gRPC server must be started with enabled mTLS. More details see `receiver-sharing-server` section in `application.yaml` of SkyWalking OAP Server.
```
wu-sheng, line 27: What is `receiver-sharing-server` section for now? I think you should refer to the correct keyword? The keyword should be able to used here to search for documentation.

image

From https://github.com/apache/skywalking/pull/7565/files#diff-c6b2f69e272d64016aa069e5b2469b95aeabec06cea77c3f88b734dea84ff4cd, it should be `Security(SSL/TLS/mTLS)`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
```diff
@@ -23,31 +23,64 @@
...
-                                                         .sslContext(builder.build());
+            managedChannelBuilder.negotiationType(NegotiationType.TLS).sslContext(builder.build());
         }
         return managedChannelBuilder;
     }
+
+    private static String toAbsolutePath(final String path) throws AgentPackageNotFoundException {
+        if (path.startsWith("/")) {
+            return path;
+        } else if (path.startsWith("./")) {
```
hanahmily, line 79: How to handle other relative paths, for instance, `../`, `.././..` and etc?

What I mean is we don't have to parse relative paths, just appending the agent package path to it looks like a tidier way

docs/en/setup/service-agent/java-agent/TLS.md
```diff
@@ -17,10 +16,23 @@ Only support **no mutual auth**.
...
+
+- Sharing gRPC server must be started with enabled mTLS. More details see `receiver-sharing-server` section in `application.yaml` of SkyWalking OAP Server.  
+- Configure Client-side SSL/TLS in `agent.conf`.
+- Change `SW_AGENT_COLLECTOR_BACKEND_SERVICES` targeting to host and port of `receiver-sharing-server`.
+
+For example:
 ```
+agent.force_tls=${SW_AGENT_FORCE_TLS:true}
+agent.ssl_trusted_ca_path=${SW_AGENT_SSL_TRUSTED_CA_PATH:/path/to/ca.crt}
+agent.ssl_key_path=${SW_AGENT_SSL_KEY_PATH:/path/to/client.pem}
```
hanahmily, line 35: We should document which private key format is supported just like the server-side.

hanahmily (commented):
Some nits

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
dmsolr: Make sense.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
```diff
@@ -23,31 +23,64 @@
...
+                    File keyFile = new File(toAbsolutePath(keyPath));
+                    File certFile = new File(toAbsolutePath(certPath));
+
+                    if (certFile.isFile() && keyFile.isFile()) {
+                        try (InputStream cert = new FileInputStream(certFile);
+                             InputStream key = PrivateKeyUtil.loadDecryptionKey(keyPath)) {
+                            builder.keyManager(cert, key);
+                        }
+                    }
+                    else if (!certFile.isFile() || !keyFile.isFile()) {
```
kezhenxu94, line 64:
```suggestion
                    } else if (!certFile.isFile() || !keyFile.isFile()) {
```

docs/en/setup/service-agent/java-agent/TLS.md
kezhenxu94: Specifically, line 11 @dmsolr

docs/en/setup/service-agent/java-agent/TLS.md
```diff
@@ -17,10 +17,23 @@ Only support **no mutual auth**.
 ## Open and config TLS
 
 ### Agent config
-- Place `ca.crt` into `/ca` folder in agent package. Notice, `/ca` is not created in distribution, please create it by yourself.
-
-- Agent open TLS automatically after the `/ca/ca.crt` file detected.
+- Agent open TLS automatically after the `ca.crt`(by default `/ca` folder in agent package) file detected.
```
kezhenxu94, line 19:
```suggestion
- Agent enables TLS automatically after the `ca.crt`(by default `/ca` folder in agent package) file is detected.
```

docs/en/setup/service-agent/java-agent/TLS.md
```diff
@@ -17,10 +17,23 @@ Only support **no mutual auth**.
...
 - TLS with no CA mode could be activated by this setting.
 ```
-agent.force_tls=${SW_AGENT_FORCE_TLS:false}
+agent.force_tls=${SW_AGENT_FORCE_TLS:true}
+```
+
+## Enable mutual TLS
+
+- Sharing gRPC server must be started with enabled mTLS. More details see `receiver-sharing-server` section in `application.yaml`. Please refer to [gRPC SSL](../../backend/grpc-ssl.md)
```
kezhenxu94:
```suggestion
- Sharing gRPC server must be started with mTLS enabled. More details can be found in `receiver-sharing-server` section in `application.yaml`. Please refer to [gRPC SSL](../../backend/grpc-ssl.md)  
```

dmsolr: The gRPC SSL refers to the main repo. So we cannot use a relative path. Could we use this URL, https://skywalking.apache.org/docs/main/latest/en/setup/backend/grpc-security/

wu-sheng: Please don't. This is a dev doc(notice `latest`). We never will have a static path for backend document because it changes with versions, including the URL. That is why I recommend for `key words` to search document. 

@Jtrust is going to add doc search for every versions of documents. Then it is more friendly and useful than right now.

docs/en/setup/service-agent/java-agent/TLS.md
dmsolr: Got it

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
```diff
@@ -23,31 +23,61 @@
...
         }
         return managedChannelBuilder;
     }
+
+    private static String toAbsolutePath(final String path) throws AgentPackageNotFoundException {
+        if (path.startsWith("/")) {
+            return AgentPackagePath.getPath() + path;
+        } else {
+            return AgentPackagePath.getPath() + "/" + path;
+        }
```
wu-sheng, line 80: You deleted line `File caFile = new File(AgentPackagePath.getPath(), CA_FILE_NAME);` is for avoiding this complex `if/else`. Could you share way removing that line?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
```diff
@@ -23,31 +23,61 @@
...
  */
 public class TLSChannelBuilder implements ChannelBuilder<NettyChannelBuilder> {
-    private static String CA_FILE_NAME = "ca" + Constants.PATH_SEPARATOR + "ca.crt";
+    private static final ILog LOGGER = LogManager.getLogger(TLSChannelBuilder.class);
 
     @Override
     public NettyChannelBuilder build(
-        NettyChannelBuilder managedChannelBuilder) throws AgentPackageNotFoundException, SSLException {
-        File caFile = new File(AgentPackagePath.getPath(), CA_FILE_NAME);
-        boolean isCAFileExist = caFile.exists() && caFile.isFile();
```
wu-sheng, line 42: Also, why remove this? It seems you are copying `caFile.isFile()` in 2 places instead. I am confused about the principle of code change.

docs/en/setup/service-agent/java-agent/TLS.md
```diff
@@ -8,19 +8,34 @@ at the same time, the SkyWalking backend is in another region (VPC).
...
+- Sharing gRPC server must be started with mTLS enabled. More details can be found in `receiver-sharing-server` section in `application.yaml`. Please refer to `gRPC Security` and `gRPC/HTTP server for receiver`.
+- Copy CA certificate, certificate and private key of client into `agent/ca`.
+- Configure client-side SSL/TLS in `agent.conf`.
+- Change `SW_AGENT_COLLECTOR_BACKEND_SERVICES` targeting to host and port of `receiver-sharing-server`.
+
+For example:
 ```
+agent.force_tls=${SW_AGENT_FORCE_TLS:true}
+agent.ssl_trusted_ca_path=${SW_AGENT_SSL_TRUSTED_CA_PATH:/ca/ca.crt}
+agent.ssl_key_path=${SW_AGENT_SSL_KEY_PATH:/ca/client.pem}
```
wu-sheng, line 35: It seems this https://github.com/apache/skywalking-java/pull/15#discussion_r698096325 still gets unresolved, could you share why?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
dmsolr: Before, CA certificate file is a default value which was `/ca/ca.cert`. Now, I make it is configurable. Because users need to configure the client's certificate and private key. I think CA certificate is still unconfigurable, which is a little odd.

As they are configurable, I think we need to adapt start with `/` or not. Sure, it is unnecessary. Required users fill the path started with `/` and documented it in docs.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
dmsolr: Remove means remove this part `caFile.exist()`, right. `File#isFile()` is true means the file existed.
JDK docs say:
> return: true if and only if the file denoted by this abstract pathname exists and is a normal file; false otherwise

Tests `caFile#isFile()` twice, that is previously logic. There is a case to enable  SSL/TLS with a none CA certificate by force switch. And the CA file is detected, SSL/TLS is enabled by default. 
The second time, the CA certificate has to load it to SSL context when it existed.

docs/en/setup/service-agent/java-agent/TLS.md
dmsolr: Sorry, I forget :(.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
wu-sheng: The previous logic only calls the method once, even check twice. The second API call is unnecessary.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
wu-sheng: Then your logic is back to https://github.com/apache/skywalking-java/pull/15#discussion_r698096140. This is not what we need to deal with, and it is endless. 

And I am not meaning about configurable. My point is `File caFile = new File(AgentPackagePath.getPath(), CA_FILE_NAME);`. This is usually the standard way to process a file/path of a parent path. Let's keep in this way.

wu-sheng: Once you did this in the codebase, we could have to face pull requests or requests to change how to process file path in the whole project, which could be confused.

dmsolr: I got your point.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -135,6 +135,21 @@
...
+
+        /**
+         * Key cert chain file. If ssl_cert_chain and ssl_key exist, will enable mTLS for gRPC channel.
+         */
+        public static String SSL_CERT_CHAIN_PATH;
+
+        /**
+         * Private key file. If ssl_cert_chain and ssl_key exist, will enable mTLS for gRPC channel.
+         */
+        public static String SSL_KEY_PATH;
```
wu-sheng, line 152: According to recent requirements, all configuration should be added in default `agent.config`, with all system env variable names should be documents in agent doc.

wu-sheng (approved):
LGTM. @kezhenxu94 @hanahmily Please recheck.

As this is a core enhancement, we have to make the e2e test after merging.

hanahmily (approved):
LGTM

PR #16, opened 2021-08-31
Fix do not have using absolute path of private key

dmsolr:
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/TLSChannelBuilder.java
```diff
@@ -58,7 +58,7 @@ public NettyChannelBuilder build(
 
                     if (certFile.isFile() && keyFile.isFile()) {
                         try (InputStream cert = new FileInputStream(certFile);
-                             InputStream key = PrivateKeyUtil.loadDecryptionKey(keyPath)) {
+                             InputStream key = PrivateKeyUtil.loadDecryptionKey(keyFile.getAbsolutePath())) {
                             builder.keyManager(cert, key);
                         }
```
wu-sheng, line 63: Is this a code format issue? The brace seems strange. Why there is `;` after try?

wu-sheng (approved):

PR #17, opened 2021-09-01
optimize lettuce plugin

darcydai:
### Fix <lettuce plugin has a wrong design for record span time elapsed>

## This is a big change for the lettuce plugin
Lettuce is an async Redis client base on netty, and wide use in our company.
Recently I found almost every lettuce span has the time elapsed less than 1ms and has a long gap between two spans. After reading the lettuce source code, I found the plugin some bug.
because  lettuce is an async framework base on netty, old interceptor named 'RedisChannelWriterInterceptor' stop span in after method can not record the time elapsed, it should be close in 'RedisCommand#complete' or 'RedisCommand#completeExceptionally' or 'RedisCommand#cancel'   called when Redis server response 
lettuce plugin record nothing when lettuce work on reactive mode (project reactor)
record Redis peer in skywalking dynamic field can simplify

The new plugin can work with Redis standalone and Redis Sentinel and Redis cluster(old version can not).  Redis cluster and Redis sentinel need to use docker-compose. I do not know how to make scenarios test in this two cluster model


test/plugin/scenarios/lettuce-scenario/support-version.list
```diff
@@ -16,4 +16,7 @@
 
 5.2.1.RELEASE
 5.1.8.RELEASE
-5.0.5.RELEASE
\ No newline at end of file
+5.0.5.RELEASE
+6.0.0.RELEASE
+6.0.5.RELEASE
```
wu-sheng, line 21:
```suggestion
6.0.5.RELEASE
```

We recommend keep one version for minor version, and ignore revision to reduce resource cost.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
```diff
@@ -96,6 +96,8 @@ private Tags() {
      */
     public static final StringTag RPC_RESPONSE_STATUS_CODE = new StringTag(18, "rpc.status_code", true);
 
+    public static final StringTag SIGNAL_TYPE = new StringTag(19, "signalType");
```
wu-sheng, line 99: What is this tag? It seems don't have general meaning.

apm-sniffer/apm-sdk-plugin/lettuce-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/lettuce/v5/RedisCommandCancelMethodInterceptor.java
```diff
@@ -18,49 +18,39 @@
...
-        Object ret) throws Throwable {
-        ContextManager.stopSpan();
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) {
+        if (objInst.getSkyWalkingDynamicField() != null) {
+            AbstractSpan span = (AbstractSpan) objInst.getSkyWalkingDynamicField();
+            span.errorOccurred();
+            Tags.SIGNAL_TYPE.set(span, COMMAND_CANCEL_VALUE);
+            span.asyncFinish();
+            //help gc
+            objInst.setSkyWalkingDynamicField(null);
```
wu-sheng, line 44: Is the `objInst` going to be reused in the whole Redis client process? If not,  I am not this `set null` whether really help GC.

wu-sheng:
> The new plugin can work with Redis standalone and Redis Sentinel and Redis cluster(old version can not). Redis cluster and Redis sentinel need to use docker-compose. I do not know how to make scenarios test in this two cluster model

It is fine to skip that. If you want, it is better to update support list doc to mention all these.

apm-sniffer/apm-sdk-plugin/lettuce-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/lettuce/v5/RedisCommandCancelMethodInterceptor.java
darcydai: yes, 'set null' another usage is prevented async finish repeatedly. this 'objInst' is Redis Command will return to Redis user, 'set null' can let Redis Command not refer to span after it finished

test/plugin/scenarios/lettuce-scenario/support-version.list
darcydai: removed

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
darcydai: removed

apm-sniffer/apm-sdk-plugin/lettuce-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/lettuce/v5/RedisChannelWriterInterceptor.java
```diff
@@ -89,22 +105,37 @@ private String getArgsStatement(RedisCommand redisCommand) {
...
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) {
         return ret;
     }
 
     @Override
-    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments,
-        Class<?>[] argumentsTypes, Throwable t) {
+    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Throwable t) {
         AbstractSpan span = ContextManager.activeSpan();
         span.log(t);
```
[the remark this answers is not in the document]
darcydai: updated

wu-sheng (approved):
LGTM. @darcydai Do you need to update the document to mention this?

> The new plugin can work with Redis standalone and Redis Sentinel and Redis cluster(old version can not). Redis cluster and Redis sentinel need to use docker-compose. I do not know how to make scenarios test in this two cluster model

darcydai:
> LGTM. @darcydai Do you need to update the document to mention this?
> 
> > The new plugin can work with Redis standalone and Redis Sentinel and Redis cluster(old version can not). Redis cluster and Redis sentinel need to use docker-compose. I do not know how to make scenarios test in this two cluster model

where should I write this？

wu-sheng:
How about adding (cluster..) after redis on this page?

https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/supported-list/

PR #18, opened 2021-09-02
fix multi-db-instance on same host port

lpcy:
### Fix https://github.com/apache/skywalking/issues/4941
       When my java application depend on multi-db-instance on same host:port，I always got wrong db.instance value in skywalking-ui. I found the same problem on that issue, there is  a bad cache design in MySQL agent plugin, so I fix it.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. skywalking/#4941.
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking/blob/master/CHANGES.md).


apm-sniffer/apm-sdk-plugin/mysql-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v6/ConnectionCreateOldInterceptor.java
```diff
@@ -41,7 +41,12 @@ public void beforeMethod(Class clazz, Method method, Object[] allArguments, Clas
     public Object afterMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes,
         Object ret) {
         if (ret instanceof EnhancedInstance) {
-            ConnectionInfo connectionInfo = ConnectionCache.get(allArguments[1].toString(), allArguments[2].toString());
+            String database = "";
+            try {
+                Method getDbMethod = parameterTypes[0].getDeclaredMethod("getDatabase");
+                database = (String) getDbMethod.invoke(allArguments[0]);
+            } catch (Throwable t) { }
```
wu-sheng, line 48: Why do you need this reflector? You could covert `parameterTypes[0]` to the target type and call method directly

wu-sheng:
First, why plugin tests fail?

> When my java application depend on multi-db-instance on same host:port，I always got wrong db.instance value in skywalking-ui. I found the same problem on that issue, there is a bad cache design in MySQL agent plugin, so I fix it.

Then, you don't explain what is wrong, and how you fix. `Found one` and `fix it` don't make sense for reviewers.

apm-sniffer/apm-sdk-plugin/mysql-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v6/ConnectionCreateOldInterceptor.java
lpcy: Because this plugin already exists one mysql-connector-java dependency, if I add the old version dependency and use the target type directly, I won't be able to compile it. Do you have any suggestions?

wu-sheng: Do you mean this `getDatabase` method is not available in some versions?

lpcy:
> First, why plugin tests fail?
> 
> > When my java application depend on multi-db-instance on same host:port，I always got wrong db.instance value in skywalking-ui. I found the same problem on that issue, there is a bad cache design in MySQL agent plugin, so I fix it.
> 
> Then, you don't explain what is wrong, and how you fix. `Found one` and `fix it` don't make sense for reviewers.

I'm still look for  the failure...  
The problem is, if I use multi-db-instance with same host:port in one application, the code for putting connection info in cache will overwrite previous data due to duplicate host:port key, the previous db instance info will be lost, and also cannot get correct connection info from cache after that.

wu-sheng:
> The problem is, if I use multi-db-instance with same host:port in one application, the code for putting connection info in cache will overwrite previous data due to duplicate host:port key, the previous db instance info will be lost, and also cannot get correct connection info from cache after that.

This connection create interceptor is going to be called repeatedly for different instances? Could you put some reference of MySQL driver's source codes about this? Is it a loop call or something?

apm-sniffer/apm-sdk-plugin/mysql-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v6/ConnectionCreateOldInterceptor.java
lpcy: mysql-connector-java version 6.0.6 doesn't have parameterTypes[0] class (ie. ConnectionString), and version 6.0.2 doesn't have com.mysql.cj.core.conf.url.HostInfo class  (ConnectionCreateNewInterceptor needing it), but I need those two classes  to coexist if use the target type directly.

wu-sheng: If that method doesn't exist, how your fix is correct? You just try to catch a known exception, but don't help for the result.

lpcy:
> > The problem is, if I use multi-db-instance with same host:port in one application, the code for putting connection info in cache will overwrite previous data due to duplicate host:port key, the previous db instance info will be lost, and also cannot get correct connection info from cache after that.
> 
> This connection create interceptor is going to be called repeatedly for different instances? Could you put some reference of MySQL driver's source codes about this? Is it a loop call or something?

```java
class ConnectionCache：
public static void save(ConnectionInfo connectionInfo) {
        for (String conn : connectionInfo.getDatabasePeer().split(CONNECTION_SPLIT_STR)) {
            if (!StringUtil.isEmpty(conn)) {
                CONNECTIONS_MAP.putIfAbsent(conn, connectionInfo);
            }
        }
    }
If the connections are :
  A:    jdbc:mysql://hostA:portA/databaseA?propertyName1=propertyValue1
  B :   jdbc:mysql://hostA:portA/databaseB?propertyName1=propertyValue1

the connection info A will be covered by B and be lost due to duplicate key "conn".

apm-sniffer/apm-sdk-plugin/mysql-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v6/ConnectionCreateOldInterceptor.java
lpcy: There are two Interceptor for two mysql-connector-java version in one plugin.My point is how to add two mysql-connector-java version dependencies for use the target type directly  and separately.

wu-sheng:
But they should be different connection, my question is why they share the same cache.

lpcy:
Because there are same host:port pair, and the key of CONNECTIONS_MAP have "host:port" format, and can't store two connections A and B above. So how to get correct cache after that?

wu-sheng:
Then, the better fix is changing the key to connection url

lpcy:
I think the cache design included purpose of identify database cluster mode. Because there is another jdbc url format: 
      jdbc:mysql://[host:port],[host:port].../database, the code "split(CONNECTION_SPLIT_STR)" is use for it, when call the get(host:port) method, the full url above can get. I don't know if it's able to change the strategy.

wu-sheng:
From current status, ip and port represents a database instance, we don't care what instances are inside. 
Right now, you want to enhance this to include the database instance in, something has to be changed. Enhancing cache(not have to be url as key) and make the spam tag more reasonable is the right way to fix this.

apm-sniffer/apm-sdk-plugin/mysql-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v6/ConnectionCreateOldInterceptor.java
wu-sheng: They should be separated, and use witness class to identify the versions.

lpcy: However, Maven cannot add two versions  of mysql-connector-java dependencies. So I use the reflector.

lpcy:
OK, but wrong db.instance value displayed on skywalking-ui cannot be tolerated to me, so I suggest fix it after all. My code don't affect value of cache (i.e. ConnectionInfo) but only key, and there are few code location to call get(host,port,databaseName) method, so it doesn't affect the current status that "ip and port represents a database instance". Now my code passed by all checks, please consider the PR.

wu-sheng:
I don't mean the current codes are correct. My keypoints for your changes are
1. We want to avoid reflections in your codes. Are they possible? If not, is it only executing once a new connection created?
2. If the database instance name matter, why don't we change the tag(instance name) value? We are an APM, if information in trace is changed, those should be consistent in metrics.

lpcy:
Our previous communication seem impossible to achieve without reflections due to rule of Maven, but reflections is called in ConnectionCreateOldInterceptor only (for mysql connector java 6.0.2,6.0.3), so its impact is limited. And it is only executing once a new connection created because point of the Interceptor is com.mysql.*.ConnectionImpl.getInstance(...) and prepared for getSkyWalkingDynamicField in jdbc Statement called later.
I can't understand "change the tag(instance name) value", and I think that's what I'm doing now，are you talking about not use connection cache in Interceptor? I didn't change the host:port, I don't know where the metrics will be change.

wu-sheng:
> Our previous communication seem impossible to achieve without reflections due to rule of Maven, but reflections is called in ConnectionCreateOldInterceptor only (for mysql connector java 6.0.2,6.0.3), so its impact is limited. And it is only executing once a new connection created because point of the Interceptor is com.mysql.*.ConnectionImpl.getInstance(...) and prepared for getSkyWalkingDynamicField in jdbc Statement called later.

Then, it is fine.

wu-sheng:
> I can't understand "change the tag(instance name) value", and I think that's what I'm doing now，are you talking about not use connection cache in Interceptor? I didn't change the host:port, I don't know where the metrics will be change.

I was thinking, you want to change the name in the database list on the UI. It is `ip:port`, I thought you want to change it to `ip:port:database-name` too.

wu-sheng:
Please update the changes.md in the root. We need that.

lpcy:
I have updated changes.md. Thank you and admire your work!

wu-sheng (approved):
LGTM

wu-sheng:
> I have updated changes.md. Thank you and admire your work!

Thank you for contributing back to upsteam.

PR #21, opened 2021-09-08
Migrate base Docker image from JDK to JRE and remove unused classes

kezhenxu94:


wu-sheng (commented):
Same asking like main repo, please add how to use doc.

kezhenxu94:
> Same asking like main repo, please add how to use doc.

It's already there, just do some polish

wu-sheng (approved):

PR #22, opened 2021-09-12
Add thrift plugin support thrift TMultiplexedProcessor.

zifeihan:
### Fix thrift plugin does not support thrift TMultiplexedProcessor, so NPE is triggered
   It has been tested through the demo submitted by the user. I have been too busy recently. I will submit the test scenario in two days.
   Check [7571]( https://github.com/apache/skywalking/issues/7571) to understand the cause of the bug and how to fix it.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #7571 .

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -13,14 +13,14 @@
 
 <!-- ==== 🔌 Remove this line WHEN AND ONLY WHEN you're adding a new plugin, follow the checklist 👇 ====
 ### Add an agent plugin to support <framework name>
-- [ ] Add a test case for the new plugin, refer to [the doc](https://github.com/apache/skywalking/blob/master/docs/en/guides/Plugin-test.md)
-- [ ] Add a component id in [the component-libraries.yml](https://github.com/apache/skywalking/blob/master/oap-server/server-bootstrap/src/main/resources/component-libraries.yml)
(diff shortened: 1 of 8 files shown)


dmsolr:
I think this is an enhancement. :)
We don't implement all `Processor`s.

dmsolr:
Hi @lzdujing Please check it if you have time.

dmsolr:
@zifeihan maybe we need to list of `Processor`s in docs which we have enhanced, WDYT?

zifeihan:
> @zifeihan maybe we need to list of `Processor`s in docs which we have enhanced, WDYT?

Ihink it is necessary. Currently thrift only has 3 processors, and we currently support all of them.

dmsolr:
> > @zifeihan maybe we need to list of `Processor`s in docs which we have enhanced, WDYT?
> 
> I think it is necessary. Currently thrift only has 3 processors, and we currently support all of them.

Actually, not only these are 3. There are also many wrapped processors.

zifeihan:
image
The object here is new, so it does not go through ServerInProtocolWrapper, so special judgment is required

dmsolr (approved):
LGTM, @zifeihan told me who is too busy to complete docs.

wu-sheng:
> LGTM, @zifeihan told me who is too busy to complete docs.

Hi, we don't have typically documents to host all logic of plugins, they are too complex to describe. AFAIK, only very few plugin added README.md in the root of plugin source codes only. @dmsolr If you want to, welcome to send a pull request later.

PR #31, opened 2021-09-22
Support for Apache HttpClient 5 plugin

heihaozi:
### Support for Apache HttpClient 5 plug

---

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes [apache/skywalking#7763](https://github.com/apache/skywalking/issues/7763).


wu-sheng:
Please fix CI, and you need to follow plugin dev doc to add integration tests for new plugins, which makes sure this works in both old and new(5.x) releases.

heihaozi:
OK, Does the integration test create the corresponding project in `\test\plugin\scenarios`?

wu-sheng:
> OK, Does the integration test create the corresponding project in `\test\plugin\scenarios`?

Yes, do you know where the document is?

heihaozi:
I don’t know, where is the document?

wu-sheng:
1. https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/java-plugin-development-guide/
2. https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/plugin-test/ linked from (1).

heihaozi:
I am using a windows system and cannot execute the `bash ./test/plugin/run.sh -f ${scenario_name}` command. 
Is there an alternative?

wu-sheng:
> I am using a windows system and cannot execute the `bash ./test/plugin/run.sh -f ${scenario_name}` command.
> Is there an alternative?

It only works on Linux or MacOS. Could you find one?

heihaozi:
I will try.

wu-sheng:
`/docs/en/setup/service-agent/java-agent/Plugin-list.md` should be updated in every time you add a new plugin.

wu-sheng:
Also `Supported-list.md`

wu-sheng:
Your new case has not been added into GitHub action control file, so, no verification will happen in the CI process.

heihaozi:
Is the `segments` in `expectedData.yaml` in order?

wu-sheng:
> Is the `segments` in `expectedData.yaml` in order?

For segments, NO. Because those are not expected, mock traffic many trace segments, we just check whether there is at least one matched.

test/plugin/scenarios/httpclient-5.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,120 @@
...
+              - {key: url, value: 'http://127.0.0.1:8080/httpclient-5.x/case/asyncGet'}
+              - {key: http.method, value: GET}
+            skipAnalysis: 'false'
+          - operationName: /httpclient-5.x/case/get
+            parentSpanId: -1
+            spanId: 0
+            spanLayer: Http
+            startTime: nq 0
+            endTime: nq 0
+            componentId: not null
```
wu-sheng, line 113: Please use real componentId in the expectedData file.

apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpAsyncClientDoExecuteInterceptor.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+            MethodInterceptResult result) throws Throwable {
+        AsyncResponseConsumer consumer = (AsyncResponseConsumer) allArguments[2];
+        HttpContext context = (HttpContext) allArguments[4];
+        FutureCallback callback = (FutureCallback) allArguments[5];
+        allArguments[2] = new AsyncResponseConsumerWrapper(consumer);
+        allArguments[5] = new FutureCallbackWrapper(callback);
+        if (ContextManager.isActive() && !ContextManager.activeSpan().isExit()) {
```
wu-sheng, line 43: Why do we need to check `isExit`?

apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpClientDoExecuteInterceptor.java
```diff
@@ -0,0 +1,137 @@
...
+            return uri;
+        }
+    }
+
+    private boolean isUrl(String uri) {
+        String lowerUrl = uri.toLowerCase();
+        return lowerUrl.startsWith("http") || lowerUrl.startsWith("https");
+    }
+
+    private String buildSpanValue(HttpHost httpHost, String uri) {
```
wu-sheng, line 118:
```suggestion
    private String buildURL(HttpHost httpHost, String uri) {
```

apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpAsyncClientDoExecuteInterceptor.java
wu-sheng: What is the case there is no `exit span` available? We usually use this kind of check when these are expected.

test/plugin/scenarios/httpclient-5.x-scenario/config/expectedData.yaml
heihaozi: OK, I add it.

apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpAsyncClientDoExecuteInterceptor.java
heihaozi: This is invalid code, I will remove it.

wu-sheng:
> Update suggested change

You should not do this, I didn't change all related things. Please apply changes locally.

test/plugin/scenarios/httpclient-5.x-scenario/config/expectedData.yaml
```diff
@@ -110,7 +110,7 @@ segmentItems:
             spanLayer: Http
             startTime: nq 0
             endTime: nq 0
-            componentId: not null
+            componentId: SpringMVC
```
wu-sheng, line 113: This is ID, not name.

heihaozi: OK, I fix it.

CHANGES.md
```diff
@@ -26,6 +26,7 @@ Release Notes.
     rename `plugin.toolkit.log.grpc.reporter.max_message_size` to `log.max_message_size`.
 * Implement Kafka Log Reporter. Add config item of `agnt.conf`, `plugin.kafka.topic_logging`.
 * Upgrade byte-buddy to 1.11.18
+* Support for Apache HttpClient 5 plug.
```
wu-sheng, line 29:
```suggestion
* Add plugin to support Apache HttpClient 5.
```

wu-sheng (approved):

wankai123 (dismissed):

test/plugin/scenarios/httpclient-5.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,120 @@
...
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: /httpclient-5.x/back
+            parentSpanId: -1
+            spanId: 0
+            spanLayer: Http
+            startTime: nq 0
+            endTime: nq 0
+            componentId: 14
```
wu-sheng, line 28: As your controller in `org.apache.skywalking`, this wouldn't be traced by SpringMVC plugin, but Tomcat plugin should work.

```suggestion
            componentId: 1
```

test/plugin/scenarios/httpclient-5.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,120 @@
...
+            skipAnalysis: 'false'
+      - segmentId: not null
+        spans:
+          - operationName: /httpclient-5.x/case/asyncGet
+            parentSpanId: -1
+            spanId: 0
+            spanLayer: Http
+            startTime: nq 0
+            endTime: nq 0
+            componentId: 14
```
wu-sheng, line 79:
```suggestion
            componentId: 1
```

Same here.

test/plugin/scenarios/httpclient-5.x-scenario/config/expectedData.yaml
wu-sheng: I expect the tests will fail.

wu-sheng (approved):

PR #34, opened 2021-09-22
Fix a tracing context leak in SpringMVC plugin

furaul:
Once the springMVC span is created in beforeMethod, it must be stopped in afterMethod. Or it may cause memory leak.

- When Exception throws in afterMethod of AbstractMethodInterceptor, the span will not be stopped, and it is not excepted.
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/7734 .
- This PR will not fix why the exception is throwed, but it will eliminate effects that the exception takes.
- The new plugin has been used in our company, on product enviroment.


wu-sheng:
I would recommend you to upgrade this interceptor to v2 API. Take a look at this, https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/java-plugin-development-guide/#v2-apis-1

Once span created, you could use `MethodInvocationContext` to propagate for the `afterMethod`. Then you don't need to check depth anymore, which is actually not being used by any plugin.

apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java
```diff
@@ -198,56 +198,58 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
...
+                // Active HTTP parameter collection automatically in the profiling context.
+                if (!SpringMVCPluginConfig.Plugin.SpringMVC.COLLECT_HTTP_PARAMS && span.isProfiling()) {
+                    if (HttpServletRequest.class.isAssignableFrom(request.getClass())) {
+                        RequestUtil.collectHttpParam((HttpServletRequest) request, span);
+                    } else if (ServerHttpRequest.class.isAssignableFrom(request.getClass())) {
+                        RequestUtil.collectHttpParam((ServerHttpRequest) request, span);
+                    }
                 }
+            } finally {
+                ContextManager.stopSpan();
```
wu-sheng, line 251: Are you only adding this `try/finally` to stop span? Could you share when this last step goes missed?

furaul: > Are you only adding this `try/finally` to stop span? Could you share when this last step goes missed?

Yes, I only did this. 

The last step goes missed when the response == null and the exception ServletResponseNotFoundException throws.

It is complicated to reproduced why the response becomes null, since it is only appear in a few requests. And if I debug into the request, drop the frame, the response becomes not null.

Maybe I could read more springMVC source codes later and find out the root cause.

wu-sheng: OK, to add this to keep tracing context safe is correct. I just think, this may not fix the real issue about why response is null, and we can't get all response related information.

wu-sheng (approved):

PR #39, opened 2021-09-27
Add an agent plugin to support Jackson

VictorZeng:
### Add an agent plugin to support `Jackson`

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/7812.

### DIFF
--- a/apm-sniffer/optional-plugins/jackson-2.x-plugin/pom.xml
+++ b/apm-sniffer/optional-plugins/jackson-2.x-plugin/pom.xml
@@ -44,12 +44,6 @@
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-annotations</artifactId>
-            <version>${jackson.version}</version>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
             <version>${jackson.version}</version>
             <scope>provided</scope>
--- a/apm-sniffer/optional-plugins/jackson-2.x-plugin/pom.xml
+++ b/apm-sniffer/optional-plugins/jackson-2.x-plugin/pom.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.skywalking</groupId>
+        <artifactId>optional-plugins</artifactId>
+        <version>8.8.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apm-jackson-2.x-plugin</artifactId>
+    <name>jackson-2.x-plugin</name>
+    <packaging>jar</packaging>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jackson.version>2.9.9</jackson.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>${jackson.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>${jackson.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson.version}</version>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/BasicMethodsInterceptor.java
+++ b/apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/BasicMethodsInterceptor.java
@@ -34,7 +34,7 @@ public class BasicMethodsInterceptor implements InstanceMethodsAroundInterceptor
     @Override
     public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
         AbstractSpan span = ContextManager.createLocalSpan(
-                OPERATION_NAME_JACKSON + method.getDeclaringClass().getName() + "/" + method.getName());
+                OPERATION_NAME_JACKSON + method.getDeclaringClass().getSimpleName() + "." + method.getName());
         span.setComponent(ComponentsDefine.JACKSON);
     }
 
--- a/apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/BasicMethodsInterceptor.java
+++ b/apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/BasicMethodsInterceptor.java
@@ -32,20 +32,24 @@ public class BasicMethodsInterceptor implements InstanceMethodsAroundInterceptor
     public static final String OPERATION_NAME_JACKSON = "Jackson/";
 
     @Override
-    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
-        AbstractSpan span = ContextManager.createLocalSpan(
-                OPERATION_NAME_JACKSON + method.getDeclaringClass().getSimpleName() + "." + method.getName());
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
+                             Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+
+        AbstractSpan span = ContextManager.createLocalSpan(OPERATION_NAME_JACKSON +
+                method.getDeclaringClass().getSimpleName() + "." + method.getName());
         span.setComponent(ComponentsDefine.JACKSON);
     }
 
     @Override
-    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
+                              Class<?>[] argumentsTypes, Object ret) throws Throwable {
         ContextManager.stopSpan();
         return ret;
     }
 
     @Override
-    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Throwable t) {
+    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments,
+                                      Class<?>[] argumentsTypes, Throwable t) {
         ContextManager.activeSpan().log(t);
     }
 }
--- a/apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/BasicMethodsInterceptor.java
+++ b/apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/BasicMethodsInterceptor.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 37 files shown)


apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/define/AbstractInstrumentation.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInstanceMethodsEnhancePluginDefine;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+/**
+ * Jackson has been known as "the Java JSON library" or "the best JSON parser for Java". Or simply as "JSON for Java".
```
kezhenxu94, line 34: Hi @VictorZeng , can you remove the words `the best` and avoid using similar words in the future? I don't mean they are wrong or Jackson is not good enough, but in Apache projects, we should not endorse for any third-parties / vendors.

apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/define/AbstractInstrumentation.java
```diff
@@ -0,0 +1,81 @@
...
+
+                @Override
+                public boolean isOverrideArgs() {
+                    return false;
+                }
+            };
+            points.add(point);
+        }
+
+        return points.toArray(new InstanceMethodsInterceptPoint[points.size()]);
```
kezhenxu94, line 71: This is recommended, I remember the Fastjson did the same thing, can you update that too?

```suggestion
        return points.toArray(new InstanceMethodsInterceptPoint[0]);
```

apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/define/ObjectMapperInstrumentation.java
```diff
@@ -0,0 +1,53 @@
...
+package org.apache.skywalking.apm.plugin.jackson.define;
+
+import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.NameMatch;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Jackson has been known as "the Java JSON library" or "the best JSON parser for Java". Or simply as "JSON for Java".
```
kezhenxu94, line 28: Same here

apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/define/ObjectMapperInstrumentation.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    protected Map<String, String> enhanceMethods() {
+        return new HashMap<String, String>() {
+            {
+                put("writeValue", "org.apache.skywalking.apm.plugin.jackson.BasicMethodsInterceptor");
+                put("writeValueAsString", "org.apache.skywalking.apm.plugin.jackson.WriteValueAsStringInterceptor");
+                put("writeValueAsBytes", "org.apache.skywalking.apm.plugin.jackson.WriteValueAsBytesInterceptor");
+                put("readValue", "org.apache.skywalking.apm.plugin.jackson.ReadValueInterceptor");
+            }
+        };
```
kezhenxu94, line 51: Consider using a builder to build the map instead of creating anonymous inner classes, which creates unnecessary class.

```suggestion
        return ImmutableMap.<String, String>builder()
                           .put(
                               "writeValue",
                               "org.apache.skywalking.apm.plugin.jackson.BasicMethodsInterceptor"
                           )
                           .put(
                               "writeValueAsString",
                               "org.apache.skywalking.apm.plugin.jackson.WriteValueAsStringInterceptor"
                           )
                           .put(
                               "writeValueAsBytes",
                               "org.apache.skywalking.apm.plugin.jackson.WriteValueAsBytesInterceptor"
                           )
                           .put(
                               "readValue",
                               "org.apache.skywalking.apm.plugin.jackson.ReadValueInterceptor"
                           )
                           .build();
```

apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/BasicMethodsInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
+
+import java.lang.reflect.Method;
+
+public class BasicMethodsInterceptor implements InstanceMethodsAroundInterceptor {
```
kezhenxu94, line 30: Import our code style settings and format this file, it has some lines that are too long

apm-sniffer/optional-plugins/jackson-2.x-plugin/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>${jackson.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
```
kezhenxu94, line 47: Is this needed? Seems it's unused

apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/define/AbstractInstrumentation.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public ConstructorInterceptPoint[] getConstructorsInterceptPoints() {
+        return new ConstructorInterceptPoint[0];
+    }
+
+    @Override
+    public InstanceMethodsInterceptPoint[] getInstanceMethodsInterceptPoints() {
+
+        final List<InstanceMethodsInterceptPoint> points = new ArrayList<InstanceMethodsInterceptPoint>(this.enhanceMethods().size());
```
kezhenxu94, line 49:
```suggestion
        final List<InstanceMethodsInterceptPoint> points = new ArrayList<>(this.enhanceMethods().size());
```

kezhenxu94 (changes requested):
Apart from the comments inline, I'd also like to discuss whether we should also include part of the JSON string / bytes in the trace, for example, collect the first 64 bytes as a tag.

Currently we only have the time cost in (de)serializing and the length of the content, but it makes little sense that we don't know what the content is when the time cost is too long, including part of the content might give the users some hints what content makes the JSON lib slower.

apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/define/AbstractInstrumentation.java
wu-sheng: Removing this whole sentence and keeping the later one should be enough.

VictorZeng: OK, I will remove it.

apm-sniffer/optional-plugins/jackson-2.x-plugin/pom.xml
VictorZeng: OK, I will remove it.

apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/BasicMethodsInterceptor.java
VictorZeng: OK, I will format it.

apm-sniffer/optional-plugins/jackson-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jackson/define/AbstractInstrumentation.java
VictorZeng: I remember the `List.toArray(T[] a)`  recommended to give the array size. Because it won't recreate an array.

```java
    public <T> T[] toArray(T[] a) {
        if (a.length < size)
            // Make a new array of a's runtime type, but my contents:
            return (T[]) Arrays.copyOf(elementData, size, a.getClass());
        System.arraycopy(elementData, 0, a, 0, size);
        if (a.length > size)
            a[size] = null;
        return a;
    }
```

VictorZeng: Since openJDK 6, recommended use the null array for this method. I will fix it.

wu-sheng:
> Apart from the comments inline, I'd also like to discuss whether we should also include part of the JSON string / bytes in the trace, for example, collect the first 64 bytes as a tag `content` and the `64` is configurable.
> 
> Currently we only have the time cost in (de)serializing and the length of the content, but it makes little sense that we don't know what the content is when the time cost is too long, including part of the content might give the users some hints what content makes the JSON lib slower.

Agree, collecting content is a risk. I prefer we don't do this.

VictorZeng:
> > Apart from the comments inline, I'd also like to discuss whether we should also include part of the JSON string / bytes in the trace, for example, collect the first 64 bytes as a tag `content` and the `64` is configurable.
> > Currently we only have the time cost in (de)serializing and the length of the content, but it makes little sense that we don't know what the content is when the time cost is too long, including part of the content might give the users some hints what content makes the JSON lib slower.
> 
> Agree, collecting content is a risk. I prefer we don't do this.

I think that providing JSON length can satisfy most user scenarios. It is not recommended to collect JSON content.

VictorZeng:
@kezhenxu94 use the guava `ImmutableMap` is a good idea, but failed the checkstyle.

kezhenxu94:
> @kezhenxu94 use the guava `ImmutableMap` is a good idea, but failed the checkstyle.

ok. We need some adjustments of the check style settings

VictorZeng:
@kezhenxu94 use the `ImmutableMap `, must package the guava dependency with plugin jar together. otherwise:
`Caused by: java.lang.ClassNotFoundException: Can't find com.google.common.collect.ImmutableMap`

Is it necessary to do this?

kezhenxu94:
> @kezhenxu94 use the `ImmutableMap `, must package the guava dependency with plugin jar together. otherwise:
> 
> `Caused by: java.lang.ClassNotFoundException: Can't find com.google.common.collect.ImmutableMap`
> 
> 
> 
> Is it necessary to do this?

I think guava is bundled with agent core already

VictorZeng:
> > @kezhenxu94 use the `ImmutableMap `, must package the guava dependency with plugin jar together. otherwise:
> > `Caused by: java.lang.ClassNotFoundException: Can't find com.google.common.collect.ImmutableMap`
> > Is it necessary to do this?
> 
> I think guava is bundled with agent core already

It doesn't look like this now.

wu-sheng:
Using this 3rd party APIs, AFAIK, you need to shade. This dependency is there, but indirectly(gRPC depending on it) and shaded in the runtime. 
Do we really need to make things this complex? Initializing `ENHANCE_METHODS` in constructor and use it should be good enough, right?

VictorZeng:
> Using this 3rd party APIs, AFAIK, you need to shade. This dependency is there, but indirectly(gRPC depending on it) and shaded in the runtime.
> Do we really need to make things this complex? Initializing `ENHANCE_METHODS` in constructor and use it should be good enough, right?

I noticed this, the source code shaded in `org.apache.skywalking.apm.dependencies.*`
I think init the `ENHANCE_METHODS` in constructor is a good idea.

wu-sheng:
Plugin default class loader is singleton and wouldn't be GC. Any new class file loaded will be a load for JVM memory permanently. Let's try to keep plugin codes as simple as possible.

VictorZeng:
> Plugin default class loader is singleton and wouldn't be GC. Any new class file loaded will be a load for JVM memory permanently. Let's try to keep plugin codes as simple as possible.

Use guava `ImmutableMap` is a better way, I think I can try to shade it.

kezhenxu94:
> > @kezhenxu94 use the `ImmutableMap `, must package the guava dependency with plugin jar together. otherwise:
> > `Caused by: java.lang.ClassNotFoundException: Can't find com.google.common.collect.ImmutableMap`
> > Is it necessary to do this?
> 
> I think guava is bundled with agent core already

OK I remember it wrong, what is bundled is GSON dependency. But the idea is not to create too many anonymous inner classes, initializing the map in constructor is also good to me

VictorZeng:
@wu-sheng @kezhenxu94 How about init `ENHANCE_METHODS ` in static method block?

kezhenxu94:
> @wu-sheng @kezhenxu94 How about init `ENHANCE_METHODS ` in static method block?

Okay to me

VictorZeng:
> > @wu-sheng @kezhenxu94 How about init `ENHANCE_METHODS ` in static method block?
> 
> Okay to me

Thank u.

VictorZeng:
@kezhenxu94 `elasticjob-3.x-scenario` health check failed, it runs successfully on my local.
```
+ status=0
+ [[ 0 == 0 ]]
+ [[ -z '' ]]
+ rm -rf /home/skywalking-java/test/plugin/workspace/elasticjob-3.x-scenario/3.0.0
+ num_of_testcases=1
+ echo -e '\033[33melasticjob-3.x-scenario has already sumbitted\033[0m'
elasticjob-3.x-scenario has already sumbitted
+ exitAndClean 0
++ date +%s
+ elapsed=404
+ [[ 0 -eq 1 ]]
+ printf 'Scenarios: elasticjob-3.x-scenario, Testcases: 1, Elapsed: %02d:%02d:%02d \n' 0 6 44
Scenarios: elasticjob-3.x-scenario, Testcases: 1, Elapsed: 00:06:44 
+ exit 0
```

kezhenxu94 (approved):
Thank you @VictorZeng . One nit: it seems that you didn't import our code style settings, the new files still don't respect to the code styles, please consider import our code style settings if you are going to contribute later

VictorZeng:
> Thank you @VictorZeng . One nit: it seems that you didn't import our code style settings, the new files still don't respect to the code styles, please consider import our code style settings if you are going to contribute later

Thank you for review, I will import the checkstyle.

PR #40, opened 2021-09-29
The httpasyncclient-4.x-plugin does not take effect every time.

heihaozi:
### Fix apache/skywalking#5653

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes apache/skywalking#5653.


wu-sheng:
The test seems failure, please recheck.

wu-sheng (approved):

PR #43, opened 2021-10-07
feature: support xxl-job version 2.3.0

xiaoxiongmao-liu:


wu-sheng:
> > You need to add test to that version. Also, in the new version if the argument is removed, how to get the job name?
> 
> 
> 
> Actually,I have  test to 2.3.0 and 2.2.0. The removed argument is 'JobParam',not the job name and just as a tag, useless.

Why job param useless? Even though, I am not using java agent or this plugin, I feel this parameter matters about how this job gets triggered.

xiaoxiongmao-liu:
> Then this is typical case you should use witness class or method to identify versions, then use different interceptor and instructions for different versions. Then the core could choose the right plugin in the runtime.

image
Actually, according to "allArguments.length" can identify version,2.2.0 work as before

xiaoxiongmao-liu:
> Then, you should fill the logic for 2.3.0 too. Let's keep them consistent

OK, I will do it

xiaoxiongmao-liu:
> @xiaoxiongmao-liu Are you still working on this?

Yes, I will do it as soon as possible.But I am busy my official work.

PR #50, opened 2021-10-12
Fix version compatibility for JSON-RPC4J Plugin

lujiajing1126:
Signed-off-by: Megrez Lu <lujiajing1126@gmail.com>

Background: This plugin has been introduced by @gy09535 in the PR https://github.com/apache/skywalking/pull/6743. He was working in our team at that time and this plugin was extracted from our private codebase. Recently I've noticed a version incompatibility with early versions of jsonrpc4j, for example, all versions `<1.5.0`.

The root cause of this bug is that we actually impose much too strong constraint on the constructor matcher,

```java
@Override
public ConstructorInterceptPoint[] getConstructorsInterceptPoints() {
    return new ConstructorInterceptPoint[]{
            new ConstructorInterceptPoint() {
                @Override
                public ElementMatcher<MethodDescription> getConstructorMatcher() {
                    return ElementMatchers.takesArguments(6).and(ElementMatchers.takesArgument(1, URL.class));
                }

                @Override
                public String getConstructorInterceptor() {
                    return INTERCEPTOR_CLASS;
                }
            },
            new ConstructorInterceptPoint() {
                @Override
                public ElementMatcher<MethodDescription> getConstructorMatcher() {
                    return ElementMatchers.takesArguments(5).and(ElementMatchers.takesArgument(1, URL.class));
                }

                @Override
                public String getConstructorInterceptor() {
                    return INTERCEPTOR_CLASS;
                }
            }
    };
}
```

neither of the constructors exists in versions below `1.5.0`.

Also, I suppose there is problem with test configuration: we have to use `<test.framework.version>`.

### Fix backwards compatibility for jsonrpc4j plugin
- [x] Reproduce this issue in CI.


wu-sheng:
> Recently I've noticed a version incompatibility with early versions of jsonrpc4j, for example, all versions <1.5.0.

The test case fails after you corrected the test codes for `1.4.6`. 

@lujiajing1126 Are you going to continue on this PR to fix this bug?

PR #55, opened 2021-10-24
add test case plugin on jdk17

dmsolr:
- [ x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/7942


wu-sheng:
Please update the document.

wu-sheng:
MySQL case fails due to codes runtime exception. It seems having JDK11 in the runtime, which from reviewing codes it should be JDK8. Please recheck.

https://stackoverflow.com/questions/67332909/why-can-java-not-connect-to-mysql-5-7-after-the-latest-jdk-update-and-how-should

docs/en/setup/service-agent/java-agent/Plugin-test.md
```diff
@@ -20,17 +20,18 @@ The test framework provides `JVM-container` and `Tomcat-container` base images i
 
 ### JVM-container Image Introduction
 
-[JVM-container](../../../../../test/plugin/containers/jvm-container) uses `openjdk:8` as the base image. `JVM-container` supports JDK14, which inherits `openjdk:14`.
+[JVM-container](../../../../../test/plugin/containers/jvm-container) uses `openjdk:8-buster` as the base image. `JVM-container` supports JDK14 and JDK17, which inherits `openjdk:14-buster` and `openjdk:17-buster`.
```
wu-sheng, line 23: What is buster suffix?

test/plugin/containers/jvm-container/pom.xml
```diff
@@ -32,95 +32,44 @@
...
+            <plugin>
+                <groupId>io.fabric8</groupId>
+                <artifactId>docker-maven-plugin</artifactId>
+                <version>0.33.0</version>
+                <configuration>
+                    <images>
+                        <image>
+                            <name>skywalking/agent-test-jvm:jdk${java_major_version}-${project.version}</name>
+                            <build>
+                                <from>openjdk:${java_major_version}-buster</from>
```
wu-sheng, line 46: Same here.

wu-sheng (dismissed):
If we need to change test JDK to `alpine`(only if necessary), we could do later. 
Personally, using `buster` should be fine.

.github/workflows/plugins-jdk14-test.0.yaml
```diff
@@ -62,11 +62,13 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v2
         with:
-          java-version: 14
+          # temurin does not distribute non-lts version
+          distribution: 'zulu'
```
kezhenxu94, line 68: Why do you use these distribution instead of those that we are using in our official Docker images?

.github/workflows/plugins-test.0.yaml
```diff
@@ -90,9 +90,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v2
         with:
-          java-version: 8
+          distribution: 'temurin'
```
kezhenxu94, line 95: And here, we only use temurin distribution when using Java 17

.github/workflows/plugins-test.1.yaml
```diff
@@ -81,9 +81,10 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - uses: actions/setup-java@v1
+      - uses: actions/setup-java@v2
         with:
-          java-version: 8
+          distribution: 'temurin'
```
kezhenxu94, line 86: Here

test/plugin/containers/jvm-container/pom.xml
kezhenxu94: Please use the same base images as what we are using in our official Docker images

https://github.com/apache/skywalking-java/blob/02d6b9c2d85da2b1b1ce3e2a8013373e0e371ba1/Makefile#L44-L51

kezhenxu94 (commented):
Please use the same base images as what we are using in our official Docker images

.github/workflows/plugins-test.0.yaml
wu-sheng: temurin-8 is another distribution of OpenJDK-8, after joining the Eclipse foundation, I think.

test/plugin/containers/jvm-container/pom.xml
wu-sheng: @dmsolr told me there are some error in script in alpine, and our previous testing is on the top of this version due to the limitation of our scripts

wu-sheng: image

I think we were using this. We tried openjdk:alpine here IIRC, the tests would fail

wu-sheng (approved):

PR #62, opened 2021-10-31
[Agent v9 core] Add support for Java Extension Class Loader. 

wu-sheng:
### <Feature description>
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #https://github.com/apache/skywalking/issues/3661.

To @apache/skywalking-committers 

I am planning to bump up the version of Java agent after this gets merged. 9.0.0 would be the next major version to indicate we added Extension ClassLoader support in the core level.
**First and most importantly, this credit and contribution belongs to @daimingzhi. I am just finishing this PR with a little conflict resolving, polishing, and fix a little error in the plugin test.**

## Extension ClassLoader
**Extension ClassLoader**: The Extension ClassLoader is a child of Bootstrap ClassLoader and loads the extensions of core java classes from the respective JDK Extension library.
It loads files from jre/lib/ext directory or any other directory pointed by the system property java.ext.dirs.
User decides which parts of the service should be loaded by the extension classloader, so, we open `plugin.plugins_in_ext_class_loader=${SW_PLUGINS_IN_EXT_CLASS_LOADER:}` in the `agent.config` file.

Through this setting, users declare plugins are for excClassloader. Multiple values should be separated by ",".
Also support wildcard(`\*`),like `ehcache\*`. 

___
This PR originally started at https://github.com/apache/skywalking-java/pull/19, @daimingzhi and I have a long discussion about this topic there.


wu-sheng:
To reviewers(@apache/skywalking-committers), when you merge this, please make sure we have @daimingzhi listed as a co-author of this PR. Because of `squash and merge policy`, this PR will belong to my account I am afraid. Let's keep the track correctly and respect our contributor.

wu-sheng:
I noticed precious plugin test codes didn't verify this feature. Because Spring classloader includes target jars in the Spring fatjar, which means the httpclient jars are not loaded through ext-classloader.
Let me fix this test case.

wu-sheng:
An update, `-Djava.ext.dirs` is obsolete

wu-sheng:
@daimingzhi According to JDK 11 is widely used and recommended to use. JDK8 would be the only LTS JDK.

JDK8 EOF update
1. After January of 2019, only customers who have signed a support contract with Oracle will still receive security updates for Oracle JDK 8.
2. Maintenance for OpenJDK 8 is planned for at least another 4 years, until September 2023. 2 years left for EOL.

So, generally, according to the roadmap of JDK, `java.ext.dirs` should be replaced by `cp` or `classpath`. I will close this PR with original one as well. But I will add a FAQ doc to say what is going on, if user wants to try this draft feature, feel free to download and try.

___
Same as this reason, I would NOT bump up agent version to v9. Let's stay in v8 for longer.

PR #65, opened 2021-11-02
Build CLI from Docker instead of source codes, add alpine and jdk17 based Docker image

kezhenxu94:


docs/en/contribution/release-java-agent.md
```diff
@@ -276,7 +276,7 @@ export NAME=skywalking-java-agent
 export HUB=apache
 export TAG=$SW_VERSION
 
-make docker.push.java8 docker.push.java11
+make docker.push.alpine docker.push.java8 docker.push.java11
```
wu-sheng, line 279:
```suggestion
make docker.push.alpine docker.push.java8 docker.push.java11 docker.push.java17
```
How about this? As 17 is already LTS.

kezhenxu94: Sure, done

wu-sheng (approved):
I noticed 8.8.0-alphine is updated, let's make *-jdk17 available too.

https://hub.docker.com/r/apache/skywalking-java-agent/tags

kezhenxu94:
> I noticed 8.8.0-alphine is updated, let's make *-jdk17 available too.
> 
> https://hub.docker.com/r/apache/skywalking-java-agent/tags

Pushed

PR #67, opened 2021-11-07
Support set instance properties in json format

JaredTan95:
### Support set instance properties in json format: `agent.instance_properties_json`

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/7762.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/InstanceJsonPropertiesUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/InstanceJsonPropertiesUtil.java
@@ -19,6 +19,7 @@
 package org.apache.skywalking.apm.agent.core.util;
 
 import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -33,7 +34,11 @@ public class InstanceJsonPropertiesUtil {
         List<KeyStringValuePair> properties = new ArrayList<>();
 
         if (StringUtil.isNotEmpty(Config.Agent.INSTANCE_PROPERTIES_JSON)) {
-            Config.Agent.INSTANCE_PROPERTIES.putAll(GSON.fromJson(Config.Agent.INSTANCE_PROPERTIES_JSON, Map.class));
+            Config.Agent.INSTANCE_PROPERTIES.putAll(
+                GSON.fromJson(
+                    Config.Agent.INSTANCE_PROPERTIES_JSON,
+                    new TypeToken<Map<String, Object>>() {}.getType()
+                ));
         }
 
         for (String key : Config.Agent.INSTANCE_PROPERTIES.keySet()) {
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/InstanceJsonPropertiesUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/InstanceJsonPropertiesUtil.java
@@ -41,12 +41,10 @@ public class InstanceJsonPropertiesUtil {
                 ));
         }
 
-        for (String key : Config.Agent.INSTANCE_PROPERTIES.keySet()) {
-            properties.add(KeyStringValuePair.newBuilder()
-                                             .setKey(key)
-                                             .setValue(Config.Agent.INSTANCE_PROPERTIES.get(key))
-                                             .build());
-        }
+        Agent.INSTANCE_PROPERTIES.forEach((key, val) -> properties.add(KeyStringValuePair.newBuilder()
+                                                                                         .setKey(key)
+                                                                                         .setValue(val)
+                                                                                         .build()));
         return properties;
     }
 }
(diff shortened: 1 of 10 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/InstanceJsonPropertiesUtil.java
```diff
@@ -0,0 +1,56 @@
...
+
+        for (String key : Config.Agent.INSTANCE_PROPERTIES.keySet()) {
+            properties.add(KeyStringValuePair.newBuilder()
+                                             .setKey(key)
+                                             .setValue(Config.Agent.INSTANCE_PROPERTIES.get(key))
+                                             .build());
+        }
+
+        if (StringUtil.isNotEmpty(Config.Agent.INSTANCE_PROPERTIES_JSON)) {
+            Map<String, Object> jsonProperties = GSON.fromJson(Config.Agent.INSTANCE_PROPERTIES_JSON, Map.class);
```
wu-sheng, line 43: I think it is easier to append `jsonProperties` into `Config.Agent.INSTANCE_PROPERTIES` first, and then convert to `List<KeyStringValuePair> properties`, in order to avoid `removeId` + `add`.

wu-sheng:
Recommend to use `INSTANCE_PROPERTIES_JSON` into e2e case, and verify this feature. I think it is easy to add now.

apm-sniffer/config/agent.config
```diff
@@ -20,6 +20,8 @@ agent.namespace=${SW_AGENT_NAMESPACE:}
 # The service name in UI
 agent.service_name=${SW_AGENT_NAME:Your_ApplicationName}
 
+agent.instance_properties_json=${SW_INSTANCE_PROPERTIES_JSON:}
```
kezhenxu94, line 23: Can you please put this in the same order as the doc (move this to the next line of `agent.instance_name`)

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/InstanceJsonPropertiesUtil.java
JaredTan95: Yes, maybe that would be more cleaner. I will update it.

kezhenxu94: Also consider avoid the warning by using

```suggestion
            Map<String, Object> jsonProperties = GSON.fromJson(Config.Agent.INSTANCE_PROPERTIES_JSON, new TypeToken<Map<String, Object>>(){}.getType());
```

test/e2e/case/expected/service-instance.yml
```diff
@@ -19,7 +19,10 @@
 - id: {{ b64enc "e2e-service-provider" }}.1_{{ b64enc "provider1" }}
   name: {{ notEmpty .name }}
   attributes:
-  {{- contains .attributes }}
+    {{- contains .attributes }}
+    - name: org
+      value: apache-skywalking
+    {{- end }}
```
kezhenxu94, line 25:
```suggestion
  {{- contains .attributes }}
  - name: org
    value: apache-skywalking
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/InstanceJsonPropertiesUtil.java
```diff
@@ -0,0 +1,47 @@
...
+        if (StringUtil.isNotEmpty(Config.Agent.INSTANCE_PROPERTIES_JSON)) {
+            Config.Agent.INSTANCE_PROPERTIES.putAll(GSON.fromJson(Config.Agent.INSTANCE_PROPERTIES_JSON, Map.class));
+        }
+
+        for (String key : Config.Agent.INSTANCE_PROPERTIES.keySet()) {
+            properties.add(KeyStringValuePair.newBuilder()
+                                             .setKey(key)
+                                             .setValue(Config.Agent.INSTANCE_PROPERTIES.get(key))
+                                             .build());
+        }
```
kezhenxu94, line 49: No need to use `keySet()` and `get`, use `entrySet` + `getKey` + `getValue` or even use `forEach` instead

```suggestion
        Agent.INSTANCE_PROPERTIES.forEach((key, val) -> properties.add(KeyStringValuePair.newBuilder()
                                                                                         .setKey(key)
                                                                                         .setValue(val)
                                                                                         .build()));
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/InstanceJsonPropertiesUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.skywalking.apm.util.StringUtil;
+
+public class InstanceJsonPropertiesUtil {
+    private static final Gson GSON = new Gson();
+
+    public static List<KeyStringValuePair> parseProperties() {
+        List<KeyStringValuePair> properties = new ArrayList<>();
+
+        if (StringUtil.isNotEmpty(Config.Agent.INSTANCE_PROPERTIES_JSON)) {
+            Config.Agent.INSTANCE_PROPERTIES.putAll(GSON.fromJson(Config.Agent.INSTANCE_PROPERTIES_JSON, Map.class));
```
kezhenxu94:
```suggestion
                        Config.Agent.INSTANCE_PROPERTIES.putAll(GSON.fromJson(Config.Agent.INSTANCE_PROPERTIES_JSON, new TypeToken<Map<String, Object>>(){}.getType()));
```

kezhenxu94 (approved):

kezhenxu94 (approved):
LGTM once tests passed

PR #72, opened 2021-11-18
TracePathMatcher should match pattern "**" with paths end by "/" (#7875)

newboy2004:
### Fix <TracePathMatcher should match pattern "**" with paths end by "/">
- [ ]  Use "Spring AntPathMatcher'   instead of  ”FastPathMatcher“


apm-sniffer/optional-plugins/trace-ignore-plugin/pom.xml
```diff
@@ -41,5 +42,11 @@
             <version>${ststem-rules.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-core</artifactId>
```
wu-sheng, line 48: We can't accept Spring core, sorry. Because we have to include all codes inside the agent plugin.

wu-sheng (changes requested):

wu-sheng:
If you want to add this, you need to enhance the existing one, rather than use Spring to replace.
We knew Spring can do from beginning, but we determined this is not suitable, due to package size and License risk.

wu-sheng:
@devkanro Could you recheck this? I think you wrote this originally.

devkanro:
ok

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
```diff
@@ -124,7 +124,7 @@ private boolean wildcardMatch(String pat, int p, String str, int s) {
     private boolean multiWildcardMatch(String pat, int p, String str, int s) {
         // End of pattern, just check the end of string is '/' quickly.
         if (p >= pat.length() && s < str.length()) {
-            return str.charAt(str.length() - 1) != '/';
+            return true;
         }
```
devkanro, line 128: Maybe the comment need be changed.

```Java
        // End of pattern, make matching success quickly.
        if (p >= pat.length() && s < str.length()) {
            return true;
        }
```

devkanro (changes requested):

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
devkanro: In my originally design, the `/eureka/**` not match the `/eureka/client/` is a feature, but I have no preference for this, I can accept it regardless of whether it matches or does not match.

Could you show your use cases and benefits for it?

wu-sheng: I think because some users think all things under `/eureka/` belong to this match rule.

wu-sheng: @devkanro Could you share a little more about why `/eureka/client/` doesn't belong to `/eureka/**`? I hope we could have a fully evaluation considering this was being asked more than once.(This is the first fix)

devkanro: Ummm...I can't remember what I thought at the time...

Maybe it's because `/eureka/` doesn't match `/eureka`, I want to make it same as `/eureka/**` case?

I have changed my mind, the `/eureka/client/` should belong to `/eureka/**`.

LGTM

devkanro: And maybe the rule of `wildcardMatch` also should be changed?
Should `/eureka/*` match with `/eureka/client/` ?

wu-sheng: >/eureka/ doesn't match /eureka

I think they should be same, do you know any different in some use cases?
I feel they are same. 

> And maybe the rule of wildcardMatch also should be changed?
Should /eureka/* match with /eureka/client/ ?


Yes, it should be supported, too.

wu-sheng:
@newboy2004 Please follow the @devkanro 's comments to polish/enhance this fix and update the changes.md in the root.

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
newboy2004: what is final dicussion?

wu-sheng: The current resolution is, what your proposal is correct. Just follow the polish requirements.

newboy2004:
> @newboy2004 Please follow the @devkanro 's comments to polish/enhance this fix and update the changes.md in the root.

ok,i continue repair it.

By the way, accessing githubcom is slow. Is there any good solution？

devkanro:
The simple resolution is remove the last '/' for the pattern and string.

If this change is done during the matching process, many states will be introduced, which will complicate the state machine of the matcher. 

I think it is a good choice to normalize(remove the last '/') the pattern and value before entering the matcher.

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
devkanro: @wu-sheng 
> /eureka/ doesn't match /eureka
Ummm, The reason I do this is to simplify the state machine. There are no other special considerations.

wu-sheng:
Agree, this kind of change seems better and easier to understand.

newboy2004:
> Agree, this kind of change seems better and easier to understand.

Continue to repair * matching, and remove the last / of pattern and string before?

wu-sheng:
Remove / before the match, I think. Then no need to change matching core, fromy understanding.
Please correct me if I am wrong.

newboy2004:
> Remove / before the match, I think. Then no need to change matching core, fromy understanding. Please correct me if I am wrong.

please see blow  origin  code:
 String patten = "/eureka/*";
  path = "/eureka/apps/";
 match = pathMatcher.match(patten, path);
 Assert.assertFalse(match);

i think remove / after,Assert  should return true. 
please check my opinion,is it right?

wu-sheng:
> i think remove / after,Assert should return true.
please check my opinion,is it right?

Yes, `/eureka/apps/` should be captured by `/eureka/*` expression. That is why @devkanro proposed a change about removing `/` in `/eureka/apps/`. Then the current FastPathMatcher should return true, too. Right?

newboy2004:
> > i think remove / after,Assert should return true.
> > please check my opinion,is it right?
> 
> Yes, `/eureka/apps/` should be captured by `/eureka/*` expression. That is why @devkanro proposed a change about removing `/` in `/eureka/apps/`. Then the current FastPathMatcher should return true, too. Right?

i wait for final opinion?  my local code repo have already finish what *  rule match and unit test  according before discussion

waiting for you  final opionion

wu-sheng:
I think the conclusion is very clear. The recommended way to fix is removing the last `/` in the operation name, before running `match`. Such as in `TraceIgnoreExtendService#trySampling`. What do you think? @newboy2004

newboy2004:
> conclusion

ok，i  understand.

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
```diff
@@ -83,6 +90,10 @@ private boolean normalMatch(String pat, int p, String str, int s) {
 
     private boolean wildcardMatch(String pat, int p, String str, int s) {
         char pc = safeCharAt(pat, p);
+        //if pat already arrival end,and str in position of s is not '/',then return true
+        if (pc == '\u0000' && safeCharAt(str, s) != '/') {
+            return true;
```
devkanro, line 95: I think this change will cause a bug that 'abc/*' matching 'abc/foo/bar'

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
```diff
@@ -124,7 +135,7 @@ private boolean wildcardMatch(String pat, int p, String str, int s) {
     private boolean multiWildcardMatch(String pat, int p, String str, int s) {
         // End of pattern, just check the end of string is '/' quickly.
         if (p >= pat.length() && s < str.length()) {
-            return str.charAt(str.length() - 1) != '/';
+            return true;
```
devkanro, line 138: I think we can remove this check after we remove the last '/' both of pattern and value,

devkanro (changes requested):

wu-sheng:
> I think the conclusion is very clear. The recommended way to fix is removing the last `/` in the operation name, before running `match`. Such as in `TraceIgnoreExtendService#trySampling`. What do you think? @newboy2004

I think the current doesn't follow this conclusion. We just need to slightly change the input.

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
newboy2004: ok

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
newboy2004: i review it

newboy2004:
> > I think the conclusion is very clear. The recommended way to fix is removing the last `/` in the operation name, before running `match`. Such as in `TraceIgnoreExtendService#trySampling`. What do you think? @newboy2004
> 
> I think the current doesn't follow this conclusion. We just need to slightly change the input.

ok,but if i remove the last '/' in the operation name,please confirm blow code,return true or false:
String patten = "/eureka/*";
path = "/eureka/";
match = pathMatcher.match(patten, path);
Assert.assertTrue(match);

i think  '*' match zero or one char,but remove the last '/' after,pattern='/eureka/*' path='/eureka',accroding Ant Rule,result should return false,do you think?

wu-sheng:
OK, I can see this becomes a tricky point. From this point of view, removing `/` and changing matching rules are both not a very good idea, as these kinds of edge conditions. 

I think we should consider whether changing plugin's operation rules fits more cases.

newboy2004:
> edge

We can consider not supporting this so-called bug，so I'll consider claiming other tasks?haha

wu-sheng:
All operation names, such as typically `/eureka/` is collected by a HTTP client plugin. If that plugin says removing a `/` as suffix, I think it is generally easy.

newboy2004:
> All operation names, such as typically `/eureka/` is collected by a HTTP client plugin. If that plugin says removing a `/` as suffix, I think it is generally easy.

The code test mentioned above should return false，please @devkanro  check

wu-sheng:
> > All operation names, such as typically `/eureka/` is collected by a HTTP client plugin. If that plugin says removing a `/` as suffix, I think it is generally easy.
> 
> The code test mentioned above should return false，please @devkanro check

The point here is, `returning false` breaks the expectation when this PR raised.

newboy2004:
> > > All operation names, such as typically `/eureka/` is collected by a HTTP client plugin. If that plugin says removing a `/` as suffix, I think it is generally easy.
> > 
> > 
> > The code test mentioned above should return false，please @devkanro check
> 
> The point here is, `returning false` breaks the expectation when this PR raised.

so  remvoing the last '/' suffix, it's not a good idea, it breaks the expectation.
or is there a logical problem with the modification of the code after the removal / modification

wu-sheng:
Removing suffix is not a good idea. And changing logic makes me concerns about new bugs.
After all, this is just an enhancement for this kind of case.

devkanro:
We should rearrange the rules so that we can analyze the edge conditions. 

Basic:
1. `?` matches any one char.
2. `*` matches one path part.
3. `**` matches one or more path part.

Edge(for *):
1. `/eureka/*` matches `/eureka/`?
2. `/eureka/*` matches `/eureka/test/`?
3. `/eureka/*/` matches `/eureka/`?
4. `/eureka/*/` matches `/eureka/test`?
5. Is `/eureka/*` same as `/eureka/*/`?

Edge(for **):
1. `/eureka/**` matches `/eureka/`?
2. `/eureka/**` matches `/eureka/test/`?
3. `/eureka/**/` matches `/eureka/`?
4. `/eureka/**/` matches `/eureka/test`?
5. Is `/eureka/**` same as `/eureka/**/`?

Edge(for normal):
1. `/eureka/` matches `/eureka`?
2. `/eureka` matches `/eureka/`?

devkanro:
My answer(all false for current behavior):

Edge(for *):
1. [x] `/eureka/*` matches `/eureka/`?
2. [x] `/eureka/*` matches `/eureka/test/`?
3. [ ] `/eureka/*/` matches `/eureka/`?
4. [ ] `/eureka/*/` matches `/eureka/test`?
5. [ ] Is `/eureka/*` same as `/eureka/*/`?

Edge(for **):
1. [x] `/eureka/**` matches `/eureka/`?
2. [x] `/eureka/**` matches `/eureka/test/`?
3. [ ] `/eureka/**/` matches `/eureka/`?
4. [ ] `/eureka/**/` matches `/eureka/test`?
5. [ ] Is `/eureka/**` same as `/eureka/**/`?

Edge(for normal):
1. [ ] `/eureka/` matches `/eureka`?
2. [x] `/eureka` matches `/eureka/`?

wu-sheng:
No update in 2 weeks.

PR #73, opened 2021-11-23
add plugin for dubbo3

tedli:
### Add an agent plugin to support [dubbo3](https://github.com/apache/dubbo)
- [ ] Add a component id in [the component-libraries.yml](https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/component-libraries.yml)  `There is already one (for dubbo 2.5 and 2.7)`
- [ ] Add a logo in [the UI repo](https://github.com/apache/skywalking-rocketbot-ui/tree/master/src/views/components/topology/assets)  `There is already one (for dubbo 2.5 and 2.7)`

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<issue number>.    [skywalking 7203](https://github.com/apache/skywalking/issues/7203)

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/dubbo-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/DubboInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/dubbo-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/DubboInterceptor.java
@@ -1,0 +1,217 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.asf.dubbo3;
+
+import org.apache.dubbo.common.constants.CommonConstants;
+import org.apache.dubbo.common.URL;
+import org.apache.dubbo.rpc.Invocation;
+import org.apache.dubbo.rpc.Invoker;
+import org.apache.dubbo.rpc.Result;
+import org.apache.dubbo.rpc.RpcContext;
+import org.apache.dubbo.rpc.RpcContextAttachment;
+import org.apache.dubbo.rpc.RpcException;
+import org.apache.dubbo.rpc.RpcServiceContext;
+import org.apache.skywalking.apm.agent.core.context.CarrierItem;
+import org.apache.skywalking.apm.agent.core.context.ContextCarrier;
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.context.tag.Tags;
+import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
+import org.apache.skywalking.apm.agent.core.context.trace.SpanLayer;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
+import org.apache.skywalking.apm.util.StringUtil;
+
+import java.lang.reflect.Method;
+
+/**
+ * {@link DubboInterceptor} define how to enhance class {@link org.apache.dubbo.monitor.support.MonitorFilter#invoke(Invoker,
+ * Invocation)}. the trace context transport to the provider side by {@link RpcContext#getServerAttachment}.but all the
+ * version of dubbo framework below 3.0.0 don't support {@link RpcContext#getClientAttachment}, we support another way
+ * to support it.
+ */
+public class DubboInterceptor implements InstanceMethodsAroundInterceptor {
+
+    public static final String ARGUMENTS = "arguments";
+
+    /**
+     * <h2>Consumer:</h2> The serialized trace context data will
+     * inject to the {@link RpcContext#getClientAttachment} for transport to provider side.
+     * <p>
+     * <h2>Provider:</h2> The serialized trace context data will extract from
+     * {@link RpcContext#getServerAttachment}. current trace segment will ref if the serialize context data is not
+     * null.
+     */
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) throws Throwable {
+        Invoker invoker = (Invoker) allArguments[0];
+        Invocation invocation = (Invocation) allArguments[1];
+
+        RpcServiceContext serviceContext = RpcContext.getServiceContext();
+        boolean isConsumer;
+        URL url = serviceContext.getUrl();
+        if (url == null) {
+            url = serviceContext.getConsumerUrl();
+            isConsumer = url != null;
+        } else {
+            isConsumer = serviceContext.isConsumerSide();
+        }
+
+        RpcContextAttachment attachment = isConsumer ? RpcContext.getClientAttachment() : RpcContext.getServerAttachment();
+        URL requestURL = invoker.getUrl();
+
+        AbstractSpan span;
+
+        final String host = requestURL.getHost();
+        final int port = requestURL.getPort();
+
+        boolean needCollectArguments;
+        int argumentsLengthThreshold;
+        if (isConsumer) {
+            final ContextCarrier contextCarrier = new ContextCarrier();
+            span = ContextManager.createExitSpan(
+                generateOperationName(requestURL, invocation), contextCarrier, host + ":" + port);
+            //invocation.getAttachments().put("contextData", contextDataStr);
+            //@see https://github.com/alibaba/dubbo/blob/dubbo-2.5.3/dubbo-rpc/dubbo-rpc-api/src/main/java/com/alibaba/dubbo/rpc/RpcInvocation.java#L154-L161
+            CarrierItem next = contextCarrier.items();
+            while (next.hasNext()) {
+                next = next.next();
+                attachment.setAttachment(next.getHeadKey(), next.getHeadValue());
+                if (invocation.getAttachments().containsKey(next.getHeadKey())) {
+                    invocation.getAttachments().remove(next.getHeadKey());
+                }
+            }
+            needCollectArguments = DubboPluginConfig.Plugin.Dubbo.COLLECT_CONSUMER_ARGUMENTS;
+            argumentsLengthThreshold = DubboPluginConfig.Plugin.Dubbo.CONSUMER_ARGUMENTS_LENGTH_THRESHOLD;
+        } else {
+            ContextCarrier contextCarrier = new ContextCarrier();
+            CarrierItem next = contextCarrier.items();
+            while (next.hasNext()) {
+                next = next.next();
+                next.setHeadValue(attachment.getAttachment(next.getHeadKey()));
+            }
+
+            span = ContextManager.createEntrySpan(generateOperationName(requestURL, invocation), contextCarrier);
+            span.setPeer(attachment.getRemoteAddressString());
+            needCollectArguments = DubboPluginConfig.Plugin.Dubbo.COLLECT_PROVIDER_ARGUMENTS;
+            argumentsLengthThreshold = DubboPluginConfig.Plugin.Dubbo.PROVIDER_ARGUMENTS_LENGTH_THRESHOLD;
+        }
+
+        Tags.URL.set(span, generateRequestURL(requestURL, invocation));
+        collectArguments(needCollectArguments, argumentsLengthThreshold, span, invocation);
+        span.setComponent(ComponentsDefine.DUBBO);
+        SpanLayer.asRPCFramework(span);
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                              Object ret) throws Throwable {
+        Result result = (Result) ret;
+        try {
+            if (result != null && result.getException() != null) {
+                dealException(result.getException());
+            }
+        } catch (RpcException e) {
+            dealException(e);
+        }
+
+        ContextManager.stopSpan();
+        return ret;
+    }
+
+    @Override
+    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments,
+                                      Class<?>[] argumentsTypes, Throwable t) {
+        dealException(t);
+    }
+
+    /**
+     * Log the throwable, which occurs in Dubbo RPC service.
+     */
+    private void dealException(Throwable throwable) {
+        AbstractSpan span = ContextManager.activeSpan();
+        span.log(throwable);
+    }
+
+    /**
+     * Format operation name. e.g. org.apache.skywalking.apm.plugin.test.Test.test(String)
+     *
+     * @return operation name.
+     */
+    private String generateOperationName(URL requestURL, Invocation invocation) {
+        StringBuilder operationName = new StringBuilder();
+        String groupStr = requestURL.getParameter(CommonConstants.GROUP_KEY);
+        groupStr = StringUtil.isEmpty(groupStr) ? "" : groupStr + "/";
+        operationName.append(groupStr);
+        operationName.append(requestURL.getPath());
+        operationName.append("." + invocation.getMethodName() + "(");
+        for (Class<?> classes : invocation.getParameterTypes()) {
+            operationName.append(classes.getSimpleName() + ",");
+        }
+
+        if (invocation.getParameterTypes().length > 0) {
+            operationName.delete(operationName.length() - 1, operationName.length());
+        }
+
+        operationName.append(")");
+
+        return operationName.toString();
+    }
+
+    /**
+     * Format request url. e.g. dubbo://127.0.0.1:20880/org.apache.skywalking.apm.plugin.test.Test.test(String).
+     *
+     * @return request url.
+     */
+    private String generateRequestURL(URL url, Invocation invocation) {
+        StringBuilder requestURL = new StringBuilder();
+        requestURL.append(url.getProtocol() + "://");
+        requestURL.append(url.getHost());
+        requestURL.append(":" + url.getPort() + "/");
+        requestURL.append(generateOperationName(url, invocation));
+        return requestURL.toString();
+    }
+
+    private void collectArguments(boolean needCollectArguments,
+                                  int argumentsLengthThreshold,
+                                  AbstractSpan span,
+                                  Invocation invocation) {
+        if (needCollectArguments && argumentsLengthThreshold > 0) {
+            Object[] parameters = invocation.getArguments();
+            if (parameters != null && parameters.length > 0) {
+                StringBuilder stringBuilder = new StringBuilder();
+                boolean first = true;
+                for (Object parameter : parameters) {
+                    if (!first) {
+                        stringBuilder.append(",");
+                    }
+                    stringBuilder.append(parameter);
+                    if (stringBuilder.length() > argumentsLengthThreshold) {
+                        stringBuilder.append("...");
+                        break;
+                    }
+                    first = false;
+                }
+                span.tag(ARGUMENTS, stringBuilder.toString());
+            }
+        }
+    }
+}
--- a/test/plugin/scenarios/dubbo-3.x-scenario/pom.xml
+++ b/test/plugin/scenarios/dubbo-3.x-scenario/pom.xml
@@ -1,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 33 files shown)


apm-sniffer/apm-sdk-plugin/dubbo-3-conflict-patch/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/patch/WrapperInstrumentation.java
```diff
@@ -0,0 +1,64 @@
...
+                public boolean isOverrideArgs() {
+                    return false;
+                }
+            }
+        };
+    }
+
+    @Override
+    protected ClassMatch enhanceClass() {
+        return byName("org.apache.dubbo.common.bytecode.Wrapper");
```
wu-sheng, line 62: I can see this is as same as Dubbo v2's `WrapperInstrumentation`. Is this duplicated **conflict patch**?

I can't see `dubbo` `v2->v3` from namespace level in `MakeWrapperInterceptor`, and there is no witness class in this instrumentation. Should we just simple change `dubbo-2.7.x-conflict-patch` to `dubbo-2.7.x-3.x-conflict-patch`

apm-sniffer/apm-sdk-plugin/dubbo-3-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/DubboInstrumentation.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.InstanceMethodsInterceptPoint;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInstanceMethodsEnhancePluginDefine;
+import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.NameMatch;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class DubboInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
+
+    private static final String ENHANCE_CLASS = "org.apache.dubbo.monitor.support.MonitorFilter";
```
wu-sheng, line 33: And this is targeting the same class of Dubbo 2. What do I miss here? I thought Dubbo v3 is different from Dubbo v2?

wu-sheng:
A new plugin test(not UT) is required like this, https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/dubbo-2.7.x-scenario.

We have plugin development doc to guide you how to write a plugin test.

apm-sniffer/apm-sdk-plugin/dubbo-3-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/DubboInstrumentation.java
tedli: I'm new to skywalking, I didn't dig for a deeper understanding of skywalking. When I finished current state of this pr, only tested using my dubbo3 demo, and it works.
As you pointed out, I believe it won't work as expect when using dubbo 2.7.
Converted to draft.

apm-sniffer/apm-sdk-plugin/dubbo-3-conflict-patch/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/patch/WrapperInstrumentation.java
tedli: Yes, indeed there is no difference at namespace level between v2 and v3. Actually dubbo 3 even gave an announcement that no need to touch a single line of code just update version from 2.x to 3.x in dependency of pom.xml, and the migration to v3 is done.

apm-sniffer/apm-sdk-plugin/dubbo-3-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/DubboInstrumentation.java
wu-sheng: > As you pointed out, I believe it won't work as expect when using dubbo 2.7.

We have a plugin for Dubbo 2.7, my question is what are the differences between these two. They seems very closing to each other, especially targeting same class(name)?

apm-sniffer/apm-sdk-plugin/dubbo-3-conflict-patch/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/patch/WrapperInstrumentation.java
wu-sheng: Then, could you share why dubbo 2.x plugin doesn't work on dubbo 3.x?

tedli: It because, for classes and methods expose to user side (rpc level) is compatible with v2, but the internal dosen't.
For example we can't rely on `RpcContext.getContext()` with v3.

wu-sheng: OK, then. In theory(you could test and verify), if this patch classes(instrumentation + interceptor) are exactly 100% same as v2, then this is not required.

> for classes and methods expose to user side (rpc level) is compatible with v2, but the internal dosen't.
For example we can't rely on RpcContext.getContext() with v3.

This would make the plugin for Dubbo 3 is required, and you should add `witness class` to dubbo 3 and dubbo 2.7 plugins, because they share the same target, you should make the agent core knows when should activate this instrumentation, and pick the right interceptor in the runtime.

Do you knoow what is the `witness class`? If not, read this, https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/java-plugin-development-guide/#implement-plugin, search this `4. Set up witnessClasses and/or witnessMethods if the instrumentation has to be activated in specific versions.`

apm-sniffer/apm-sdk-plugin/dubbo-3-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/DubboInstrumentation.java
tedli: Differences: (not all, ones currently I know)
1. RpcContext.getContext() didn't work;
2. attachment divided into `ClientAttachment` and ServerAttachment;
3. dubbo constants definition class divided into multiple classes (like `org.apache.dubbo.common.constants.CommonConstants`);
4. methods added in `org.apache.dubbo.common.URL`

apm-sniffer/apm-sdk-plugin/dubbo-3-conflict-patch/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/patch/WrapperInstrumentation.java
wu-sheng: Then follow [this doc](https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/plugin-test/) and take a reference from what I shared for dubbo 2.7 plugin test, you could add a new for dubbo 3.0.

Once both of them pass the test, we should be good.

tedli: Ok, just need some time, won't be too lang.

apm-sniffer/apm-sdk-plugin/dubbo-3-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/DubboInstrumentation.java
wu-sheng: OK then, follow https://github.com/apache/skywalking-java/pull/73#discussion_r754864763, You should be good.

apm-sniffer/apm-sdk-plugin/dubbo-commons/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-deploy-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
```
wu-sheng, line 39: What is this about?

wu-sheng:
The dubbo plugin is not very complex like JDBC, do we really need the common package and set up this complex hierarchy structure? Dubbo v2 will dead(EOL) one day, keeping them separate will make us easier to remove Dubbo 2 plugin and keep 3 plugin w/o change.

tedli:
> > witness classes and methods added, and extract common code of both dubbo 2.7 and 3 into individual maven module.
> 
> The dubbo plugin is not very complex like JDBC, do we really need the common package and set up this complex hierarchy structure? Dubbo v2 will dead(EOL) one day, keeping them separate will make us easier to remove Dubbo 2 plugin and keep 3 plugin w/o change.

Ok, rollback.

**TODO:**
- [ ] add witness to tell is dubbo v2.7 or v3
- [ ] a new plugin test(not UT)

tedli:
> [2021-11-25 07:08:58:225] [ERROR] - org.apache.skywalking.plugin.test.agent.tool.Main.verify(Main.java:53) -
assert failed.
SegmentSizeNotEqualsException:  dubbo-2.7.x-scenario
expected:       ge 3
actual:         2

Current state of `main` branch, I can't get plugin test of dubbo 2.7.x pass. 

```bash
# by using this command
./test/plugin/run.sh -f dubbo-2.7.x-scenario
```

wu-sheng:
Have you passed maven install and both plugin cases locally?

tedli:
> Have you passed maven install and both plugin cases locally?

 Yes, it passed (without this pr). With this pr I can pass other scenarios except dubbo2.7 and 3. So it turns out I broke something.

wu-sheng:
I think it is better you reset anything back to main branch, then add your new plugin implementation with witness class.

tedli:
> I think it is better you reset anything back to main branch, then add your new plugin implementation with witness class.

Yes, I did exactly the same thing. Checked out the main branch, add witness class only to dubbo2.7 and it's conflict patch, do nothing with dubbo3.

Then the 2.7 scenario test broke. However, with these witness class, my 2.7 demo and 3 demo works as expect (the log file in logs folder shows that 2.7 demo dosen't meet 3's witness, and vice versa, and agent can produce spans which appear on skywalking ui as expect).

It needs some more time, I will keep on working on this.

wu-sheng:
Thanks for updating. If there is any help we could provide, please let us know.

tedli:
I got the point why the scenario test failed. It's because the chosen witness classes fit one minor version but not for other minor versions.
The dubbo core classes are stable among various major version like `org.apache.dubbo.rpc.RpcContext`, however it can't be used to tell it's dubbo 2.7.x or 3.
Meanwhile other classes like `org.apache.dubbo.rpc.protocol.rest.NettyServer`, exists in 2.7.x not in 3. But in some newer patch version of 2.7.x, it also not exists.
So it takes more time to find classes which could be used to determine it's 2.7.x or 3.x, and such classes intuitively will stay stable among all the 'x' minor or patch versions.

wu-sheng:
We had a contributed contributing this tool to help you locate the witness, https://github.com/SkyAPM/uranus.

BTW, it doesn't have to be a witness class, a method signature is also works.

wu-sheng:
@tedli Any luck with this plugin?

tedli:
> @tedli Any luck with this plugin?

Hi, the witness methods works fine, by adding witness methods like this:
```java

// dubbo 2.7.x
@Override
protected List<WitnessMethod> witnessMethods() {
    return Collections.singletonList(new WitnessMethod(
            "org.apache.dubbo.rpc.RpcContext",
            named("getServerContext").and(returns(named("org.apache.dubbo.rpc.RpcContext")))));
}

// dubbo 3.x
@Override
protected List<WitnessMethod> witnessMethods() {
    return Collections.singletonList(new WitnessMethod(
            "org.apache.dubbo.rpc.RpcContext",
            named("getServerContext").and(returns(named("org.apache.dubbo.rpc.RpcContextAttachment")))));
}
```

the agent mechanism accurately load the correct plugin. And the scenario test of 2.7.x passed as it is.

How ever the scenario test of 3 still failed, it because `request url` got from `invocation` seemed weird, the produced span didn't match the expect value.

It need some further investigation about how to construct a span for dubbo 3. 

:sob: It's the end of the year, I have to handle some of my own mess, which has higher priority. I'm very glad to continue working on this issue. I will make some time for this.

wu-sheng:
Got it. Make your time, no hurry about this. I just want to check the status, and we will plan to release the next 8.9.0 for agent after we have Dubbo plugin. Because we know people at China love Dubbo project, and would require this plugin.

tedli:
Hi, @wu-sheng 

FYI

https://github.com/apache/dubbo/blob/00ca43643d243f0866389abd294c6160e155c8a5/dubbo-monitor/dubbo-monitor-api/src/main/java/org/apache/dubbo/monitor/support/MonitorFilter.java#L54-L55

https://github.com/apache/dubbo/blob/5a5e3f3fd7d0482b6124bd5b52b10f4900856438/dubbo-monitor/dubbo-monitor-api/src/main/java/org/apache/dubbo/monitor/support/MonitorFilter.java#L52-L53

```java

// dubbo 2.7.x
@Activate(group = {PROVIDER, CONSUMER})
public class MonitorFilter implements Filter, Filter.Listener {
}

// dubbo 3.x
@Activate(group = {PROVIDER})
public class MonitorFilter implements Filter, Filter.Listener {
}

```

After some debugging, I found out why the scenario test of dubbo3 failed, it's because the occasion that `MonitorFilter` takes place is different in dubbo 3.x from 2.7.x, only provider side produced spans. So the test failed.

It need to find other filters to adjust instrumentation class intercepting behavior, so that consumer side can also produce spans.

wu-sheng:
Is any `CONSUMER` filter available? How monitoring works for `CONSUMER` side Dubbo3?

tedli:
> Is any `CONSUMER` filter available? How monitoring works for `CONSUMER` side Dubbo3?

Yes, there is.

https://github.com/apache/dubbo/blob/dubbo-3.0.4/dubbo-monitor/dubbo-monitor-api/src/main/java/org/apache/dubbo/monitor/support/MonitorClusterFilter.java

I notice that some other plugins has different instrumentation classes for server and client like the grpc plugin.

I'm trying to add an instrumentation class for dubbo 3 consumer in the same way.

wu-sheng:
Sure, you could disable the whole `MonitorFilter`-based instrumentation for Dubbo3, and pick a better one.

tedli:
Hi @wu-sheng ,

I found that if using the same scenario test code like dubbo 2.7.x, the consumer proxy object will use an InJvmInvoker, or if explicitly set inJvm to false, a DubboInvoker will be used, neither of them loads filters.

 Then I ran a [EmbeddedZooKeeper](https://github.com/apache/dubbo-samples/blob/master/dubbo-samples-echo/src/main/java/org/apache/dubbo/samples/echo/EmbeddedZooKeeper.java), by using a registry, 

https://github.com/apache/dubbo/blob/5a5e3f3fd7d0482b6124bd5b52b10f4900856438/dubbo-cluster/src/main/java/org/apache/dubbo/rpc/cluster/filter/ProtocolFilterWrapper.java#L68-L75

```java
@Override
public <T> Invoker<T> refer(Class<T> type, URL url) throws RpcException {
    if (UrlUtils.isRegistry(url)) {
        return protocol.refer(type, url);
    }
    FilterChainBuilder builder = getFilterChainBuilder(url);
    return builder.buildInvokerChain(protocol.refer(type, url), REFERENCE_FILTER_KEY, CommonConstants.CONSUMER);
}
```

the consumer proxy object loads filters. So skywalking interceptor on consumer side affects. The produced spans are as expect.

```yaml
segmentItems:
- serviceName: dubbo-3.x-scenario
  segmentSize: 3
  segments:
  - segmentId: 42afe68bf5ad435787439867ed63a554.64.16390270335870000
    spans:
    - operationName: HEAD:/dubbo-3.x-scenario/case/healthCheck
      operationId: 0
      parentSpanId: -1
      spanId: 0
      spanLayer: Http
      startTime: 1639027033589
      endTime: 1639027033684
      componentId: 1
      isError: false
      spanType: Entry
      peer: ''
      skipAnalysis: false
      tags:
      - {key: url, value: 'http://localhost:8080/dubbo-3.x-scenario/case/healthCheck'}
      - {key: http.method, value: HEAD}
  - segmentId: 42afe68bf5ad435787439867ed63a554.86.16390270340630000
    spans:
    - operationName: org.apache.skywalking.apm.testcase.dubbo3.services.GreetService.doBusiness(String)
      operationId: 0
      parentSpanId: -1
      spanId: 0
      spanLayer: RPCFramework
      startTime: 1639027034063
      endTime: 1639027034068
      componentId: 3
      isError: false
      spanType: Entry
      peer: 172.17.0.2:40328
      skipAnalysis: false
      tags:
      - {key: url, value: 'dubbo://172.17.0.2:20080/org.apache.skywalking.apm.testcase.dubbo3.services.GreetService.doBusiness(String)'}
      - {key: arguments, value: helloWorld}
      refs:
      - {parentEndpoint: 'GET:/dubbo-3.x-scenario/case/dubbo', networkAddress: '172.17.0.2:0',
        refType: CrossProcess, parentSpanId: 1, parentTraceSegmentId: 42afe68bf5ad435787439867ed63a554.65.16390270336920000,
        parentServiceInstance: ae9af82e06fe419097b7483b8e2d6522@172.17.0.2, parentService: dubbo-3.x-scenario,
        traceId: 42afe68bf5ad435787439867ed63a554.65.16390270336920001}
  - segmentId: 42afe68bf5ad435787439867ed63a554.65.16390270336920000
    spans:
    - operationName: org.apache.skywalking.apm.testcase.dubbo3.services.GreetService.doBusiness(String)
      operationId: 0
      parentSpanId: 0
      spanId: 1
      spanLayer: RPCFramework
      startTime: 1639027033946
      endTime: 1639027034073
      componentId: 3
      isError: false
      spanType: Exit
      peer: 172.17.0.2:0
      skipAnalysis: false
      tags:
      - {key: url, value: 'dubbo://172.17.0.2:0/org.apache.skywalking.apm.testcase.dubbo3.services.GreetService.doBusiness(String)'}
      - {key: arguments, value: helloWorld}
    - operationName: GET:/dubbo-3.x-scenario/case/dubbo
      operationId: 0
      parentSpanId: -1
      spanId: 0
      spanLayer: Http
      startTime: 1639027033692
      endTime: 1639027034075
      componentId: 1
      isError: false
      spanType: Entry
      peer: ''
      skipAnalysis: false
      tags:
      - {key: url, value: 'http://localhost:8080/dubbo-3.x-scenario/case/dubbo'}
      - {key: http.method, value: GET}
```

However, because a registry is used, the `address` no longer be the `localhost` and depends. So the `expectedData.yaml` can't keep using `localhost`.

And I didn't find something like `startWith`, `contains`, `endWith` in the skywalking-agent-test-tool, can I simply use `not null`?

wu-sheng:
> And I didn't find something like startWith, contains, endWith in the skywalking-agent-test-tool, can I simply use not null?

Yes, that is fine.

wu-sheng:
You need to add your case into GHA control file, then I could activate CI for you.

wu-sheng:
`Supported-list.md`, `Plugin-list.md` and `changes.md` should be updated, otherwise, the CI would fail.

tedli:
> `Supported-list.md`, `Plugin-list.md` and `changes.md` should be updated, otherwise, the CI would fail.

I saw that the checking for `Plugin-list.md` reads `skywalking-plugin.def`, I don't know how this file work. If I add new entry for dubbo3, should I adjust content of `skywalking-plugin.def` to something like this?

```
dubbo3.x=org.apache.skywalking.apm.plugin.asf.dubbo3.DubboProviderInstrumentation
dubbo3.x=org.apache.skywalking.apm.plugin.asf.dubbo3.DubboConsumerInstrumentation
```

By the way, I saw `retransform-class-scenario` failed in previous CI, did I broke it?

wu-sheng:
The name of skywalking-plugin.def provides the bootstrap configurable capabilities to disable this plugin directly though config file or system env, rather than removing jar files. So  it is better to rename preview to dubbo2.x and add dubbo3.x

wu-sheng:
retransform case should only fail unrelatedly. I guess. I can rerun if it happens again. Don't need to worry for now.

test/plugin/scenarios/dubbo-3.x-scenario/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+    <pluginRepositories>
+        <pluginRepository>
+            <id>spring-snapshots</id>
+            <url>https://repo.spring.io/snapshot</url>
+        </pluginRepository>
+        <pluginRepository>
+            <id>spring-milestones</id>
+            <url>https://repo.spring.io/milestone</url>
+        </pluginRepository>
+    </pluginRepositories>
```
wu-sheng, line 124: Do we really need this? It is better to use official Maven central, as we don't test snapshot versions.

apm-sniffer/apm-sdk-plugin/dubbo-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/asf/dubbo3/DubboInterceptor.java
```diff
@@ -0,0 +1,211 @@
...
+
+        RpcServiceContext serviceContext = RpcContext.getServiceContext();
+        boolean isConsumer;
+        URL url = serviceContext.getUrl();
+        if (url == null) {
+            url = serviceContext.getConsumerUrl();
+            isConsumer = url != null;
+        } else {
+            isConsumer = serviceContext.isConsumerSide();
+        }
```
wu-sheng, line 74: Should not `serviceContext.isConsumerSide()` have high priority? I can see from the existing plugin, `url == null` is not an issue, it is just for server side operation name.

wu-sheng: Is there anything different from Dubbo 2.7? I can see this in the existing codes.

```java
boolean isConsumer = rpcContext.isConsumerSide();
```

tedli: https://github.com/apache/dubbo/blob/5a5e3f3fd7d0482b6124bd5b52b10f4900856438/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcServiceContext.java#L142-L145

```java
@Override
public boolean isConsumerSide() {
    return getUrl().getSide(PROVIDER_SIDE).equals(CONSUMER_SIDE);
}

@Override
public boolean isProviderSide() {
    return !isConsumerSide();
}
```

`isConsumerSide` depends on `getUrl`, when I debugging locally, the url sometime is `null`, caused NullPointerException. So I think apm should not crash rpc, and add this checking.

wu-sheng: Interesting, I think you are better to report a bug to Dubbo. I am OK for your plugin codes, but Dubbo users may complain, and `sometimes null` seems a bug :)

wu-sheng: Then please fix the format(unexpected code changes), I am good.

wu-sheng (approved):

Cuihongsen:
When will a new version support this new feature be released

wu-sheng:
In this month, maybe. We are waiting for a grpc next release.

PR #76, opened 2021-12-01
feat: Support version 2.8 and 2.9 of pulsar client

wallezhang:
### Fix <apache/skywalking#8175>

The pulsar client has changed the generator of the protocol code since version 2.8.0, I create a new plugin `pulsar-2.8.x-plugin` to use a new method adding the carrier item to message sending to the pulsar broker. And also abstract the common logic between `pulsar-plugin` and `pulsar-2.8.x-plugin` to a new module called `pulsar-common`.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes apache/skywalking#8175.


wallezhang:
TBR

wu-sheng:
@codelipenghui  Could you review here?

docs/en/setup/service-agent/java-agent/Plugin-list.md
```diff
@@ -127,3 +127,4 @@
 - clickhouse-0.3.x
 - kylin-jdbc-2.6.x-3.x-4.x
 - okhttp-2.x
+- pulsar-2.8.x
```
wu-sheng, line 130: Supported-list should be updated.

wu-sheng: `pulsar-plugin` should be renamed to `pulsar-2.2-2.7-plugin`.

CHANGES.md
```diff
@@ -10,6 +10,7 @@ Release Notes.
 * Build CLI from Docker instead of source codes, add alpine based Docker image.
 * Support set instance properties in json format.
 * Upgrade grpc-java to 1.42.1 and protoc to 3.17.3 to allow using native Mac osx-aarch_64 artifacts.
+* Support version 2.8.0+ of pulsar client.
```
```suggestion
* Support 2.8 and 2.9 of pulsar client.
```

I don't know whether this plugin works for versions in the future.

tjiuming:
I reviewed this pr, it seems that add a different impl for `MessagePropertiesInjector` mainly, others changed a little, need smoke test to ensure.

wu-sheng:
> I reviewed this pr, it seems that add a different impl for `MessagePropertiesInjector` mainly, others changed a little, need smoke test to ensure.

There is no smoke test, SkyWalking is always doing fully e2e testing automatically. For Pulsar, you could find tests with following versions. https://github.com/wallezhang/skywalking-java/blob/pulsar-plugin-suppport-2.8.x/test/plugin/scenarios/pulsar-scenario/support-version.list @tjiuming

wu-sheng (approved):

wu-sheng:
https://github.com/apache/skywalking-java/runs/4410118738?check_suite_focus=true You could find the test logs(all versions) here.

tjiuming:
> https://github.com/apache/skywalking-java/runs/4410118738?check_suite_focus=true You could find the test logs(all versions) here.

got it

PR #77, opened 2021-12-01
Also apply to the JsonLogger.

geekymv:


wu-sheng:
I can't see this applies to the JsonLogger. Could you explain?

geekymv:
@wu-sheng hi, I read the Javadoc of [JsonLogger.java](https://github.com/apache/skywalking-java/blob/main/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/logging/core/JsonLogger.java), when logging.resolver=JSON, logging.pattern can also use %agent_name, so i think the %agent_name is also apply to the JsonLogger, Only apply to the PatternLogger is not correct.

wu-sheng:
`logging.pattern` is not `resolver`, `%agent_name` is a resolver only. Pattern means what is the layout of logs, but there is pointless to define sequence of a JSON, and it is not used in this way.

PR #81, opened 2021-12-09
Fix <TracePathMatcher should match pattern "**" with paths end by "/">

DominikHubacek:

### DIFF
--- a/apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
+++ b/apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
@@ -21,6 +21,12 @@ package org.apache.skywalking.apm.plugin.trace.ignore.matcher;
 public class FastPathMatcher implements TracePathMatcher {
     @Override
     public boolean match(String pattern, String path) {
+        if (pattern.indexOf(pattern.length() - 1) == '/') {
+            pattern = pattern.substring(0, pattern.length() - 2);
+        }
+        if (path.indexOf(path.length() - 1) == '/') {
+            path = path.substring(0, path.length() - 2);
+        }
         return normalMatch(pattern, 0, path, 0);
     }
 
@@ -97,11 +103,16 @@ public class FastPathMatcher implements TracePathMatcher {
                     return normalMatch(pat, p + 1, str, s + 1);
                 }
 
+                //if pattern ends, and str ends with '/'
+                if(pc == '\u0000' && safeCharAt(str, s+1) == '\u0000'){
+                    return true;
+                }
+
                 // Not matched string in current path part.
-                //             ↓        ↓
-                // pattern: a/*      a/*d
-                //              ↓        ↓
-                // string:  a/bc/    a/bc/
+                //             ↓
+                // pattern:  a/*d
+                //              ↓
+                // string:   a/bc/
                 return false;
             }
 
@@ -122,11 +133,6 @@ public class FastPathMatcher implements TracePathMatcher {
     }
 
     private boolean multiWildcardMatch(String pat, int p, String str, int s) {
-        // End of pattern, just check the end of string is '/' quickly.
-        if (p >= pat.length() && s < str.length()) {
-            return str.charAt(str.length() - 1) != '/';
-        }
-
         while (true) {
             // Try to enter normal mode, if not matched, increasing pointer of string and try again.
             if (!normalMatch(pat, p, str, s)) {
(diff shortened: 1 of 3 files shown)


wu-sheng:
@devkanro Please help on reviewing this new one.

wu-sheng:
@DominikHubacek You need to fix CI, look like code style issue.

wu-sheng:
@DominikHubacek It is better you could confirm all these edge cases. https://github.com/apache/skywalking-java/pull/72#issuecomment-975170969

DominikHubacek:
> @DominikHubacek You need to fix CI, look like code style issue.

Will do

DominikHubacek:
> @DominikHubacek It is better you could confirm all these edge cases. https://github.com/apache/skywalking-java/pull/72#issuecomment-975170969

I changed tests based on that comment, but only those which are checked(true), will add those unchecked (false) as well

devkanro:
ok

devkanro:
I have improved the state machine to support #72, check the comment to get more infomation.
@DominikHubacek @wu-sheng

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
```diff
@@ -84,6 +84,22 @@ private boolean normalMatch(String pat, int p, String str, int s) {
     private boolean wildcardMatch(String pat, int p, String str, int s) {
         char pc = safeCharAt(pat, p);
 
+        // End of pattern, check string only has zero or one '/' at end.
```
devkanro, line 87: If we get EOS(end of string) when we enter wildcard match state.
We just check the string only has zero or one '/' at end.

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
```diff
@@ -122,9 +138,11 @@ private boolean wildcardMatch(String pat, int p, String str, int s) {
...
 
     private boolean multiWildcardMatch(String pat, int p, String str, int s) {
-        // End of pattern, just check the end of string is '/' quickly.
-        if (p >= pat.length() && s < str.length()) {
-            return str.charAt(str.length() - 1) != '/';
+        switch (safeCharAt(pat, p)) {
+            // End of pattern, just return true quickly.
+            case 0: return true;
+            // Skip next '/' for pattern to match zero path part.
+            case '/': p++;
```
devkanro, line 145: Add this branch to support `eureka/**/test` matches `eureka/test` (** means zero path part)

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
```diff
@@ -122,9 +138,11 @@ private boolean wildcardMatch(String pat, int p, String str, int s) {
     }
 
     private boolean multiWildcardMatch(String pat, int p, String str, int s) {
-        // End of pattern, just check the end of string is '/' quickly.
-        if (p >= pat.length() && s < str.length()) {
-            return str.charAt(str.length() - 1) != '/';
+        switch (safeCharAt(pat, p)) {
+            // End of pattern, just return true quickly.
+            case 0: return true;
```
devkanro, line 143: If we get EOS when we enter multi-wildcard match state, return true quickly, because `**` mean anything

wu-sheng:
OK, so, this becomes a hybrid PR. @DominikHubacek Please recheck this PR, and if you don't feel comfortable about this PR will(if merged) be on your name, we could consider other options.

apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
DominikHubacek: > OK, so, this becomes a hybrid PR. @DominikHubacek Please recheck this PR, and if you don't feel comfortable about this PR will(if merged) be on your name, we could consider other options.

its ok, I checked if my changes still work as intended and fixed 1 test based on devkanro comment

DominikHubacek:
> OK, so, this becomes a hybrid PR. @DominikHubacek Please recheck this PR, and if you don't feel comfortable about this PR will(if merged) be on your name, we could consider other options.

its ok, I checked if my changes still work as intended and fixed 1 test based on devkanro comment

wu-sheng:
Could you two confirm this is good now?

wu-sheng (approved):

PR #88, opened 2022-01-12
Compatible with the versions after dubbo-2.7.14 

wuwen5:
[The ```rpcContext.getAttachments()``` method returns is  copy Map](https://github.com/apache/skywalking/issues/8415)

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/8415.

### DIFF
--- a/test/plugin/scenarios/dubbo-2.7.x-scenario/support-version.list
+++ b/test/plugin/scenarios/dubbo-2.7.x-scenario/support-version.list
@@ -15,3 +15,4 @@
 # limitations under the License.
 
 2.7.3
+2.7.14
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ Release Notes.
 * Fix TracePathMatcher should match pattern "**" with paths end by "/"
 * Add support `returnedObj` expression for apm-customize-enhance-plugin
 * Fix the bug that httpasyncclient-4.x-plugin  puts the dirty tracing context in the connection context
+* Compatible with the versions after dubbo-2.7.14
 
 #### Documentation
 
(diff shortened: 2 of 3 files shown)


apm-sniffer/apm-sdk-plugin/dubbo-2.7.x-plugin/pom.xml
```diff
@@ -32,7 +32,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <dubbo.version>2.7.0</dubbo.version>
+        <dubbo.version>2.7.15</dubbo.version>
```
wu-sheng, line 35: I don't think this is the correct change. Java agent plugin covers different versions. You change the codes according to the latest version, which makes everyone using < 2.7.14 + SkyWalking agent 8.9.0+ faces unexpected bugs.

wu-sheng: This kind of API behavior change is strange in this tiny bug-fix version change. Could you report to Dubbo community for an explanation?

wu-sheng: If this API exists for all previous 2.7.x versions. please don't upgrade this dependency version.

wuwen5: Agree

wu-sheng:

wu-sheng:
https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/dubbo-2.7.x-scenario/support-version.list#L17

The new in-compatible version should be added here to verify this fix.

test/plugin/scenarios/dubbo-2.7.x-scenario/support-version.list
```diff
@@ -14,4 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-2.7.3
\ No newline at end of file
+2.7.3
+2.7.14
+2.7.15
```
wu-sheng, line 19: Do 2.7.14 and 2.7.15 have any difference to test?

wu-sheng (approved):
If all versions of plugin tests pass, we should be good.

test/plugin/scenarios/dubbo-2.7.x-scenario/support-version.list
wuwen5: No, just because it's the latest version, I'll removed it.

PR #89, opened 2022-01-12
Support collecting metrics of ZGC

xu1009:
### Add an agent plugin to support <zgc>
- [ ] Add a test case for the new plugin, refer to [the doc](https://github.com/apache/skywalking-java/blob/main/docs/en/setup/service-agent/java-agent/Plugin-test.md)
- [ ] Add a component id in [the component-libraries.yml](https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/component-libraries.yml)

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<8411>.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,7 +20,7 @@ Release Notes.
 * Fix the bug that httpasyncclient-4.x-plugin  puts the dirty tracing context in the connection context
 * Compatible with the versions after dubbo-2.7.14
 * Follow protocol grammar fix `GCPhrase -> GCPhase`.
-* Support zgc gc time and count metric collect
+* Support ZGC GC time and count metric collect. (Require 9.0.0 OAP)
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ Release Notes.
 * Fix TracePathMatcher should match pattern "**" with paths end by "/"
 * Add support `returnedObj` expression for apm-customize-enhance-plugin
 * Fix the bug that httpasyncclient-4.x-plugin  puts the dirty tracing context in the connection context
+* support zgc gc time and count metric collect
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@ Release Notes.
 * Fix TracePathMatcher should match pattern "**" with paths end by "/"
 * Add support `returnedObj` expression for apm-customize-enhance-plugin
 * Fix the bug that httpasyncclient-4.x-plugin  puts the dirty tracing context in the connection context
-* support zgc gc time and count metric collect
+* Support zgc gc time and count metric collect
 
 #### Documentation
 
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/CMSGCModule.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/CMSGCModule.java
@@ -37,7 +37,7 @@ public class CMSGCModule extends GCModule {
     }
 
     @Override
-    protected String getNormalGcName() {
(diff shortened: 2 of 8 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/CMSGCModule.java
```diff
@@ -35,4 +35,9 @@ protected String getOldGCName() {
     protected String getNewGCName() {
         return "ParNew";
     }
+
+    @Override
+    protected String getNormalGcName() {
```
wu-sheng, line 40: Gc should be GC, it is a short term.

xu1009:
wecom-temp-45c9bc374a0b653652ebb6999dbfd5df

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/CMSGCModule.java
xu1009: ok

xu1009:
ok

xu1009:
it seems, the test not update submodule to compile and genrate code, so it fails

wu-sheng:
Are you changing protocol back to old version by mistake?

xu1009:
> Are you changing protocol back to old version by mistake?

no, i change GCPhrase to GCPhase follow the proto

xu1009:
i get it

wu-sheng:
Could you try this Java agent(with ZGC support) in ZGC JVM, but send data to OAP without ZGC support? I want to check whether the new added `normal` phase would work and what is the result?

xu1009:
ok, i try it In the afternoon,  and feedback

wu-sheng:
Thanks, you could add a breakpoint on JVM data deserialization, and check what value it will be in older proto data.

xu1009:
wecom-temp-0757b9a96491765557364a18219cf844

wecom-temp-fe2d7e38e89136c139313c9a9585224d

image


the enum is unknown value when deserialization, the value is still 2

last the phase not record, and ui not display

CHANGES.md
```diff
@@ -20,6 +20,7 @@ Release Notes.
 * Fix the bug that httpasyncclient-4.x-plugin  puts the dirty tracing context in the connection context
 * Compatible with the versions after dubbo-2.7.14
 * Follow protocol grammar fix `GCPhrase -> GCPhase`.
+* Support zgc gc time and count metric collect
```
wu-sheng, line 23:
```suggestion
* Support ZGC GC time and count metric collect. (Require 9.0.0 OAP)
```

I think it is better we add this. WDYT?

xu1009: ok

wu-sheng (approved):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/GCProvider.java
```diff
@@ -62,6 +62,9 @@ private GCMetricAccessor findByBeanName(String name) {
         } else if (name.equals("MarkSweepCompact")) {
             // Serial collector ( -XX:+UseSerialGC )
             return new SerialGCModule(beans);
+        } else if (name.indexOf("ZGC") > -1) {
+            // Serial collector ( -XX:+UseZGC )
```
kezhenxu94, line 66: @xu1009 can you open a PR to correct this comment?

```suggestion
            // Z garbage collector ( -XX:+UseZGC )
```

xu1009: ok

PR #92, opened 2022-01-17
Verify ShardingSphere 5.0.0 GA

wu-sheng:


wu-sheng:
@TeslaCN From this check, it seems ShardingSphere APIs break again. The plugin doesn't work anymore.

See this, https://github.com/apache/skywalking-java/runs/4835317197?check_suite_focus=true

PR #95, opened 2022-01-20
Support thread pool metric collect

xu1009:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/8433.


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/theadpool/ThreadPoolRegister.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final List<ThreadPoolSourceData> THREAD_POOL_SOURCE_DATA_LIST = synchronizedList(new ArrayList<>());
+
+    public static void registerThreadPoolSource(ThreadPoolName threadPoolName, ThreadPoolExecutor threadPoolExecutor) {
+        if (THREAD_POOL_SOURCE_DATA_LIST.stream().noneMatch(it -> it.getThreadPoolName().equals(threadPoolName))) {
+            ThreadPoolSourceData threadPoolSourceData = new ThreadPoolSourceData();
+            threadPoolSourceData.setThreadPoolName(threadPoolName);
+            threadPoolSourceData.setHasBuildMetric(false);
+            threadPoolSourceData.setThreadPoolExecutor(threadPoolExecutor);
+            THREAD_POOL_SOURCE_DATA_LIST.add(threadPoolSourceData);
+        }
```
wu-sheng, line 38: Without `lock` block(ReentrantLock, this `synchronizedList` would not help. You still could face thread pools in the same duplicate name.

apm-sniffer/optional-plugins/undertow-worker-thread-pool-plugin/src/main/java/org/apache/skywalking/apm/plugin/undertow/worker/thread/pool/UndertowWorkerThreadPoolConstructorIntercept.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+import org.apache.skywalking.apm.agent.core.theadpool.ThreadPoolName;
+import org.apache.skywalking.apm.agent.core.theadpool.ThreadPoolRegister;
+
+public class UndertowWorkerThreadPoolConstructorIntercept implements InstanceConstructorInterceptor {
+
+    @Override
+    public void onConstruct(EnhancedInstance objInst, Object[] allArguments) throws Throwable {
+        ThreadPoolExecutor threadPoolExecutor = (ThreadPoolExecutor) objInst;
+        ThreadPoolRegister.registerThreadPoolSource(ThreadPoolName.UNDERTOW_WORKER_POOL, threadPoolExecutor);
```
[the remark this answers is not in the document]
wu-sheng: `EnhancedInstance` object has a private, and new `dynamicField` to host anything you need. I prefer you don't define a static object to host the created metric..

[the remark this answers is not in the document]
wu-sheng: Such as the Guage you created.

wu-sheng:
`MeterService` is already hosting the created metric. Please read how `CounterConstructInterceptor` works. And you should reduce unnecessary codes.

wu-sheng:
BTW, I am asking @mrproliu to enhance the [plugin test tool](https://github.com/apache/skywalking-agent-test-tool), which will be able to verify metrics, like it did for tracing data.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/theadpool/ThreadPoolService.java
```diff
@@ -0,0 +1,102 @@
...
+        String poolNameTag = "pool_name";
+        String metricTypeTag = "metric_type";
+        List<ThreadPoolSourceData> threadPoolSourceDataList = ThreadPoolRegister.getThreadPoolSourceDataList();
+        threadPoolSourceDataList.forEach(it -> {
+            if (!it.getHasBuildMetric()) {
+                String poolName = it.getThreadPoolName().name().toLowerCase();
+                ThreadPoolExecutor threadPoolExecutor = it.getThreadPoolExecutor();
+                METRIC_MAP.forEach(
+                        (key, value) -> MeterFactory.gauge(threadPoolMeterName, value.apply(threadPoolExecutor))
+                                .tag(poolNameTag, poolName).tag(metricTypeTag, key).build());
```
mrproliu, line 92: You don't need to create a scheduler to create the gauge, `MeterFactory#gauge` or other `MeterFactory#metric` means you could just register once, and [MeterSender](https://github.com/apache/skywalking-java/blob/main/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/meter/MeterSender.java) would collect the metric with scheduler.

mrproliu:
> BTW, I am asking @mrproliu to enhance the [plugin test tool](https://github.com/apache/skywalking-agent-test-tool), which will be able to verify metrics, like it did for tracing data.

I have already enhanced it, It could receive the meter data from meter report protocol. If you want to use it, here is the [documentation](https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/plugin-test/#the-example-process-of-writing-meter-expected-data) and also have [plugin test been used](https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/apm-toolkit-trace-scenario/config/expectedData.yaml#L482-L504).

xu1009:
the way is to make new thread type add is simple,  just get ThreadPoolExecutor instacne and register it, and then, the main metric can be send to the backend

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/theadpool/ThreadPoolService.java
xu1009: bacasue plugin resgister and create gauge order is not certain， so scheduler get it and create the gauge, but only create onece

wu-sheng:
> the way is to make new thread type add is simple, just get ThreadPoolExecutor instacne and register it, and then, the main metric can be send to the backend

You just need to create the metric, register it and hold it in the pool object. What we recommended is making your PR easier.

wu-sheng:
@xu1009 Please take a deep like at our comments. Most things you did have been provided at the core level. Because most of this registration mechanism is a common requirement.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/theadpool/ThreadPoolService.java
wu-sheng: We provide you with a provider, rather than a static value. When the core is calling you, you could return the latest one. The registration is only about the name of the metric.

xu1009:
> > the way is to make new thread type add is simple, just get ThreadPoolExecutor instacne and register it, and then, the main metric can be send to the backend
> 
> You just need to create the metric, register it and hold it in the pool object. What we recommended is making your PR easier.

but, this way, every new plugin needs to create metric, i think eh

> @xu1009 Please take a deep like at our comments. Most things you did have been provided at the core level. Because most of this registration mechanism is a common requirement.

in this way, every new plugin needs to create metirc, i want to extract these common things to core

wu-sheng:
> in this way, every new plugin needs to create metirc, i want to extract these common things to core

What is the issue of creating a metric?
You don't need to consider that. How to create a metric is a core level thing. The plugin should keep all logic simple.
If you want to discuss what core APIs lack, we could work on that separately.

wu-sheng:
Please fix the CI and add meter plugin test according to the link above. https://github.com/apache/skywalking-java/pull/95#issuecomment-1018114627

xu1009:
ok, i add it lately

xu1009:
the test seems not work in new mac book with m1 cpu

wu-sheng:
> the test seems not work in new mac book with m1 cpu

Yes, we may not have enough images for m1 testing. As most of the team are still using the intel chip.

Could you find a Linux to run this? Or try to locate what is the block(s) in AARM?

wu-sheng:
Please take a little look at https://github.com/apache/skywalking-java/pull/73, these 2 files should be updated, one for CI, the other for people reading.

image

xu1009:
> > the test seems not work in new mac book with m1 cpu
> 
> Yes, we may not have enough images for m1 testing. As most of the team are still using the intel chip.
> 
> Could you find a Linux to run this? Or try to locate what is the block(s) in AARM?

image

not have image,  i will try it in other tomorrow, our company has used M1 widely ==

wu-sheng:
> not have image, i will try it in other tomorrow, our company has used M1 widely ==

I see, take a look at https://hub.docker.com/_/eclipse-temurin?tab=tags. AdoptOpenJDK has moved to Eclipse Temurin. I can see AARM images are available there. 

FYI @dmsolr

wu-sheng:
If you will submit any pull request about enhancing M1 testing, please submit through another pull request with detailed steps about running which case locally.
I think I am the only few ppl in the project committer team using the M1 series(M1 Max) chip. I have to verify this locally.

wu-sheng:
Are you going to update this PR? No update for 3 days.

xu1009:
> Are you going to update this PR? No update for 3 days.

sorry, very busy with my own job these days, i am lookig this

apm-sniffer/optional-plugins/undertow-worker-thread-pool-plugin/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>optional-plugins</artifactId>
```
wu-sheng, line 24: Questions, 
1. Should we merge this plugin into the existing undertow tracing plugin? 
2. Should we make it works in default? I can see this is an optional plugin, but a thread pool plugin should not impact the performance, right?

wu-sheng: About (1), if you want this plugin to be disabled separately from the tracing plugin, I am fine to keep the codes in another plugin jar. But (2) should be considered.

mrproliu (approved):
LGTM

apm-sniffer/optional-plugins/undertow-worker-thread-pool-plugin/pom.xml
wu-sheng: @xu1009 Please follow this, then we could merge on both sides.

xu1009: thread pool plugin may impact performance, because it has lock when java get the metric, so it is optional plugin, peoeple really need can add it

wu-sheng: The fetching period is 20s, I think it's find to keep it in default.
This lock happens in every getting and returning threads from pool. It is not a big impact.
Agree?

xu1009: yes, but the fetching period can be modified by user, if it is set 1s, and the service has high qps, it may impact

wu-sheng: Manual setting equals optional plugin. So, don't worry.

xu1009: so we make the plugin in default, an add config control it ?

wu-sheng: Moving it to default is enough. Don't need to add a new config.

xu1009: ok

wu-sheng (approved):

CHANGES.md
```diff
@@ -4,7 +4,7 @@ Release Notes.
 
 8.10.0
 ------------------
-
+* Support Java thread pool metric collect.
```
kezhenxu94, line 7: This PR only provides thread pool metrics of Undertow, but the change log makes me (and probably other users) feel that we support JDK's thread pool. WDYT @wu-sheng ?

wu-sheng: Updated, check the latest codes

leeeevincent0123: I noted that the V9.0.0 just display tomcat thread pool metric. Now I have many customized thread pool which created by "org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor". How can I collect the metric of my thread pool and display it in UI？Is need to write another customized  plugins or just need write some config?

wu-sheng: Check #146 you may be interested.

leeeevincent0123: I had checked #146 before i asked and i had read the code. The code does not  contains any METRIC about "CorePoolSize","MaximumPoolSize".
image

I find a plugin more closer I wanted but hard coded[org.apache.skywalking.apm.plugin.undertow.worker.thread.pool.UndertowWorkerThreadPoolConstructorIntercept]. So I want to  konw if i have to write some customized plugin which the count is same to the ThreadPool count？Or have some config more convenient？

image

wu-sheng: Why only including two methods, there was a long discussion on that PR, you could check.

You definitely could write a new plugin to do whatever you want, but you should know, that the discussion context is recommended for you before you take action. There are risks to instrument all, which is why we don't do that.

leeeevincent0123: Thanks, i will check later

songhao1105: @wu-sheng There are many thread pools in my project, such as ThreadPoolA and ThreadPoolB. If I want to monitor the metrics of these two thread pools separately, do I need to add a new plugin to the agent and two rules to the threadpool.ymal of OAP like:
\- name: ThreadPoolA
  exp: ThreadPoolA.avg(['metric_type', 'pool_name', 'instance', 'service'])
\- name: ThreadPoolB
  exp: ThreadPoolB.avg(['metric_type', 'pool_name', 'instance', 'service'])

songhao1105: > Thanks, i will check later

wu-sheng: Learn MAL, your questions are irrelevant

PR #96, opened 2022-01-24
[Feature] Support collecting database connection pool metric,support dbcp2/druid/hikari

sunOnly:


wu-sheng:
Please follow the discussion starting from https://github.com/apache/skywalking-java/pull/95#issuecomment-1019757059.

Don't build the register mechanism. This is not a good practice. 

Also, follow this, https://github.com/apache/skywalking-java/pull/95#issuecomment-1019815455, meter plugin test is required like the trace plugin test.

wu-sheng:
Are you going to update this PR? No update for 3 days.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
```diff
@@ -72,4 +90,9 @@ public boolean isOverrideArgs() {
     public StaticMethodsInterceptPoint[] getStaticMethodsInterceptPoints() {
         return new StaticMethodsInterceptPoint[0];
     }
+
+    @Override
+    public boolean isBootstrapInstrumentation() {
+        return true;
```
wu-sheng, line 96: Why need bootstrap instrumentation? Is this a mischange?

apm-sniffer/apm-sdk-plugin/druid-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/druid/v1/define/DruidDataSourceInstrumentation.java
```diff
@@ -94,4 +109,10 @@ public boolean isOverrideArgs() {
     public StaticMethodsInterceptPoint[] getStaticMethodsInterceptPoints() {
         return new StaticMethodsInterceptPoint[0];
     }
+
+    @Override
+    public boolean isBootstrapInstrumentation() {
+        return true;
```
wu-sheng, line 115: Same as above

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/define/HikariPoolInstrumentation.java
```diff
@@ -0,0 +1,75 @@
...
+        return new InstanceMethodsInterceptPoint[0];
+    }
+
+    @Override
+    public StaticMethodsInterceptPoint[] getStaticMethodsInterceptPoints() {
+        return new StaticMethodsInterceptPoint[0];
+    }
+    @Override
+    public boolean isBootstrapInstrumentation() {
+        return true;
```
wu-sheng, line 73: Same as above

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/define/HikariPoolInstrumentation.java
```diff
@@ -0,0 +1,75 @@
...
+    @Override
+    public InstanceMethodsInterceptPoint[] getInstanceMethodsInterceptPoints() {
+        return new InstanceMethodsInterceptPoint[0];
+    }
+
+    @Override
+    public StaticMethodsInterceptPoint[] getStaticMethodsInterceptPoints() {
+        return new StaticMethodsInterceptPoint[0];
+    }
+    @Override
```
wu-sheng, line 71: You need to import code style to make sure your codes are following the requirements.

wu-sheng:
Please fix CI and follow the comments. Then, you need to add plugin tests.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
sunOnly: Without rewriting, enhancement does not work

apm-sniffer/apm-sdk-plugin/druid-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/druid/v1/define/DruidDataSourceInstrumentation.java
sunOnly: Without rewriting, enhancement does not work

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/define/HikariPoolInstrumentation.java
sunOnly: No, no problem

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
wu-sheng: It doesn't make sense. Bootstrap instrumentation is for rt.jar, such as javax.*

wu-sheng: Check all existing plugins, there are very few activating this setting. And for DBCP, druid and hikaricp existing plugins, all of them don't require this. Please recheck.

wu-sheng: I can see you delete one, but keep the other two?

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/define/HikariPoolInstrumentation.java
sunOnly: > You need to import code style to make sure your codes are following the requirements.

Yes, multiple verification deletions cannot be enhanced

wu-sheng: What is this comment about?

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
sunOnly: > I can see you delete one, but keep the other two?

Yes, multiple verification deletions cannot be enhanced

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/define/HikariPoolInstrumentation.java
sunOnly: > What is this comment about?

Sorry, isBootstrapInstrumentation Override commented

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
wu-sheng: I am afraid you have to check. This is not right in theory. 
Let's take DBCP plugin as an example, the interceptor and instrumentation existed for months, and are verified by plugin e2e tests in every commit. Why does your instrumentation not work? 
I have to ask you to example, from my agent core experience, this is not the right way to write a plugin.

sunOnly: > I am afraid you have to check. This is not right in theory. Let's take DBCP plugin as an example, the interceptor and instrumentation existed for months, and are verified by plugin e2e tests in every commit. Why does your instrumentation not work? I have to ask you to example, from my agent core experience, this is not the right way to write a plugin.

I debug, screenshot

wu-sheng: Again, compare to exiting codes and plugins. And begin to write the official plugin e2e tests. The existing codes there have verified, this option is not required to return `true`. I have to insist on this way.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
```diff
@@ -60,6 +63,22 @@ public String getMethodsInterceptor() {
...
 
+                    @Override
+                    public boolean isOverrideArgs() {
+                        return false;
+                    }
+                },
+                new InstanceMethodsInterceptPoint() {
+                    @Override
+                    public ElementMatcher<MethodDescription> getMethodsMatcher() {
+                        return named("setUrl").and(takesArgument(0, String.class));
```
lujiajing1126, line 74: I suggest not adding the hook for metrics at this point.

1. the datasource may not be fully initialized then.
2. Versions `1.0.x` and `>=1.1.0` may behavior quite differently.

The `Alibaba/Druid` authors use the following method  (static method `DruidDataSourceStatManager.addDataSource`) to register `MBean` which means at this stage, the metrics can be exposed.

https://github.com/alibaba/druid/blob/f937ecd43363b2df61afdc120858097cb2510580/src/main/java/com/alibaba/druid/pool/DruidDataSource.java#L2151

Besides, we can make sure that each DataSource is only registered once which is guarded by the `mbeanRegistered` property.

lujiajing1126 (changes requested):

apm-sniffer/apm-sdk-plugin/druid-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/druid/v1/define/DruidDataSourceInstrumentation.java
```diff
@@ -82,6 +84,22 @@ public String getMethodsInterceptor() {
...
 
+                    @Override
+                    public boolean isOverrideArgs() {
+                        return false;
+                    }
+                },
+                new InstanceMethodsInterceptPoint() {
+                    @Override
+                    public ElementMatcher<MethodDescription> getMethodsMatcher() {
+                        return named("setUrl").and(takesArgument(0, String.class));
```
lujiajing1126, line 95: I suggest not adding the hook for metrics at this point.

the datasource may not be fully initialized then.
Versions 1.0.x and >=1.1.0 may behavior quite differently.
The Alibaba/Druid authors use the following method (static method DruidDataSourceStatManager.addDataSource) to register MBean which means at this stage, the metrics can be exposed.

https://github.com/alibaba/druid/blob/f937ecd43363b2df61afdc120858097cb2510580/src/main/java/com/alibaba/druid/pool/DruidDataSource.java#L2151

Besides, we can make sure that each DataSource is only registered once which is guarded by the mbeanRegistered property.

lujiajing1126 (changes requested):

lujiajing1126:
Additionally for HikariCP, they have some extra metrics at runtime,

https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/metrics/IMetricsTracker.java

Are you going to consider these?

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/define/HikariPoolInstrumentation.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public ConstructorInterceptPoint[] getConstructorsInterceptPoints() {
+        return new ConstructorInterceptPoint[]{
+                new ConstructorInterceptPoint() {
+
+                    @Override
+                    public ElementMatcher<MethodDescription> getConstructorMatcher() {
+                        return any();
```
lujiajing1126, line 51: For `HikariCP >= 3.0`, `HikariDataSource.seal()` method may be a better interceptor point to ensure there is a sealed `HikariDataSource`.

wu-sheng:
The other undertow thread pool PR has been merged. You could take a reference.

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/HikariDataSourceConstructorInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+        METRIC_MAP.put("totalConnections", (HikariPoolMXBean hikariPoolMXBean) -> () -> (double) hikariPoolMXBean.getTotalConnections());
+        METRIC_MAP.put("idleConnections", (HikariPoolMXBean hikariPoolMXBean) -> () -> (double) (hikariPoolMXBean.getIdleConnections()));
+        METRIC_MAP.put("threadsAwaitingConnection", (HikariPoolMXBean hikariPoolMXBean) -> () -> (double) hikariPoolMXBean.getThreadsAwaitingConnection());
+    }
+
+    @Override
+    public void onConstruct(EnhancedInstance objInst, Object[] allArguments) throws Throwable {
+        if (LOGGER.isInfoEnable()) {
+            LOGGER.debug("metric hikari init");
+        }
```
kezhenxu94, line 54: It doesn't make sense ` if isInfoEnable()` but log in `debug` level

wu-sheng:
@sunOnly Are you going to continue on this?

sunOnly:
> @sunOnly Are you going to continue on this?
Yes

wu-sheng:
Thanks for the feedback.

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/define/HikariPoolInstrumentation.java
sunOnly: > For `HikariCP >= 3.0`, `HikariDataSource.seal()` method may be a better interceptor point to ensure there is a sealed `HikariDataSource`.

This has been implemented and submitted

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
wu-sheng: @sunOnly Why does DBCP plugin still keep this?

sunOnly: > @sunOnly Why does DBCP plugin still keep this?

The local test does not take effect. See if there is a better entry point

wu-sheng:
Please notice https://github.com/apache/skywalking-java/pull/96#discussion_r800378221 and add plugin tests for different versions and frameworks.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
wu-sheng: This class has been instrumented by the existing plugin, you are just adding another method. This should work. If not, I am guessing your test case is special? Did you debug for the existing interceptor?

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/PoolingSetUrlSourceInterceptor.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * {@link PoolingSetUrlSourceInterceptor} intercepted the method of DBCP set url and register metric monitor.
+ */
+public class PoolingSetUrlSourceInterceptor implements InstanceMethodsAroundInterceptor {
+    private static final String METER_NAME = "datasource";
+    private static final ILog LOGGER = LogManager.getLogger(PoolingSetUrlSourceInterceptor.class);
+    private static final Map<String, Function<BasicDataSource, Supplier<Double>>> METRIC_MAP = new HashMap<String, Function<BasicDataSource, Supplier<Double>>>();
```
wu-sheng, line 42: Hi @sunOnly I asked @ascrutae to help on locating what is going on about https://github.com/apache/skywalking-java/pull/96/files#r800393136

All of your plugins have issues with using the target class(in here `BasicDataSource`) in instrumentation define and also in interceptor's static field declaring. 
Then in this case, `BasicDataSource` is loaded by the agent class loader already, which is impossible to be enhanced by the plugin system. Also, we checked the plugins of druid and hikaricp plugins, they are not working too.

The right way to fix this using `javax.sql.DataSource` here for the declaring, and do typecast if you have to use it in the runtime.
Also, you should know, even you opened **BootstrapInstrumentation**, it just overrides the error, but your plugin is still not working. I think you could verify locally when you debug locally or write plugin tests.

Please change the ways we recommended, and add all required plugin tests.

sunOnly: > Hi @sunOnly I asked @ascrutae to help on locating what is going on about https://github.com/apache/skywalking-java/pull/96/files#r800393136
> 
> All of your plugins have issues with using the target class(in here `BasicDataSource`) in instrumentation define and also in interceptor's static field declaring. Then in this case, `BasicDataSource` is loaded by the agent class loader already, which is impossible to be enhanced by the plugin system. Also, we checked the plugins of druid and hikaricp plugins, they are not working too.
> 
> The right way to fix this using `javax.sql.DataSource` here for the declaring, and do typecast if you have to use it in the runtime. Also, you should know, even you opened **BootstrapInstrumentation**, it just overrides the error, but your plugin is still not working. I think you could verify locally when you debug locally or write plugin tests.
> 
> Please change the ways we recommended, and add all required plugin tests.

received

apm-sniffer/apm-sdk-plugin/druid-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/druid/v1/PoolingAddDruidDataSourceInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public void beforeMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes, MethodInterceptResult result) {
+
+    }
+
+    @Override
+    public Object afterMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes, Object ret) {
+        if (LOGGER.isInfoEnable()) {
+            LOGGER.info("metric druid init");
+        }
```
wu-sheng, line 65: Could you remove codes like this? This seems like local debug info.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/PoolingSetUrlInterceptor.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        if (LOGGER.isInfoEnable()) {
+            LOGGER.info("metric dbcp init");
+        }
```
wu-sheng, line 63: Same here, remove please.

apm-sniffer/apm-sdk-plugin/hikaricp-3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/hikaricp/PoolingSealInterceptor.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        if (LOGGER.isInfoEnable()) {
+            LOGGER.info("metric hikari init");
+        }
```
wu-sheng, line 70: Same here, remove please.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/PoolingSetUrlInterceptor.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * {@link PoolingSetUrlInterceptor} intercepted the method of DBCP set url and register metric monitor.
+ */
+public class PoolingSetUrlInterceptor implements InstanceMethodsAroundInterceptor {
+    private static final String METER_NAME = "datasource";
+    private static final ILog LOGGER = LogManager.getLogger(PoolingSetUrlInterceptor.class);
+    private static final Map<String, Function<BasicDataSourceMXBean, Supplier<Double>>> METRIC_MAP = new HashMap<String, Function<BasicDataSourceMXBean, Supplier<Double>>>();
```
wu-sheng, line 42: This is fine, considering `BasicDataSourceMXBean` is only an interface. But still, this class metadata will be loaded into the default agent class loader, which could cause unnecessary memory costs. It is better to move it to a constructor? As this interceptor would only be created when a new pool is created, which is a rare case in the runtime, we don't need to be concerned about the performance. 

`TomcatThreadExecutorInterceptor` using `ThreadPoolExecutor` is fine because the ThreadPoolExecutor is loaded by JDK's class loader, then the default agent class loader doesn't need to load again.

wu-sheng:
`TBD` tag is removed, once the above comments are resolved and all plugin tests passed, we should be good.

wu-sheng:
`Test / dbcp-2.x-scenario (pull_request) ` fail due to `IllegalDataFileException expectedData.yaml `. Please make sure you have run the tests passed locally. Our CI resources are limited.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
```diff
@@ -60,6 +62,22 @@ public String getMethodsInterceptor() {
...
 
+                    @Override
+                    public boolean isOverrideArgs() {
+                        return false;
+                    }
+                },
+                new InstanceMethodsInterceptPoint() {
+                    @Override
+                    public ElementMatcher<MethodDescription> getMethodsMatcher() {
+                        return named("setUrl");
```
lujiajing1126, line 73: Just a final comment:

As I've stated before, we have to be careful with the interceptor points for `Alibaba/Druid` and `HikariCP`. But I'am not quite familiar with this `dbcp2` DataSource.

So, could you please justify the choice of this interceptor point?

wu-sheng:
It seems one changed case still fails.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
wu-sheng: @sunOnly First of all, good to see the tests passed. Let's follow this one. `#setUrl` is not a good point to begin the monitoring, because in the runtime, a datasource could be `not-created` at this moment, even never(code bug).

I was looking through the codes, it is likely `createDataSource` is a better method to begin all metrics registration and reporting. Meanwhile, I understand you intercept `#setUrl` is to get the URL parameter, I noticed there is `getUrl` method in the class, but with a `synchronized`, which I have concerns about deadlock. So, it is better you keep this `setUrl` method intercepting, but keep the value of URL into the `EnhancedInstance#setSkyWalkingDynamicField`. Then in new `createDataSource` interceptor, you could use `#getSkyWalkingDynamicField` to read this.

This would be more robust.

wu-sheng:
Another question, I noticed in the commit log, there are 2 ID, one is @sunOnly (GitHub), the other is `xiaojianjun`. Are these belong to the same person?

sunOnly:
> Another question, I noticed in the commit log, there are 2 ID, one is @sunOnly (GitHub), the other is `xiaojianjun`. Are these belong to the same person?

It's the same person

wu-sheng:
Please follow the comment above, there is one more todo.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
sunOnly: > @sunOnly First of all, good to see the tests passed. Let's follow this one. `#setUrl` is not a good point to begin the monitoring, because in the runtime, a datasource could be `not-created` at this moment, even never(code bug).
> 
> I was looking through the codes, it is likely `createDataSource` is a better method to begin all metrics registration and reporting. Meanwhile, I understand you intercept `#setUrl` is to get the URL parameter, I noticed there is `getUrl` method in the class, but with a `synchronized`, which I have concerns about deadlock. So, it is better you keep this `setUrl` method intercepting, but keep the value of URL into the `EnhancedInstance#setSkyWalkingDynamicField`. Then in new `createDataSource` interceptor, you could use `#getSkyWalkingDynamicField` to read this.
> 
> This would be more robust.
get

test/plugin/pom.xml
```diff
@@ -131,6 +131,9 @@
                     <sourceDirectories>
                         <sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>
                         <sourceDirectory>${project.build.testSourceDirectory}</sourceDirectory>
+<sourceDirectory>scenarios/dbcp-2.x-scenario</sourceDirectory>
+<sourceDirectory>scenarios/druid-1.x-scenario</sourceDirectory>
```
mrproliu, line 135: Please remove these declare, this should be generated by CI.

wu-sheng:
@lujiajing1126 Please recheck.

wu-sheng (dismissed):
Once tests pass, I am good.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/PoolingSetUrlInterceptor.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        ConnectionInfo connectionInfo = URLParser.parser((String) allArguments[0]);
+        String tagValue = connectionInfo.getDatabaseName() + "_" + connectionInfo.getDatabasePeer();
+        if (tagValue != null) {
```
lujiajing1126, line 43: As I understood, it could not be null?

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
lujiajing1126: > I was looking through the codes, it is likely createDataSource is a better method to begin all metrics registration and reporting. Meanwhile, I understand you intercept #setUrl is to get the URL parameter, I noticed there is getUrl method in the class, but with a synchronized, which I have concerns about deadlock. So, it is better you keep this setUrl method intercepting, but keep the value of URL into the EnhancedInstance#setSkyWalkingDynamicField. Then in new createDataSource interceptor, you could use #getSkyWalkingDynamicField to read this.

What about using `jmxRegister()` called within `createDataSource()` (if dataSource is not created) as the interception point? 

Since every call to `getConnection()` will invoke `createDataSource()`, the performance may be impacted. It also exists from `2.0.1` to `2.9.0`.

lujiajing1126 (changes requested):

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
wu-sheng: > Since every call to getConnection() will invoke createDataSource()

Really? This doesn't seem reasonable.

wu-sheng: Just check the document, it seems being called repeatedly. We may need to find another way.

wu-sheng: > What about using jmxRegister() called within createDataSource() (if dataSource is not created) as the interception point?

@sunOnly Could you try this?

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/PoolingSetUrlInterceptor.java
sunOnly: > As I understood, it could not be null?

The external call may be set to null



> As I understood, it could not be null?
change to
    if (allArguments[0] != null) {
            ConnectionInfo connectionInfo = URLParser.parser((String) allArguments[0]);
            String tagValue = connectionInfo.getDatabaseName() + "_" + connectionInfo.getDatabasePeer();
            objInst.setSkyWalkingDynamicField(tagValue);
        }

wu-sheng: I think hrmeans _ is in the string combining. So, it can't be null.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/define/BasicDataSourceInstrumentation.java
lujiajing1126: > > Since every call to getConnection() will invoke createDataSource()
> 
> Really? This doesn't seem reasonable.

https://github.com/apache/commons-dbcp/blob/466091e51d875d2a4499adbbb63058e32ad5a7c7/src/main/java/org/apache/commons/dbcp2/BasicDataSource.java#L732

Seems so.

apm-sniffer/apm-sdk-plugin/dbcp-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/dbcp/v2/PoolingSetUrlInterceptor.java
lujiajing1126: > I think he means _ is in the string combining. So, it can't be null.

Exactly.

lujiajing1126:
I'am good with the last version. Let's wait for the CI.

lujiajing1126 (approved):
LGTM. Thanks for your patience!

PR #98, opened 2022-01-27
Support for ShardingSphere 5.0.0

dachuan9e:
Support for ShardingSphere 5.0.0


wu-sheng:
The changes.md in the root should be updated

test/plugin/scenarios/shardingsphere-5.0.0-scenario/support-version.list
```diff
@@ -14,4 +14,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-5.0.0-beta
+5.0.0
```
wu-sheng, line 17: Is `beta` still supported? Or it is out of the supported list. Either way, please make it clear in the change log.

wu-sheng:
It seems you breaks something, please recheck your codes.

test/plugin/scenarios/shardingsphere-5.0.0-scenario/support-version.list
dachuan9e: Beta can be supported too.

dachuan9e: However, we don't recommend the beta version.

wu-sheng: OK, just make sure it is clear in the change log. The version recommendation is your community's call for sure.

wu-sheng:
I fixed the existing cases. Please fix conflicts.

wu-sheng (approved):

wu-sheng:
Thanks for your contribution.

PR #107, opened 2022-02-09
Add plugin to support google guava EventBus

wallezhang:
### Add an agent plugin to support <Guava EventBus>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes apache/skywalking#8482.


apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -219,5 +219,5 @@
 
     public static final OfficialComponent APACHE_KYLIN_JDBC_DRIVER = new OfficialComponent(121, "apache-kylin-jdbc-driver");
 
-    public static final OfficialComponent GUAVA_EVENT_BUS = new OfficialComponent(122, "GuavaEventBus");
+    public static final OfficialComponent GUAVA_EVENT_BUS = new OfficialComponent(123, "GuavaEventBus");
```
wu-sheng, line 222: Please register this to the main repo first.

wallezhang: OK, I just submit a PR to main repo.

wu-sheng: https://github.com/apache/skywalking/pull/8524

wallezhang:
The graphql-9.x-scenario test failed due to connection problem. How to re-run the CI flow?

wu-sheng:
> The graphql-9.x-scenario test failed due to connection problem. How to re-run the CI flow?

In the `Details`, there is a `rerun` button on the top right. I am not sure whether it is only working for committer.

image

I reran it anyway.

wallezhang:
> > The graphql-9.x-scenario test failed due to connection problem. How to re-run the CI flow?
> 
> In the `Details`, there is a `rerun` button on the top right. I am not sure whether it is only working for committer.
> 
> image
> 
> I reran it anyway.

Thanks. It's only work for committer, I can't see the button.

wu-sheng:
Got it. Feel free to ping me in the next time.

wu-sheng:
Guava plugin breaks the existing plugin of ServiceComb

> Guava/EventBus/org.apache.servicecomb.serviceregistry.task.MicroserviceInstanceRegisterTask/onMicroserviceRegisterTask

This is found in the actual data.

wu-sheng:
I guess ServiceComb is using this framework somehow, and which causes new spans added and expected data doesn't match anymore.

wallezhang:
> I guess ServiceComb is using this framework somehow, and which causes new spans added and expected data doesn't match anymore.

I take a look. Is there any recommend way to fix it? Modify the expect data to match it as before?

test/plugin/scenarios/guava-eventbus-scenario/support-version.list
```diff
@@ -0,0 +1,29 @@
...
+22.0
+23.0
+24.0-jre
+25.0-jre
+26.0-jre
+27.0-jre
+28.0-jre
+29.0-jre
+30.0-jre
+31.0-jre
```
wu-sheng, line 29: Please pick some versions once all tests passed. This seems to cause too many resources.

wu-sheng:
> > I guess ServiceComb is using this framework somehow, and which causes new spans added and expected data doesn't match anymore.
> 
> I take a look. Is there any recommend way to fix it? Modify the expect data to match it as before?

Could you take a deep look first? I am not sure how much it is affecting the ServiceComb case. If there are only some new local spans added, then I am good to just update the expect data.

wu-sheng:
I noticed the servicecomb-java-chassis 1.x has been deprecated in Sep. last year. All following releases are 2.x. I will go to check with them whether they are going to enhance this plugin, or we are going to remove them.

test/plugin/scenarios/guava-eventbus-scenario/support-version.list
wallezhang: OK, I will pick some versions. All versions tests have passed.

wu-sheng:
@wallezhang According to https://github.com/apache/servicecomb-java-chassis/issues/2714#issuecomment-1033514042, I am removing servicecomb plugin 0.x, so you don't need to fix that case.

wallezhang:
> @wallezhang According to [apache/servicecomb-java-chassis#2714 (comment)](https://github.com/apache/servicecomb-java-chassis/issues/2714#issuecomment-1033514042), I am removing servicecomb plugin 0.x, so you don't need to fix that case.

OK, Thanks a lot.

wu-sheng:
Please sync when you fixed the case. That plugin is removed.

wallezhang:
@wu-sheng ServiceComb 1.x plugin also have the problem. According to the discussion in apache/servicecomb-java-chassis#2714, the 1.x ServiceComb is still in using for end users.

wu-sheng:
Yes, correct. So we need to take a look at how the new segments look like after guava plugin added.

wallezhang:
> Yes, correct. So we need to take a look at how the new segments look like after guava plugin added.

I will take a look.

wallezhang:
@wu-sheng ServiceComb v1.0.0 uses the native Guava EventBus, so the guava-event-bus plugin will affect the test results. Since version 1.1.0, ServiceComb has implemented EventBus by itself, so the scenario test cases after version 1.1.0 can match the previous expected data. 

So shall we delete support version 1.0.0 from scenario test?

wu-sheng:
Yes, go ahead.

wu-sheng (approved):
I am good once all tests passed.

PR #113, opened 2022-02-28
Add witness classes to graphql and vertx-core-3.x plugins 

BFergerson:
Added witness classes to graphql and vertx-core-3.x plugins.

`graphql.execution.ExecutionPath` which is available for provided plugins:
https://github.com/graphql-java/graphql-java/blob/v8.0/src/main/java/graphql/execution/ExecutionPath.java
https://github.com/graphql-java/graphql-java/blob/v9.0/src/main/java/graphql/execution/ExecutionPath.java
https://github.com/graphql-java/graphql-java/blob/v12.0/src/main/java/graphql/execution/ExecutionPath.java

Is not available for v16+:
https://github.com/graphql-java/graphql-java/blob/v16.0/src/main/java/graphql/execution/ExecutionPath.java

---

`io.vertx.core.http.impl.WebSocketFrameFactoryImpl` which is available for beginning to end of 3.x
https://github.com/eclipse-vertx/vert.x/blob/3.0.0/src/main/java/io/vertx/core/http/impl/WebSocketFrameFactoryImpl.java
https://github.com/eclipse-vertx/vert.x/blob/3.9.12/src/main/java/io/vertx/core/http/impl/WebSocketFrameFactoryImpl.java

Is not available for 4.x:
https://github.com/eclipse-vertx/vert.x/blob/4.0.0/src/main/java/io/vertx/core/http/impl/WebSocketFrameFactoryImpl.java

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Release Notes.
 * Support HikariCP Connection pool metrics collecting.
 * Support Dbcp2 Connection pool metrics collecting.
 * Ignore the synthetic constructor created by the agent in the Spring patch plugin.
+* Add witness class for vertx-core-3.x plugin.
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Release Notes.
 * Support Dbcp2 Connection pool metrics collecting.
 * Ignore the synthetic constructor created by the agent in the Spring patch plugin.
 * Add witness class for vertx-core-3.x plugin.
+* Add witness class for graphql plugin.
 
 #### Documentation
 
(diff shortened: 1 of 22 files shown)


wu-sheng:
Are these plugins will throw exception or leak tracecontext for higher versions? Such as target classes still have the same method signature but very difficult logic inside?

wu-sheng (approved):
I checked, these witness classes make sense. Would you plan to contribute new plugins for newer versions?

BFergerson:
I'm not too sure what your first question means. If you're asking if they will make future versions fail, I would say they are already doing that. I'm using Vert.x 4.x and Graphql 17.0 and getting exceptions whenever I use them.

For your second question, I'll be adding support for Vert.x 4.x. I'm not very familiar with the Graphql library/plugin, but I'll see if I can add that too.

wu-sheng:
> I'm not too sure what your first question means. If you're asking if they will make future versions fail, I would say they are already doing that. I'm using Vert.x 4.x and Graphql 17.0 and getting exceptions whenever I use them.

Yes, that was my question, and after I checked the codes in my morning, I had known this breaking exists already. Then I merged this PR before you wake up :)

BFergerson:
I haven't had my coffee yet. It took me another read before I knew what you asked lol. Thanks for merging

wu-sheng:
> I haven't had my coffee yet. It took me another read before I knew what you asked lol. Thanks for merging

:P Go for your coffee, I am feeling I had too many today

PR #114, opened 2022-03-01
fix the bug that maybe generate multiple trace when invoke http reque…

zimmem:
### Fix the bug that maybe generate multiple trace when invoke http request by spring webflux webclient.
- [x ] Explain briefly why the bug exists and how to fix it.  
     spring webflux is base on  async framework [reactor](https://projectreactor.io/) , and  the webclient create exit span without restore Skywalking context from ContextSnapshot. This may generate multiple trace.   
     This commit set Skywalking ContextSnapshot  into Reactor Context when enter Spring WebFlux,  and  get ContextSnapshot from Reactor Context then call ContextManager.continue before construct webcliet request.

- [x ] Update the [`CHANGES` log](https://github.com/apache/skywalking-java/blob/main/CHANGES.md).

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.apm.plugin.spring.webflux.v5.webclient;
 
 import org.apache.skywalking.apm.agent.core.context.ContextCarrier;
 import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.context.ContextSnapshot;
 import org.apache.skywalking.apm.agent.core.context.tag.Tags;
 import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
 import org.apache.skywalking.apm.agent.core.context.trace.SpanLayer;
@@ -34,40 +35,13 @@ import reactor.core.publisher.Mono;
 
 import java.lang.reflect.Method;
 import java.net.URI;
+import java.util.Optional;
 
 public class WebFluxWebClientInterceptor implements InstanceMethodsAroundInterceptorV2 {
 
     @Override
     public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInvocationContext context) throws Throwable {
-        if (allArguments[0] == null) {
-            //illegal args,can't trace ignore
-            return;
-        }
 
-        ClientRequest request = (ClientRequest) allArguments[0];
-        final ContextCarrier contextCarrier = new ContextCarrier();
-
-        URI uri = request.url();
-        final String requestURIString = getRequestURIString(uri);
-        final String operationName = requestURIString;
-        final String remotePeer = getIPAndPort(uri);
-        AbstractSpan span = ContextManager.createExitSpan(operationName, contextCarrier, remotePeer);
-
-        //set components name
-        span.setComponent(ComponentsDefine.SPRING_WEBCLIENT);
-        Tags.URL.set(span, uri.toString());
-        Tags.HTTP.METHOD.set(span, request.method().toString());
-        SpanLayer.asHttp(span);
-
-        if (request instanceof EnhancedInstance) {
-            ((EnhancedInstance) request).setSkyWalkingDynamicField(contextCarrier);
-        }
-
-        //user async interface
-        span.prepareForAsync();
-        ContextManager.stopSpan();
-
-        context.setContext(span);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DispatcherHandlerHandleMethodInterceptor.java
```diff
@@ -74,7 +75,7 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
         span.prepareForAsync();
         ContextManager.stopSpan(span);
 
-        exchange.getAttributes().put("SKYWALING_SPAN", span);
+        exchange.getAttributes().put("SKYWALKING_SPAN", span);
```
wu-sheng, line 78: Is this typo causing the bug? Or it is just a typo.

zimmem: just a typo

apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
```diff
@@ -77,19 +51,43 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
...
-                }
+        return Mono.subscriberContext().flatMap(ctx -> {
+
+            ClientRequest request = (ClientRequest) allArguments[0];
+            URI uri = request.url();
+            final String operationName = getRequestURIString(uri);
+            final String remotePeer = getIPAndPort(uri);
+            AbstractSpan span = ContextManager.createExitSpan(operationName, remotePeer);
+
+            final Optional<Object> optional = ctx.getOrEmpty("SKYWALKING_CONTEXT_SNAPSHOT");
```
wu-sheng, line 62: Where is this `SKYWALKING_CONTEXT_SNAPSHOT` from? Is it inside WebFlux plugin?

wu-sheng:
I have approved the CI running. Let's see whether this change breaks anything, such as do we need to update some plugin tests.

apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
zimmem: Yes , it is set at apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/DispatcherHandlerHandleMethodInterceptor.java.

zimmem: So, Maybe move spring-webflux-5.x-plugin to  apm-sdk-plugin will more suitable.

wu-sheng: Could you add some comments to describe this? Because this seems lacking context when reading codes?

wu-sheng: > So, Maybe move spring-webflux-5.x-plugin to apm-sdk-plugin will more suitable.

We keep that optional, because in many SpringMVC cases, people don't care about WebFlux tracing. The controller has provided enough context

apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
```diff
@@ -77,19 +51,45 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
...
+        return Mono.subscriberContext().flatMap(ctx -> {
+
+            ClientRequest request = (ClientRequest) allArguments[0];
+            URI uri = request.url();
+            final String operationName = getRequestURIString(uri);
+            final String remotePeer = getIPAndPort(uri);
+            AbstractSpan span = ContextManager.createExitSpan(operationName, remotePeer);
+
+            // get ContextSnapshot from reactor context,  the snapshot is set to reactor context by any other plugin
+            // such as spring-webflux-5.x-plugin
```
wu-sheng, line 63:
```suggestion
            // such as DispatcherHandlerHandleMethodInterceptor in spring-webflux-5.x-plugin
```

apm-sniffer/apm-sdk-plugin/spring-plugins/spring-webflux-5.x-webclient-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/webclient/WebFluxWebClientInterceptor.java
zimmem: > Could you add some comments to describe this? Because this seems lacking context when reading codes?

Sure.  The Comment has added.

zimmem: > > So, Maybe move spring-webflux-5.x-plugin to apm-sdk-plugin will more suitable.
> 
> We keep that optional, because in many SpringMVC cases, people don't care about WebFlux tracing. The controller has provided enough context

mvc-annotation-commons also have some program on webflux project， especially for timing.  Because it create/stop span in sync way.  I  will submit a issue to follow up。

wu-sheng: I mean in those cases, spans are only created when Controller entrance is invoked, it doesn't matter the kernel is running in sync or async, because for most users, logic/codes inside the controller are running in sync mode unless they are using reactor mode.

wu-sheng (approved):
LGTM

sivern:
When an Internetptor thread is created or used by another thread, this problem should exist, right?

PR #118, opened 2022-03-05
Vertx 4 Core Plugin

BFergerson:
### Add an agent plugin to support <framework name>
- [x] Add a logo in [the UI repo](https://github.com/apache/skywalking-rocketbot-ui/tree/master/src/views/components/topology/assets)

Vertx 4.x added native tracing support (https://vertx.io/blog/eclipse-vert-x-4-milestone-1-released/#tracing) making this plugin pretty straightforward.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/vertx-plugins/vertx-core-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/vertx4/SWVertxTracer.java
+++ b/apm-sniffer/apm-sdk-plugin/vertx-plugins/vertx-core-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/vertx4/SWVertxTracer.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.apm.plugin.vertx4;
 
 import io.vertx.core.Context;
 import io.vertx.core.eventbus.Message;
+import io.vertx.core.eventbus.impl.clustered.ClusteredMessage;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.spi.observability.HttpRequest;
 import io.vertx.core.spi.observability.HttpResponse;
@@ -34,6 +35,7 @@ import org.apache.skywalking.apm.agent.core.context.ContextSnapshot;
 import org.apache.skywalking.apm.agent.core.context.tag.Tags;
 import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
 import org.apache.skywalking.apm.agent.core.context.trace.SpanLayer;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
 
 import java.util.HashMap;
@@ -118,8 +120,15 @@ public class SWVertxTracer implements VertxTracer<AbstractSpan, AbstractSpan> {
         } else if (request instanceof Message) {
             Message clientRequest = (Message) request;
 
+            String remotePeer = "localhost";
+            if (clientRequest instanceof ClusteredMessage) {
+                EnhancedInstance enhancedInstance = (EnhancedInstance) clientRequest;
+                if (enhancedInstance.getSkyWalkingDynamicField() != null) {
+                    remotePeer = (String) enhancedInstance.getSkyWalkingDynamicField();
+                }
+            }
             ContextCarrier contextCarrier = new ContextCarrier();
-            AbstractSpan span = toExitSpan(clientRequest.address(), "localhost", contextCarrier, context);
+            AbstractSpan span = toExitSpan(clientRequest.address(), remotePeer, contextCarrier, context);
             SpanLayer.asRPCFramework(span);
 
             return toExitAsyncSpan(context, headers, contextCarrier, span);
--- a/apm-sniffer/apm-sdk-plugin/vertx-plugins/vertx-core-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/vertx4/SWVertxTracer.java
+++ b/apm-sniffer/apm-sdk-plugin/vertx-plugins/vertx-core-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/vertx4/SWVertxTracer.java
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.vertx4;
+
+import io.vertx.core.Context;
+import io.vertx.core.eventbus.Message;
+import io.vertx.core.impl.ContextInternal;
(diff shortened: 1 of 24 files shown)


apm-sniffer/apm-sdk-plugin/vertx-plugins/vertx-core-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/vertx4/SWVertxTracer.java
```diff
@@ -0,0 +1,201 @@
...
+        if (TracingPolicy.IGNORE.equals(policy)) {
+            return null;
+        }
+
+        if (request instanceof HttpRequest) {
+            HttpRequest serverRequest = (HttpRequest) request;
+
+            ContextCarrier contextCarrier = getContextCarrier(headers);
+            AbstractSpan span = toEntrySpan(
+                    "{" + serverRequest.method() + "}" + serverRequest.uri(),
```
wu-sheng, line 60: According to latest operation name style, this should be `method` + `:` + `URI`, referring from 

https://github.com/apache/skywalking-java/blob/2bd58bce7a95e61eac810ecb6340beaa3dfe3892/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java#L263

This is following OpenAPI style, https://skywalking.apache.org/docs/main/v8.9.1/en/setup/backend/endpoint-grouping-rules/#endpoint-name-grouping-by-openapi-definitions

docs/en/setup/service-agent/java-agent/Plugin-list.md
```diff
@@ -117,6 +117,7 @@
 - toolkit-exception
 - undertow-2.x-plugin
 - vertx-core-3.x
+- vertx-core-4.x
```
wu-sheng, line 120: Supported-list.md doc should be updated too.

wu-sheng (approved):
Thanks for adding this so quickly. And I am interested to see `tracer` concept is added in the vertx core.

BFergerson:
@wu-sheng, this is almost ready but there is something off about the UI. I'm seeing `VNode: e3ebfc9ac13e462b8ad982a36a6d6689.175.16464888093840002` with `[Broken] CROSS_PROCESS`, which looks like:
image

I've simplified the test to debug this issue. What the code does is:
- Enters at `GET:/vertx-core-4-scenario/case/core-case`
  - Exits at `/vertx-core-4-scenario/case/executeTest`
    - Enters at `GET:/vertx-core-4-scenario/case/executeTest`
       - Exits at `local-message-receiver`

These all then collapse back down once `local-message-receiver` receives the response.

---

Given the above, why do you think I get the `[Broken] CROSS_PROCESS` issue?

wu-sheng:
> Given the above, why do you think I get the [Broken] CROSS_PROCESS issue?

From the screenshot, this is generally caused by the reference in level 3(`Enters at GET:/vertx-core-4-scenario/case/executeTest (HTTP)` includes a segment ID and a span ID, but no matched span according to the given trace ID. The root cause could be 
1. The reference(from HTTP header) is incorrect, somehow it is set wrong.
2. The segment in the previous service(Level 1 entry and Level 2 exit) is not reported to OAP server. Maybe the span is not closed correctly due to such as HTTP error/timeout,then missing an intercepting/span-closing point

wu-sheng:
> These all then collapse back down

Do you mean the trace is back to expected?

BFergerson:
> > These all then collapse back down
> 
> Do you mean the trace is back to expected?

Vert.x does everything async. I just meant that none of those spans are finished till the final action is complete.

---

Thanks for the tips. I'll take a closer look.

wu-sheng:
> Vert.x does everything async. I just meant that none of those spans are finished till the final action is complete.

Yes, then this is expected. Once a trace is half submitted(due to downstream is not completed), UI should show in this way.

BFergerson:
@wu-sheng, I was still debugging the above issue. I don't think the issue is from the downstream not being completed. I made sure all the spans that were opened were closed and waited a few minutes for the UI to update, which it doesn't.

The issue seems to be coming from my usage of `ContextManager.continued()`. If I remove that, the error goes away but there appears to be less linkage happening between the spans. I'm still looking into this but if you have tips on how to properly use that, I'd appreciate them.

wu-sheng:
I saw the expected data verified, so, I thought the link exists.

`ContextManager.continued()` is not the root cause, as it is just doing `CROSS_THREAD`, but you are facing `CROSS_PROCESS` breaking. You could debug and check whether it is in the `snapshot`, and what is in the HTTP header, because actually you are facing `CROSS_PROCESS` breaking, but the `ContextManager.continued()` is only relative to `CROSS_THREAD`

wu-sheng:
Check the HTTP header and reference of `level-3` exit span received. I think you will not find the existing trace ID and segment ID in the database.

wu-sheng:
>  made sure all the spans that were opened were closed and waited a few minutes 

I think you just thought all spans are closed, but actually some are not. You need to debug to get all span created with a specific trace ID, then you could know which spans are collected, and which are really closed.

wu-sheng:
If there are too many threads, try to print all span creating and span closing with trace ID, segment ID and span ID. Then check the breaking segment with the trace id. Then you may notice which spans are missed.

BFergerson:
> If there are too many threads, try to print all span creating and span closing with trace ID, segment ID and span ID. Then check the breaking segment with the trace id. Then you may notice which spans are missed.
> 
> Logs like this would help
> 
> ```
> TraceID, xxxx, SegmentID, xxxx, SpanID, xxxxx, created.
> TraceID, xxxx, SegmentID, xxxx, SpanID, xxxxx, closed. 
> TraceID, xxxx, SegmentID, xxxx, SpanID, xxxxx, referring to TraceID yyyy, SegmentID yyyy, SpanID yyyy
> ```

Thanks, adding logging was useful. I also refactored `GlobalIdGenerator` to return simple reproducible outputs. That made it almost impossible not to see the issue. Maybe you've thought about having reproducible ids for testing scenarios? It might help given my fix (https://github.com/apache/skywalking-java/pull/119) doesn't change the expected data.

wu-sheng:
Thanks for the fixing. Could byou be more specific about ID generator refactoring?

BFergerson:
I mean it would be nice to have reproducible ids. So no matter how many times I run SkyWalking, it always returns the same segments/trace ids as long as I do exactly the same processing. As far as I know, this is called "reproducible random". I hacked it together like this https://github.com/BFergerson/skywalking-java/commit/310c097bd1450b076dd14446eb01ec07dbae3ab0, but it could be done smarter.

This made it easy to debug since segment/trace ids always started at 1 and incremented each time I debugged. Eventually, I realized each span was getting a new segment/trace id and that was the problem to solve. When I had it fixed, the trace ids where all the same number.

It might make sense to use a reproducible random for test scenarios so you could validate the expected segment/trace ids. Just a thought.

wu-sheng:
I am thinking support using x-request-id from HTTP request header as trace ID. WDYT?

BFergerson:
I'm not too aware of what that would mean for the reproducibility of the ids. If using that means you will always get the same results for the same processing, then I think it could work.

The main goal is reproducibility as that would allow test scenarios to declare which segment/trace ids are expected. Doing so might point out how some existing test scenarios are actually producing the wrong expected results. It was only via the UI that I realized I was doing it wrong.

wu-sheng:
Yes, I followed your point. x-request-id is a standard HTTP header, which means when request to the target URL, you actually could control the request, which is your required prediction, right?

BFergerson:
I guess I would need to see how it's implemented. I just keep thinking of how there might be situations that might not work for like TCP connections. It just also seems like that would make the test scenarios more complicated.

wu-sheng:
Yes, this is for HTTP, but all test cases expose HTTP endpoint for test framework only. 
For automatical tests, I think the tricky point, which ID(assume we know the value sequence) should we write in expected file?

wu-sheng:
I think we should open an issue to discuss this? It seems we are missing this key feature in plugin tests for years.

BFergerson:
> I think we should open an issue to discuss this? It seems we are missing this key feature in plugin tests for years.

Yeah, okay. The result of that issue should likely be to deprecate testing with:
```yml
segmentId: not null
traceId: not null
```

And that seems to be all the tests.

PR #120, opened 2022-03-05
Renamed graphql-12.x-plugin to graphql-12.x-15.x-plugin

BFergerson:
Looking into creating additional support and wondering why it kept using version 9 instrumentation.

Calling this `graphql-12.x-scenario` doesn't make much sense btw since it actually tests 12-15.


BFergerson:
Went ahead and renamed plugin/scenario in preparation for adding additional plugin/scenario.

apm-sniffer/apm-sdk-plugin/graphql-plugin/graphql-12.x-15.x-plugin/src/main/resources/skywalking-plugin.def
```diff
@@ -14,4 +14,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-graphql=org.apache.skywalking.apm.plugin.graphql.v12.define.GraphqlInstrumentation
+graphql=org.apache.skywalking.apm.plugin.graphql.v12tov15.define.GraphqlInstrumentation
```
wu-sheng, line 17: Should we change the key?

wu-sheng: Like Vertx plugin, we have `vertx-core-3.x` for version 3.x. It is better we could have graphql version in the key. Notice, once you change, the Plugin-list.md should be changed too.

wu-sheng, line 17:
```suggestion
graphql-12.x-15.x=org.apache.skywalking.apm.plugin.graphql.v12tov15.define.GraphqlInstrumentation
```

wu-sheng (dismissed):
LGTM, waiting for tests passed.

docs/en/setup/service-agent/java-agent/Plugin-list.md
```diff
@@ -25,6 +25,7 @@
 - feign-pathvar-9.x
 - finagle
 - graphql
```
wu-sheng, line 27:
```suggestion
```

I think you forgot to remove this.

wu-sheng (approved):

docs/en/setup/service-agent/java-agent/Plugin-list.md
BFergerson: There is still the older versions, graphql 8 and 9

docs/en/setup/service-agent/java-agent/Plugin-list.md
```diff
@@ -24,7 +24,7 @@
 - feign-default-http-9.x
 - feign-pathvar-9.x
 - finagle
-- graphql
```
wu-sheng, line 27: I removed this, but it breaks the CI. Maybe there are some other plugin keys should be updated?

docs/en/setup/service-agent/java-agent/Plugin-list.md
BFergerson: Guessing you would like those keys updated too?

wu-sheng (approved):

PR #122, opened 2022-03-06
test: support to the custom base image by configuration

dmsolr:

### DIFF
--- a/.github/workflows/plugins-jdk14-test.0.yaml
+++ b/.github/workflows/plugins-jdk14-test.0.yaml
@@ -47,7 +47,8 @@ jobs:
       - name: Build
         uses: ./.github/actions/build
         with:
-          java_version: jdk14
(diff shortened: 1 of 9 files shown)


.github/workflows/plugins-jdk14-test.0.yaml
```diff
@@ -47,7 +47,8 @@ jobs:
       - name: Build
         uses: ./.github/actions/build
         with:
-          java_version: jdk14
+          base_image_java: adoptopenjdk/openjdk14:alpine
+          base_image_tomcat: tomcat:8.5-jdk14-openjdk
```
wu-sheng, line 51: @wutingjia After this PR, you could use this way in a new `plugins-tomcat10-test.yaml` to make sure your test service running in the real Tomcat10 container as an add-on of Tomcat embed. As I mentioned, they are different in runtime.

wu-sheng (approved):

PR #124, opened 2022-03-14
Mono.subscribeOn(Scheduler) method not yielding span

sivern:
Since the subscribeOn(Scheduler) method of webflux uses a separate thread pool for processing, no trace will be generated without enabling jdk-threading-plugin, and multiple traces may be generated when jdk-threading-plugin is enabled. 
E.g:
```java
import reactor.core.publisher.Mono;
import reactor.core.scheduler.Schedulers;

   @GetMapping("/test")   
    public Mono<Object> subscribeOn(String param) {   
       return   Mono.just(param).subscribeOn(Schedulers.boundedElastic()).map(parm-> {return db.selectOne(param);})   
   }
````
The above test code will generate two traces when using jdk-threading-plugin, including http["/test"] trace and Thread[db] trace. When jdk-threading-plugin is not used, only one http[" /test"] trace, lost trace of db 

### CODE REVIEW

**Comment by wu-sheng** [comment]
If you want to submit a pull request, please provide a full description and test scenarios should be updated. 
Document is much more important, at here, I can't know what is going on here.

**Comment by sivern** [comment]
> If you want to submit a pull request, please provide a full description and test scenarios should be updated. Document is much more important, at here, I can't know what is going on here.

Ok,I will submit later

File: test/plugin/scenarios/webflux-scenario/webflux-projectB-scenario/src/main/java/test/apache/skywalking/apm/testcase/sc/webflux/projectB/controller/DbOperate.java [hunk 1]
```diff
@@ -0,0 +1,22 @@
...
+
+import org.springframework.stereotype.Service;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+
+/**
+ * @ClassName
+ * @Description TODO
+ * @Author guoxiwen
+ * @Date 2022/3/14 18:45
```
wu-sheng, line 11: This is not expected.

test/plugin/scenarios/webflux-scenario/webflux-projectB-scenario/src/main/java/test/apache/skywalking/apm/testcase/sc/webflux/projectB/controller/TestAnnotationController.java
```diff
@@ -59,4 +60,12 @@ public String webclientServer() {
     public Mono<String> hello(@RequestBody(required = false) String body) {
         return Mono.just("Hello World");
     }
+
+    @RequestMapping("/testcase/subscribeOn/elastic")
+    public Mono<Mono<String>> boudElastic(@RequestBody(required = false) String body){
```
wu-sheng: The expect file should be updated according to latest traces.

wu-sheng (commented):
Please follow the plugin test doc, nd run the tests locally, provide the output screenshot once it is done.

test/plugin/scenarios/webflux-scenario/webflux-projectB-scenario/src/main/java/test/apache/skywalking/apm/testcase/sc/webflux/projectB/controller/DbOperate.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+@Service
+public class DbOperate {
+    // This method is to generate an annotation trace for testing
+    public String selectOne(String param) {
+        return "hello"+param;
```
wu-sheng, line 12: You need to follow code style we provided for all codes.

test/plugin/scenarios/webflux-scenario/webflux-projectB-scenario/src/main/java/test/apache/skywalking/apm/testcase/sc/webflux/projectB/controller/DbOperate.java
```diff
@@ -0,0 +1,15 @@
+package test.apache.skywalking.apm.testcase.sc.webflux.projectB.controller;
+
+import org.springframework.stereotype.Service;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+
+
+@Service
+public class DbOperate {
+    // This method is to generate an annotation trace for testing
```
wu-sheng, line 10: What does this method? Why doesn't a DBOperate include a real database access?

sivern: In fact, after being processed by the Schedulers.elastic() method, all traces will be split into a single piece, so an Annotiation trace can be used here to simulate the subsequent span split. Later I will modify to h2 based DB access.

sivern: image

sivern:
> 请按照插件测试文档，在本地运行测试，完成后提供输出屏幕截图。

image

test/plugin/scenarios/webflux-scenario/config/expectedData.yaml
```diff
@@ -38,6 +38,27 @@ segmentItems:
...
+      componentId: 67
+      isError: false
+      spanType: Entry
+      peer: ''
+      tags:
+      - {key: url, value: 'http://localhost:18080/testcase/subscribeOn/elastic'}
+      - {key: http.method, value: GET}
+      - {key: http.status_code, value: '200'}
+      refs:
+      - {parentEndpoint: GET:/projectA/testcase, networkAddress: 'localhost:18080', refType: CrossProcess,
```
wu-sheng, line 58: 2 things.
1. About test output, I mean following [this doc](https://skywalking.apache.org/docs/skywalking-java/v8.9.0/en/setup/service-agent/java-agent/plugin-test/), run plugin tests locally, and verify this expectedData.yaml.
2. This is a new entry span, where is the exit span? I am confused about this.

wu-sheng:
@sivern Any update?

sivern:
> @sivern任何更新？

I'm watching a test document to update my test case

apm-sniffer/optional-plugins/optional-spring-plugins/spring-webflux-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/webflux/v5/ScheduledScalarRunMethodInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+        AbstractSpan span = ContextManager.createLocalSpan(generateOperationName(objInst, method));
+        final Object storedField = objInst.getSkyWalkingDynamicField();
+        if (storedField != null) {
+            final ContextSnapshot contextSnapshot = (ContextSnapshot) storedField;
+            ContextManager.continued(contextSnapshot);
+        }
+        span.setComponent(ComponentsDefine.JDK_THREADING);
```
wu-sheng, line 41: This is for the JDK plugin, which you are not.

sivern:
> @sivern Any update?

image
I would like to ask, when running a test case for webflux-scenario, why go to the directory of /bootstrap-plugins to look for apm-spring-webflux-5.x-plugin-*.jar that should be under /optional-plugins?

wu-sheng:
`optional plugin` means wouldn't be activated, but some test scenarios require that, so it use `configuration.yaml` to activate them.

https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/jdk-http-scenario/configuration.yml

You are writing webflux plugin, it can't be in optional plugin folder.

wu-sheng:
@sivern Any update?

wu-sheng:
No update in a week, close for now.

PR #126, opened 2022-03-22
[Bug] SkyWalking Java agent okhttp plugin Wrong url used to create span #8695

PepoRobert:


wu-sheng:
Please fix CI and make sure the plugin tests could verify this case. Also, you should add a full description about what is bug scenario, and why this could fix it.

wu-sheng:
It is better if you have verified cases in https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/okhttp-scenario. And add this unsupported scenario into that cases with new expected spans.

PepoRobert:
> It is better if you have verified cases in https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/okhttp-scenario. And add this unsupported scenario into that cases with new expected spans.
image

This is the sequence diagram of the synchronous call, with the new interceptor in red. You can see that the original interceptor was intercepted twice
这是同步调用的时序图 ,红色部分为这次新增的拦截器 .可以看到原先的是拦截了两次。

Using the chain of responsibility pattern to process requests and allowing the user to customize interceptors allowing the user to rewrite requests interceptors fall into two further categories
okhttp使用责任链模式处理请求。并且允许用户自定义拦截器。允许用户重写重新请求。拦截器又被分为两类。[官方说明](https://square.github.io/okhttp/features/interceptors/)
image

I think agent interception should be at least after the user can customize the application of interceptors
我认为agent拦截至少应该在用户可以自定义的应用拦截器之后

PR #105, opened 2022-01-30
Feature/support tomcat thread pool

xu1009:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Close [#<8491>.](https://github.com/apache/skywalking/issues/8491)
- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking-java/blob/main/CHANGES.md).

PR #128, opened 2022-03-24
Make sure Tomcat thread plugin working in Tomcat 9

wu-sheng:

@xu1009 We found a cast exception in Tomcat 9. We supported Tomcat 9 as plugin test container, could you add another test?

Link to https://github.com/apache/skywalking-java/pull/105

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/tomcat-thread-pool-plugin/src/main/java/org/apache/skywalking/apm/plugin/tomcat/thread/pool/TomcatThreadExecutorInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/tomcat-thread-pool-plugin/src/main/java/org/apache/skywalking/apm/plugin/tomcat/thread/pool/TomcatThreadExecutorInterceptor.java
@@ -26,7 +26,7 @@ import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceC
 public class TomcatThreadExecutorInterceptor implements InstanceConstructorInterceptor {
 
     private static final String METER_NAME = "thread_pool";
-    private static final String METRIC_POOL_NAME_TAG_NAME = "metric_type";
+    private static final String METRIC_POOL_NAME_TAG_NAME = "pool_name";
     private static final String THREAD_POOL_NAME = "tomcat_execute_pool";
     private static final String METRIC_TYPE_TAG_NAME = "metric_type";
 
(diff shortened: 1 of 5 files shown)


wu-sheng:
@xu1009 And I am not sure whether this fix works, `org.apache.tomcat.util.threads.ThreadPoolExecutor`'s hierarchy tree is changed from 8 to 9.

wu-sheng:
This doesn't work.

```
  java.lang.LinkageError: loader (instance of  org/springframework/boot/loader/LaunchedURLClassLoader): attempted  duplicate class definition for name: "org/apache/tomcat/util/threads/ThreadPoolExecutor"
  	at java.lang.ClassLoader.defineClass1(Native Method) ~[?:1.8.0_322]
  	at java.lang.ClassLoader.defineClass(ClassLoader.java:756) ~[?:1.8.0_322]
  	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142) ~[?:1.8.0_322]
```

kezhenxu94 (approved):

apm-sniffer/apm-sdk-plugin/tomcat-thread-pool-plugin/src/main/java/org/apache/skywalking/apm/plugin/tomcat/thread/pool/TomcatThreadExecutorInterceptor.java
```diff
@@ -18,37 +18,40 @@
...
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+import org.apache.tomcat.util.threads.ThreadPoolExecutor;
 
 public class TomcatThreadExecutorInterceptor implements InstanceConstructorInterceptor {
 
+    private static final String METER_NAME = "thread_pool";
+    private static final String METRIC_POOL_NAME_TAG_NAME = "pool_name";
     private static final String THREAD_POOL_NAME = "tomcat_execute_pool";
-    private static final Map<String, Function<ThreadPoolExecutor, Supplier<Double>>> METRIC_MAP = new HashMap<String, Function<ThreadPoolExecutor, Supplier<Double>>>() {{
```
wu-sheng, line 33: @xu1009 I removed this, but hard code them in the method.(Verified by H2 case adjusted by @kezhenxu94 )

The reason is, `org.apache.tomcat.util.threads.ThreadPoolExecutor` is rewrited between `9.0.21` and `9.0.54`(don't check for more details). It doesn't **extends java.util.concurrent.ThreadPoolExecutor** anymore. Then we faced type casting exception.

image

Ref to 9.0.52, https://github.com/apache/tomcat/blob/9.0.52/java/org/apache/tomcat/util/threads/ThreadPoolExecutor.java

wu-sheng: If I use  `org.apache.tomcat.util.threads.ThreadPoolExecutor` in this static block, the JVM class loader will fail when loading this interceptor, which leaves me no choice but to move all logic inside the `onConstruct` method.

kezhenxu94 (approved):

wu-sheng:
> Test / h2-scenario (pull_request) Successful in 1m

This is passed. @kezhenxu94 I think after merge, another showcase update should be able to fix this.

PR #129, opened 2022-03-25
Make sure the parent endpoint from existing first ENTRY span

wu-sheng:

In MQ and some fetching cases, the first span may not be an entry span. This bug exists for years, sorry I forgot to fix this but known this bug for a long time.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -33,6 +33,7 @@ Release Notes.
 * The namespace and cluster would be reported as instance properties, keys are `namespace` and `cluster`. Notice, if
   instance_properties_json includes these two keys, they would be overrided by the agent core.
 * [Breaking Change] Remove the namespace from `cross process propagation` key.
+* Make sure the parent endpoint in tracing context from existing first ENTRY span, rather than first span only.
 
 #### Documentation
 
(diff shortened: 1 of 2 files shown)


wu-sheng:
I am not sure whether I would break existing plugin tests, let's see.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
```diff
@@ -572,4 +569,31 @@ public long createTime() {
...
+            this.name = name;
+        }
+
+        /**
+         * Set endpoint name according to priority
+         */
+        private void set(final boolean isEntry, final String name) {
+            if (!this.isEntry && isEntry) {
+                this.name = name;
+            }
```
kezhenxu94, line 596: I think the `isEntry` should be updated as well.

CHANGES.md
```diff
@@ -33,6 +33,7 @@ Release Notes.
 * The namespace and cluster would be reported as instance properties, keys are `namespace` and `cluster`. Notice, if
   instance_properties_json includes these two keys, they would be overrided by the agent core.
 * [Breaking Change] Remove the namespace from `cross process propagation` key.
+* Make sure the parent endpoint in tracing context from existing first ENTRY span, rather than first span only.
```
kezhenxu94, line 36: I feel tough to understand the meaning of this sentence without reading the codes.

```suggestion
* Make sure the parent endpoint in tracing context is the first ENTRY span, rather than first span only.
```

wu-sheng: `is from` is not correct, because this trace context could only have local/exit spans.

wankai123 (approved):

PR #130, opened 2022-03-25
Remove unnecessary InterruptedException catch

Oliverwqcwrw:

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/commands/CommandService.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/commands/CommandService.java
@@ -73,6 +73,8 @@ public class CommandService implements BootService, Runnable {
                 serialNumberCache.add(command.getSerialNumber());
             } catch (InterruptedException e) {
                 LOGGER.error(e, "Failed to take commands.");
+                // restore the interrupted status
+                Thread.currentThread().interrupt();
             } catch (CommandExecutionException e) {
                 LOGGER.error(e, "Failed to execute command[{}].", e.command().getCommand());
             } catch (Throwable e) {
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/commands/CommandService.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/commands/CommandService.java
@@ -71,11 +71,6 @@ public class CommandService implements BootService, Runnable {
 
                 commandExecutorService.execute(command);
                 serialNumberCache.add(command.getSerialNumber());
-            } catch (InterruptedException e) {
-                LOGGER.error(e, "Stop waiting new commands.");
-                // restore the interrupted status
-                Thread.currentThread().interrupt();
-                break;
             } catch (CommandExecutionException e) {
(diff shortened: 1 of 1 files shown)


wu-sheng:
What is the scenario of doing this?

Oliverwqcwrw:
the main purpose is let higher layer code know the thread state and can response for interrupted

wu-sheng:
Which thread do you mean?

Oliverwqcwrw:
the  throwing interruptedexception  thread

wu-sheng:
I don't think there is really one thread is going to interrupt. This is just a protection mechanism from coding API requirements, nothing more.

Oliverwqcwrw:
I think if the other thread interrupt the current thread ,we should return interrupted status to caller ,but not do nothing

wu-sheng:
My point is, there should be no such case. So, I can't be sure, interrupting current thread could be helpful.

kezhenxu94:
> I think if the other thread interrupt the current thread ,we should return interrupted status to caller ,but not do nothing

I agree with "we should return interrupted status to caller", but don't agree with "but not do nothing", interrupts can be also used as a signal between threads, especially in this case, we just use the interrupt as a signal that "we should not keep waiting for the command queue anymore (for example the process is about to terminate)".

"return interrupted status to the caller" is a better practice if you don't know who is the caller and how will they handle the interrupt status (for example in the framework / SDK).

In this particular PR, because the `BaseCommand command = commands.take();` is inside a `while` loop, you cannot simply set the interrupt signal, which will result in an endless loop when it comes to `commands.take()` and it throws, you need to add a `break` inside `catch (InterruptedException e) {`

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/commands/CommandService.java
```diff
@@ -73,6 +73,8 @@ public void run() {
                 serialNumberCache.add(command.getSerialNumber());
             } catch (InterruptedException e) {
                 LOGGER.error(e, "Failed to take commands.");
+                // restore the interrupted status
+                Thread.currentThread().interrupt();
```
kezhenxu94, line 77:
```suggestion
                LOGGER.error(e, "Stop waiting new commands.");
                // restore the interrupted status
                Thread.currentThread().interrupt();
                break;
```

Oliverwqcwrw:
you are right,I am omiting the loop

kezhenxu94 (dismissed):

wu-sheng:
I am still not following why we need to `set the flag`.

> Thread.interrupt() sets the interrupted status/flag of the target thread to true which when checked using Thread.interrupted() can help in stopping the endless thread.

This is what I understand, please correct me if I am wrong. The reason we catch this exception is, we don't want anything to close this thread, except JVM is existing(then it exists from `#shutdown`)

____

I took another look of this method block, it is enough OK to remove this catch. As InterruptedException is an exception, not a RuntimeException, so, it means there is no case to intercept. 

___

Could either of you explain what is really going on? Basically, I am still not following.

wu-sheng (changes requested):
TBD

Oliverwqcwrw:
If there does not have other thread  possibly interrupt current thread  ,I think removing catch block is ok

wu-sheng:
> If there does not have other thread possibly interrupt current thread

No, there isn't, which was why I keep asking.

wu-sheng (approved):

PR #132, opened 2022-03-29
optimize charset

Oliverwqcwrw:
In `org.apache.skywalking.apm.toolkit.opentracing.ByteBufferContext` 

Optimized acquisition charset

### DIFF
--- a/apm-application-toolkit/apm-toolkit-opentracing/src/main/java/org/apache/skywalking/apm/toolkit/opentracing/ByteBufferContext.java
+++ b/apm-application-toolkit/apm-toolkit-opentracing/src/main/java/org/apache/skywalking/apm/toolkit/opentracing/ByteBufferContext.java
@@ -21,11 +21,12 @@ package org.apache.skywalking.apm.toolkit.opentracing;
 import io.opentracing.SpanContext;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
 
 public class ByteBufferContext implements SpanContext {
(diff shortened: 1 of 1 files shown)


apm-application-toolkit/apm-toolkit-opentracing/src/main/java/org/apache/skywalking/apm/toolkit/opentracing/ByteBufferContext.java
```diff
@@ -21,11 +21,12 @@
 import io.opentracing.SpanContext;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
 
 public class ByteBufferContext implements SpanContext {
-    static final Charset CHARSET = Charset.forName("UTF-8");
+    static final Charset CHARSET = StandardCharsets.UTF_8;
```
wu-sheng, line 29: UTF-8 is a built-in Charset. What is this `optimization` actually about?

Superskyyy: > UTF-8 is a built-in Charset. What is this `optimization` actually about?

I think it's more of a chore/styling-wise change. The performance is exactly the same.

wu-sheng: This is a static field, only loaded once in class loader.
So, I doubt the performance impact.

Any benchmark you could share? It is hard to be specific, I think, maybe only nano second difference?

PR #134, opened 2022-03-31
Feature msg business consumer

kebingxing:
### support trace the real result of kafka message processing at the business side
- [x] If this is non-trivial feature, paste the links/URLs to the design doc.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<no>.

### instruction
Previously, kafka and other message consumers could only trace the messages pulled by the clients, but the actual processing results of the business could not be traced. Our project had a large number of message processing scenarios, Tracking business message processing results can improved our troubleshooting efficiency~
1.before:
image
2. after
image
After the message is processed by the business side, can customize record the processing results
-- methods as follows
TraceMsgContext.ConsumerMsgSucceed
TraceMsgContext.ConsumerMsgFailure
image

It is my honor to contribute to the skywalking. Thank you for taking the time to read it, and I hope you can give me some suggestions. If there is any mistake, I will adjust it in time

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/KafkaConsumerInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/KafkaConsumerInterceptor.java
@@ -18,6 +18,8 @@
 
 package org.apache.skywalking.apm.plugin.kafka;
 
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.header.Header;
@@ -33,9 +35,11 @@ import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInt
 import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
 import org.apache.skywalking.apm.plugin.kafka.define.Constants;
 import org.apache.skywalking.apm.plugin.kafka.define.KafkaContext;
+import org.apache.skywalking.apm.toolkit.trace.msg.TraceMsg;
 
 import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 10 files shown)


apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/KafkaConsumerInterceptor.java
```diff
@@ -44,6 +48,17 @@
 
     public static final String OPERATE_NAME_PREFIX = "Kafka/";
     public static final String CONSUMER_OPERATE_NAME = "/Consumer/";
+    private static final Gson GSON = new Gson();
+
+    private static boolean TRACE_MSG_OPEN;
+
+    public KafkaConsumerInterceptor() {
+        try {
+            Class.forName("org.apache.skywalking.apm.toolkit.trace.msg.TraceMsg");
```
wu-sheng, line 57: This is totally wrong.

wu-sheng (changes requested):
If you want to propose a new API and a new feature, you should submit a discussion first. 
Your changes here are not acceptable for upstream, and wrong from tech perspective too.

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/KafkaConsumerInterceptor.java
kebingxing: Thank you for your guidance.
The original target here, The KafkaConsumerInterceptor can automatically stores trace information into the message that are implemented the TraceMsg interface.

I will optimize this code;  The pseudo-code used is as follows:
1.  String trace=TraceMsgContext.getCurretTrace();    //provided to obtain trace information
2.  msg.setTraceData(trace);  // customized trace data  in message body
3. TraceMsgContext.ConsumerMsgSucceed(msg.getTraceData()) //trace the real result of kafka message processing

In addition, for the scenarios supported by this feature, whether it is valuable~? 
I submit a discussion  later;

Thank you again for your guidance; and Thank SkyWalking for bringing a lot of value to the project;

wu-sheng: Please submit a discussion to sync the case you faced and try to resolve.
Then we could move on tech details, finally as a pull request.
Right now, I am confused with this PR.

PR #135, opened 2022-04-01
add SAMPLE_AVG_CPU_LOAD_LIMIT config

zhyyu:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<[8773](https://github.com/apache/skywalking/issues/8773)>.


wu-sheng:
Take a look at how `TraceIgnoreExtendService` extends from `SamplingService`. I think it would be more elegant.
This doesn't have to change from the core level.

Another thing is, you add another thread just for CPU usage, which concerns me much. The agent is very sensitive about a new thread. It should be more reasonable to ask `JVMService` holds the latest CPU value, and provide a method for this new plugin.

zhyyu:
> 

Ok, I will try it later

zhyyu:
> Take a look at how `TraceIgnoreExtendService` extends from `SamplingService`. I think it would be more elegant. This doesn't have to change from the core level.
> 
> Another thing is, you add another thread just for CPU usage, which concerns me much. The agent is very sensitive about a new thread. It should be more reasonable to ask `JVMService` holds the latest CPU value, and provide a method for this new plugin.

Ok, I will try it later using `JVMService,` add a new plugin like `trace-ignore-plugin`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMService.java
```diff
@@ -104,8 +106,16 @@ public void run() {
             jvmBuilder.setClazz(ClassProvider.INSTANCE.getClassMetrics());
 
             sender.offer(jvmBuilder.build());
+
+            // refresh cpu usage percent
+            cpuUsagePercent = jvmBuilder.getCpu().getUsagePercent();
```
wu-sheng, line 111: I think `getUsagePercent` again would cause an issue. Have you tested?

apm-sniffer/optional-plugins/trace-cpu-limit-plugin/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>optional-plugins</artifactId>
+        <groupId>org.apache.skywalking</groupId>
+        <version>8.10.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-trace-cpu-limit-plugin</artifactId>
```
wu-sheng, line 27:
```suggestion
    <artifactId>trace-sampler-cpu-policy-plugin</artifactId>
```

apm-sniffer/optional-plugins/trace-cpu-limit-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/cpu/limit/conf/CpuLimitConfigInitializer.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Properties;
+import org.apache.skywalking.apm.agent.core.boot.AgentPackageNotFoundException;
+import org.apache.skywalking.apm.agent.core.boot.AgentPackagePath;
+import org.apache.skywalking.apm.agent.core.conf.ConfigNotFoundException;
+import org.apache.skywalking.apm.agent.core.logging.api.ILog;
+import org.apache.skywalking.apm.agent.core.logging.api.LogManager;
+import org.apache.skywalking.apm.util.ConfigInitializer;
+import org.apache.skywalking.apm.util.PropertyPlaceholderHelper;
+
+public class CpuLimitConfigInitializer {
```
wu-sheng, line 35: This is the wrong way to initialize a simple value config on your own initializer. The ignore plugin has its own defined config file, so, it does in this way. You just need a threshold, so, you could use the standard way and share the `agent.config` to load the feature.

Take a look at `SpringMVCPluginConfig`.

apm-sniffer/optional-plugins/pom.xml
```diff
@@ -44,6 +44,7 @@
     <modules>
         <module>optional-spring-plugins</module>
         <module>trace-ignore-plugin</module>
+        <module>trace-cpu-limit-plugin</module>
```
wu-sheng, line 47: Add the last of the list.

apm-sniffer/config/apm-trace-cpu-limit-plugin.config
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# If the operation name of the first span is matching, this segment should be ignored
+#  ant path match style
+#  /path/?   Match any single character
+#  /path/*   Match any number of characters
+#  /path/**  Match any number of characters and support multilevel directories
+#  Multiple path comma separation, like trace.ignore_path=/eureka/**,/consul/**
```
wu-sheng, line 22: Copy/paste issue.

wu-sheng:
I have commented on the codes, and I think you haven't tried this in the real case, please do so before submitting codes.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMService.java
zhyyu: I have tested it using debug, it returned a val less than 100.

apm-sniffer/optional-plugins/trace-cpu-limit-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/cpu/limit/conf/CpuLimitConfigInitializer.java
zhyyu: ok

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMService.java
zhyyu: What issue do you think will cause, maybe my debug way miss it.

wu-sheng: You are using a `#build`-ed builder to read data. I think it is much safe to read it from `build` result or create a local variable to host the value.

zhyyu:
> I have commented on the codes, and I think you haven't tried this in the real case, please do so before submitting codes.

I just used idea to debug the agent code, maybe not production scenario... But it do sent the segment data to OAP. I will change the code according to the comment above.

wu-sheng:
As you are adding a CPU based sampler, so, you should verify whether this is really working.
The code review only could find out some issues, but can't approve this could work as expected. Also, this plugin doesn't like other plugins, we can't test it automatically.

So, please try to establish a runtime env to verify this plugin.

wu-sheng:
Any update?

zhyyu:
> Any update?

Recently my work is busy, I will update it when I am free

zhyyu:
image

I create a thread run `while(true) `code, to increase the cpu usage to 15% (124/ 8核)

and then debug the `trySampling` code

image

my config is 10% cpu usage limit, 15% cpu usage is above the limit, `trySampling` return false which meet the expectation.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/JVMService.java
```diff
@@ -48,6 +48,8 @@ public class JVMService implements BootService, Runnable {
     private volatile ScheduledFuture<?> sendMetricFuture;
     private JVMMetricsSender sender;
```
wu-sheng, line 50: I think you don't need an empty line here.

wu-sheng:
Please update `changes.md` and optional plugin doc with `agent.config` accordingly.

apm-sniffer/config/agent.config
```diff
@@ -270,3 +270,5 @@ plugin.neo4j.trace_cypher_parameters=${SW_PLUGIN_NEO4J_TRACE_CYPHER_PARAMETERS:f
 plugin.neo4j.cypher_parameters_max_length=${SW_PLUGIN_NEO4J_CYPHER_PARAMETERS_MAX_LENGTH:512}
 # If set to positive number, the `db.statement` would be truncated to this length, otherwise it would be completely saved, which may cause performance problem.
 plugin.neo4j.cypher_body_max_length=${SW_PLUGIN_NEO4J_CYPHER_BODY_MAX_LENGTH:2048}
+# If set to positive number and activate `trace sampler cpu policy plugin`, the trace will not be sampled when agent process cpu usage percent is greater than `plugin.cpupolicy.sample_cpu_usage_percent_limit`.
```
wu-sheng, line 273:
```suggestion
# If set to a positive number and activate `trace sampler CPU policy plugin`, the trace would not be collected when agent process CPU usage percent is greater than `plugin.cpupolicy.sample_cpu_usage_percent_limit`.
```

docs/en/setup/service-agent/java-agent/Optional-plugins.md
```diff
@@ -20,4 +20,5 @@ Now, we have the following known optional plugins.
 * Plugin of guava-cache in the optional plugin folder. The reason for being an optional plugin is, this plugin enhanced cache framework, generates large number of local spans, which have a potential performance impact.
 * Plugin of fastjson serialization lib in optional plugin folder.
 * Plugin of jackson serialization lib in optional plugin folder.
-* Plugin of Apache ShenYu(incubating) Gateway 2.4.x in optional plugin folder. Please only activate this plugin when you install agent in Apache ShenYu Gateway. 
\ No newline at end of file
+* Plugin of Apache ShenYu(incubating) Gateway 2.4.x in optional plugin folder. Please only activate this plugin when you install agent in Apache ShenYu Gateway.
+* Plugin of trace sampler cpu policy in the optional plugin folder. Please only activate this plugin when you need limit agent sample when the agent process cpu usage is too high.
```
wu-sheng, line 24:
```suggestion
* Plugin of trace sampler CPU policy in the optional plugin folder. Please only activate this plugin when you need to disable trace collecting when the agent process CPU usage is too high(over threshold).
```

docs/en/setup/service-agent/java-agent/configurations.md
```diff
@@ -110,6 +110,7 @@ This is the properties list supported in `agent/config/agent.config`.
...
+| `plugin.cpupolicy.sample_cpu_usage_percent_limit`               | If set to positive number and activate `trace sampler cpu policy plugin`, the trace will not be sampled when agent process cpu usage percent is greater than `plugin.cpupolicy.sample_cpu_usage_percent_limit`.                                                                                                                                                                                                                                                                                                                                        | SW_SAMPLE_CPU_USAGE_PERCENT_LIMIT                                | `-1`                                                                    |
```
wu-sheng, line 113: Please update this according to above comment.

CHANGES.md
```diff
@@ -10,6 +10,7 @@ Release Notes.
 * Remove redundant `shade.package` property.
 * Add servicecomb-2.x plugin and Testcase.
 * Fix NPE in gateway plugin when the timer triggers webflux webclient call.
+* Add optional plugin trace-sampler-cpu-policy-plugin.
```
```suggestion
* Add an optional plugin, trace-sampler-cpu-policy-plugin, which could disable trace collecting in high CPU load.
```

wu-sheng (approved):

PR #136, opened 2022-04-04
optimize UT

Oliverwqcwrw:


apm-commons/apm-datacarrier/src/test/java/org/apache/skywalking/apm/commons/datacarrier/common/AtomicRangeIntegerTest.java
```diff
@@ -49,19 +49,22 @@ public void testGetAndIncrement() {
     @Test
     @Benchmark
     public void testGetAndIncrementV1Performance() {
-        ATOMIC_V1.getAndIncrement();
+        final int result = ATOMIC_V1.getAndIncrement();
+        Assert.assertTrue(result >= 0 && result <= 100);
```
wu-sheng, line 53: Why do we need `Assert`? This is benchmark testing.

wu-sheng (changes requested):

Oliverwqcwrw:
OK

PR #138, opened 2022-04-04
[Chore] Fix comment's grammar and modifier's order

Oliverwqcwrw:

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
@@ -169,7 +169,7 @@ public class ContextManager implements BootService {
     }
 
     /**
-     * If not sure has the active span, use this method, will be causing NPE when has no active span, use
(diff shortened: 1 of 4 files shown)


wu-sheng (dismissed):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
```diff
@@ -169,7 +169,7 @@ public static AbstractTracerContext awaitFinishAsync(AbstractSpan span) {
     }
 
     /**
-     * If not sure has the active span, use this method, will be cause NPE when has no active span, use
+     * If not sure has the active span, use this method, will be causing NPE when has no active span, use
```
lujiajing1126, line 172: But the grammar is still wrong? 

```suggestion
     * Using this method will cause NPE if active span does not exist. If one is not sure whether there is an active span, use
```

Oliverwqcwrw: My initial thought was that "cause" should be an adjective, but I didn't think there was anything wrong with the order

wu-sheng (approved):

PR #140, opened 2022-04-06
Fix Grpc 1.x plugin could leak context due to gRPC cancelled

Cool-Coding:

When the onCancel method is called,The `contextSnapshot` and `asyncSpan` fields can be all null in the `org.apache.skywalking.apm.plugin.grpc.v1.server.TracingServerCallListener`.  this results in the span operation chain is not completed.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/8820

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,7 +35,7 @@ Release Notes.
 * [Breaking Change] Remove the namespace from `cross process propagation` key.
 * Make sure the parent endpoint in tracing context from existing first ENTRY span, rather than first span only.
 * Fix the bug that maybe causing memory leak and repeated traceId when use gateway-2.1.x-plugin or gateway-3.x-plugin.
-* Fix Grpc 1.x plugin's tracing problem
+* Fix Grpc 1.x plugin could leak context due to gRPC cancelled.
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,7 @@ Release Notes.
 * [Breaking Change] Remove the namespace from `cross process propagation` key.
 * Make sure the parent endpoint in tracing context from existing first ENTRY span, rather than first span only.
 * Fix the bug that maybe causing memory leak and repeated traceId when use gateway-2.1.x-plugin or gateway-3.x-plugin.
+* Fix Grpc 1.x plugin's tracing problem
 
 #### Documentation
 
--- a/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
+++ b/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
@@ -73,6 +73,9 @@ public class TracingServerCallListener<REQUEST> extends ForwardingServerCallList
 
     @Override
     public void onCancel() {
+        if (contextSnapshot == null) {
+            return;
+        }
         final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_CANCEL_OPERATION_NAME);
         span.setComponent(ComponentsDefine.GRPC);
         span.setLayer(SpanLayer.RPC_FRAMEWORK);


wu-sheng:
So, as it is null, it would cause NPE in the runtime?

wu-sheng:
> The contextSnapshot and asyncSpan are all null at some situation.

In what cases this could be NULL? This kind of null check may avoid the issue, but we need to know why. Is this interceptor being called in some unexpected cases or any reason.

Cool-Coding:
> > The contextSnapshot and asyncSpan are all null at some situation.
> 
> In what cases this could be NULL? This kind of null check may avoid the issue, but we need to know why. Is this interceptor being called in some unexpected cases or any reason.

@wu-sheng The value of contextSnapshot and asyncSpan is set in onReady method.

`org.apache.skywalking.apm.plugin.grpc.v1.server.TracingServerCallListener`
```
    @Override
    public void onReady() {
        final AbstractSpan span = ContextManager.createEntrySpan(operation, contextCarrier);
        span.setComponent(ComponentsDefine.GRPC);
        span.setLayer(SpanLayer.RPC_FRAMEWORK);
        contextSnapshot = ContextManager.capture();
        asyncSpan = span.prepareForAsync();
        ContextManager.stopSpan(asyncSpan);

        super.onReady();
    }
```

this onReady method is called by the `io.grpc.internal.ServerCallImpl.onReady` method
```
    @Override
    public void onReady() {
      if (call.cancelled) {
        return;
      }
      listener.onReady();
    }

    public void closed(Status status) {
      try {
        if (status.isOk()) {
          listener.onComplete();
        } else {
          call.cancelled = true;
          listener.onCancel();
        }
      } finally {
        // Cancel context after delivering RPC closure notification to allow the application to
        // clean up and update any state based on whether onComplete or onCancel was called.
        context.cancel(null);
      }
    }
```

In debug mode, I found the value of `call.cancelled` variable is true. so the onReady is directly returned.  when the close method is invoked, Both of the value of contextSnapshot and asyncSpan is null.

Cool-Coding:
> So, as it is null, it would cause NPE in the runtime?

if the `contextSnapshot ` is null, `ContextManager.continued(contextSnapshot)` will throw `IllegalArgumentException` exception.  this results in the `onCancel` span is not stopped.  I explained it detailed in [issue 8820](https://github.com/apache/skywalking/issues/8820)

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
```diff
@@ -76,15 +76,19 @@ public void onCancel() {
...
         try {
             super.onCancel();
         } catch (Throwable t) {
             ContextManager.activeSpan().log(t);
             throw t;
         } finally {
             ContextManager.stopSpan();
-            asyncSpan.asyncFinish();
+            if (asyncSpan != null) {
+                asyncSpan.asyncFinish();
```
wu-sheng, line 90: Once we did `asyncFinish`, I think we should set `asyncSpan = null`

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/server/TracingServerCallListener.java
```diff
@@ -76,15 +76,19 @@ public void onCancel() {
         final AbstractSpan span = ContextManager.createLocalSpan(operationPrefix + REQUEST_ON_CANCEL_OPERATION_NAME);
         span.setComponent(ComponentsDefine.GRPC);
         span.setLayer(SpanLayer.RPC_FRAMEWORK);
-        ContextManager.continued(contextSnapshot);
+        if (contextSnapshot != null) {
```
wu-sheng, line 79: If no contextSnapshot to continue, are we still reasonable to create a local span? I think this would be a no-parent span.

Cool-Coding: @wu-sheng I just have verified it. you are right. if the `onCancel` creates local span, it is just alone.
image

image

this trace data is useless. if the `contextSnapshot` is null, the `onCancel` method should directly return.  as for `asyncSpan`, it don't need to be added not-null condition. because if the `contextSnapshot` is not null, it proved that the `onReady` method is executed. Then the  value  of `asyncSpan` must not be null.

CHANGES.md
```diff
@@ -35,6 +35,7 @@ Release Notes.
 * [Breaking Change] Remove the namespace from `cross process propagation` key.
 * Make sure the parent endpoint in tracing context from existing first ENTRY span, rather than first span only.
 * Fix the bug that maybe causing memory leak and repeated traceId when use gateway-2.1.x-plugin or gateway-3.x-plugin.
+* Fix Grpc 1.x plugin's tracing problem
```
wu-sheng, line 38:
```suggestion
* Fix Grpc 1.x plugin could leak context due to gRPC cancelled.
```

wu-sheng (approved):

CHANGES.md
Cool-Coding: > 

The leaked context causes other grpc request's tracing problem. I think it may be more reasonable belonging to tracing problem. WDYT?

PR #146, opened 2022-04-07
[Feature] Add JDK ThreadPoolExecutor plugin(#8743)

Cool-Coding:
### Add an agent plugin to support JDK ThreadPoolExecutor and its subclasses

### <Feature description>

*ThreadPoolExecutor.submit(Callable)*
image

*ThreadPoolExecutor.execute(Runnable)*
image

*ThreadPoolExecutor.submit(Runnable)*
image

*ScheduledThreadPoolExecutor.submit(Callable)*
image

*ScheduledThreadPoolExecutor.execute(Runnable)*
image

*ScheduledThreadPoolExecutor.submit(Runnable)*
image

*submit(Callable) exception*
image

*execute(Runnable) exception*
image

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/8743.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
@@ -50,6 +50,7 @@ public class SnifferConfigInitializer {
     private static final String ENV_KEY_PREFIX = "skywalking.";
     private static Properties AGENT_SETTINGS;
     private static boolean IS_INIT_COMPLETED = false;
+    private static boolean IS_PLUGIN_INIT_COMPLETED = false;
 
     /**
      * If the specified agent config path is set, the agent will try to locate the specified agent config. If the
@@ -233,5 +234,13 @@ public class SnifferConfigInitializer {
             default:
                 LogManager.setLogResolver(new PatternLogResolver());
         }
+    }
+
+    public static void pluginInitCompleted() {
+        IS_PLUGIN_INIT_COMPLETED = true;
+    }
+
+    public static boolean isPluginInitCompleted() {
+        return IS_PLUGIN_INIT_COMPLETED;
     }
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
@@ -50,7 +50,6 @@ public class SnifferConfigInitializer {
     private static final String ENV_KEY_PREFIX = "skywalking.";
     private static Properties AGENT_SETTINGS;
     private static boolean IS_INIT_COMPLETED = false;
-    private static boolean IS_PLUGIN_INIT_COMPLETED = false;
 
     /**
      * If the specified agent config path is set, the agent will try to locate the specified agent config. If the
@@ -234,13 +233,5 @@ public class SnifferConfigInitializer {
             default:
                 LogManager.setLogResolver(new PatternLogResolver());
         }
-    }
-
-    public static void pluginInitCompleted() {
-        IS_PLUGIN_INIT_COMPLETED = true;
-    }
-
-    public static boolean isPluginInitCompleted() {
-        return IS_PLUGIN_INIT_COMPLETED;
     }
 }
--- a/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
+++ b/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
@@ -130,7 +130,7 @@ public class SkyWalkingAgent {
                     .with(new Listener())
                     .installOn(instrumentation);
 
-        SnifferConfigInitializer.pluginInitCompleted();
+        PluginFinder.pluginInitCompleted();
 
         try {
             ServiceManager.INSTANCE.boot();
--- a/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
+++ b/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
@@ -130,6 +130,8 @@ public class SkyWalkingAgent {
                     .with(new Listener())
                     .installOn(instrumentation);
 
+        SnifferConfigInitializer.pluginInitCompleted();
+
         try {
             ServiceManager.INSTANCE.boot();
         } catch (Exception e) {
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin;
+
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import java.lang.reflect.Method;
+
+public abstract class AbstractThreadingPoolInterceptor implements InstanceMethodsAroundInterceptor {
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+        if (!ContextManager.isActive()) {
+            return;
+        }
+
+        if (allArguments == null || allArguments.length < 1) {
+            return;
+        }
+
+        Object argument = allArguments[0];
+
+        Object updateResult = wrap(argument);
+        if (updateResult != null) {
+            allArguments[0] = updateResult;
+        }
+    }
+
+    public abstract Object wrap(Object param);
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        return ret;
+    }
+
+    @Override
+    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Throwable t) {
+        if (ContextManager.isActive()) {
+            ContextManager.activeSpan().log(t);
+        }
+    }
+}
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
@@ -57,8 +57,6 @@ public abstract class AbstractThreadingPoolInterceptor implements InstanceMethod
 
     @Override
     public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Throwable t) {
-        if (ContextManager.isActive()) {
-            ContextManager.activeSpan().log(t);
-        }
+        ContextManager.activeSpan().log(t);
     }
 }
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
@@ -43,6 +43,11 @@ public abstract class AbstractThreadingPoolInterceptor implements InstanceMethod
         }
     }
 
+    /**
+     * wrap the Callable or Runnable object if needed
+     * @param param  Callable or Runnable object
+     * @return Wrapped object or null if not needed to wrap
+     */
     public abstract Object wrap(Object param);
 
     @Override
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
@@ -37,9 +37,9 @@ public abstract class AbstractThreadingPoolInterceptor implements InstanceMethod
 
         Object argument = allArguments[0];
 
-        Object updateResult = wrap(argument);
-        if (updateResult != null) {
-            allArguments[0] = updateResult;
+        Object wrappedObject = wrap(argument);
+        if (wrappedObject != null) {
+            allArguments[0] = wrappedObject;
         }
     }
 
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolExecuteMethodInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolExecuteMethodInterceptor.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin;
+
+import org.apache.skywalking.apm.plugin.wrapper.SwRunnableWrapper;
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import java.util.concurrent.RunnableFuture;
+
+public class ThreadPoolExecuteMethodInterceptor extends AbstractThreadingPoolInterceptor {
+
+    @Override
+    public Object wrap(Object param) {
+        if (param instanceof SwRunnableWrapper) {
+            return param;
+        }
+
+        if (param instanceof RunnableFuture) {
+            return null;
+        }
+
+        if (!(param instanceof Runnable)) {
+            return null;
+        }
+
+        Runnable runnable = (Runnable) param;
+        return new SwRunnableWrapper(runnable, ContextManager.capture());
+    }
+
+}
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolExecuteMethodInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolExecuteMethodInterceptor.java
@@ -27,7 +27,7 @@ public class ThreadPoolExecuteMethodInterceptor extends AbstractThreadingPoolInt
     @Override
     public Object wrap(Object param) {
         if (param instanceof SwRunnableWrapper) {
-            return param;
+            return null;
         }
 
         if (param instanceof RunnableFuture) {
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/define/ThreadPoolExecutorInstrumentation.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/define/ThreadPoolExecutorInstrumentation.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.define;
+
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import net.bytebuddy.matcher.ElementMatchers;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.ConstructorInterceptPoint;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.InstanceMethodsInterceptPoint;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.StaticMethodsInterceptPoint;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassEnhancePluginDefine;
+import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.HierarchyMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.MultiClassNameMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.logical.LogicalMatchOperation;
+
+public class ThreadPoolExecutorInstrumentation extends ClassEnhancePluginDefine {
+
+    protected static final String ENHANCE_CLASS = "java.util.concurrent.ThreadPoolExecutor";
+
+    private static final String INTERCEPT_EXECUTE_METHOD = "execute";
+
+    private static final String INTERCEPT_SUBMIT_METHOD = "submit";
+
+    private static final String INTERCEPT_EXECUTE_METHOD_HANDLE = "org.apache.skywalking.apm.plugin.ThreadPoolExecuteMethodInterceptor";
+
+    private static final String INTERCEPT_SUBMIT_METHOD_HANDLE = "org.apache.skywalking.apm.plugin.ThreadPoolSubmitMethodInterceptor";
+
+    @Override
+    public boolean isBootstrapInstrumentation() {
+        return true;
+    }
+
+    @Override
+    protected ClassMatch enhanceClass() {
+        return LogicalMatchOperation.or(HierarchyMatch.byHierarchyMatch(ENHANCE_CLASS), MultiClassNameMatch.byMultiClassMatch(ENHANCE_CLASS));
+    }
+
+    @Override
+    public ConstructorInterceptPoint[] getConstructorsInterceptPoints() {
+        return new ConstructorInterceptPoint[0];
+    }
+
+    @Override
+    public InstanceMethodsInterceptPoint[] getInstanceMethodsInterceptPoints() {
+        return new InstanceMethodsInterceptPoint[]{
+                new InstanceMethodsInterceptPoint() {
+                    @Override
+                    public ElementMatcher<MethodDescription> getMethodsMatcher() {
+                        return ElementMatchers.named(INTERCEPT_EXECUTE_METHOD);
+                    }
+
+                    @Override
+                    public String getMethodsInterceptor() {
+                        return INTERCEPT_EXECUTE_METHOD_HANDLE;
+                    }
+
+                    @Override
+                    public boolean isOverrideArgs() {
+                        return true;
+                    }
+                },
+                new InstanceMethodsInterceptPoint() {
+                    @Override
+                    public ElementMatcher<MethodDescription> getMethodsMatcher() {
+                        return ElementMatchers.named(INTERCEPT_SUBMIT_METHOD);
+                    }
+
+                    @Override
+                    public String getMethodsInterceptor() {
+                        return INTERCEPT_SUBMIT_METHOD_HANDLE;
+                    }
+
+                    @Override
+                    public boolean isOverrideArgs() {
+                        return true;
+                    }
+                }
+        };
+    }
+
+    @Override
+    public StaticMethodsInterceptPoint[] getStaticMethodsInterceptPoints() {
+        return new StaticMethodsInterceptPoint[0];
+    }
+
+}
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/define/ThreadPoolExecutorInstrumentation.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/define/ThreadPoolExecutorInstrumentation.java
@@ -23,8 +23,7 @@ import net.bytebuddy.matcher.ElementMatcher;
 import net.bytebuddy.matcher.ElementMatchers;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.ConstructorInterceptPoint;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.InstanceMethodsInterceptPoint;
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.StaticMethodsInterceptPoint;
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassEnhancePluginDefine;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInstanceMethodsEnhancePluginDefine;
 import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
 import org.apache.skywalking.apm.agent.core.plugin.match.HierarchyMatch;
 import org.apache.skywalking.apm.agent.core.plugin.match.MultiClassNameMatch;
@@ -94,6 +93,4 @@ public class ThreadPoolExecutorInstrumentation extends ClassInstanceMethodsEnhan
                 }
         };
     }
-
-
 }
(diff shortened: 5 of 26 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
```diff
@@ -234,4 +235,12 @@ static void configureLogger() {
...
         }
     }
+
+    public static void pluginInitCompleted() {
+        IS_PLUGIN_INIT_COMPLETED = true;
+    }
+
+    public static boolean isPluginInitCompleted() {
+        return IS_PLUGIN_INIT_COMPLETED;
+    }
```
wu-sheng, line 245: What are these about?

test/plugin/scenarios/jdk-threadpool-scenario/bin/startup.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+home="$(cd "$(dirname $0)"; pwd)"
+
+java -jar ${agent_opts} ${home}/../libs/jdk-threadpool-scenario.jar &
```
wu-sheng, line 21: Your new test is not added into GHA.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
Cool-Coding: the `FileWriter` class in Skywalking agent uses `ThreadPoolExecutor`. If plugins are not loaded, FileWriter is created. Then `ThreadPoolExecutor won't be instructed`. so we must make sure that after plugins are loaded, FileWriter can be created.

```
    private FileWriter() {
        logBuffer = new ArrayBlockingQueue(1024);
        final ArrayList<String> outputLogs = new ArrayList<String>(200);
        Executors.newSingleThreadScheduledExecutor(new DefaultNamedThreadFactory("LogFileWriter"))
                 .scheduleAtFixedRate(new RunnableWithExceptionProtection(new Runnable() {
                     @Override
                     public void run() {
                         try {
                             logBuffer.drainTo(outputLogs);
                             for (String log : outputLogs) {
                                 writeToFile(log + Constants.LINE_SEPARATOR);
                             }
                             try {
                                 fileOutputStream.flush();
                             } catch (IOException e) {
                                 e.printStackTrace();
                             }
                         } finally {
                             outputLogs.clear();
                         }
                     }
                 }, new RunnableWithExceptionProtection.CallbackWhenException() {
                     @Override
                     public void handle(Throwable t) {
                     }
                 }), 0, 1, TimeUnit.SECONDS);
    }
```

test/plugin/scenarios/jdk-threadpool-scenario/bin/startup.sh
Cool-Coding: > GHA

I don't ever develop a plugin. How to add it into GHA? Can you teach me? Thanks!

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
wu-sheng: But simply change like this, you would make a lot of logs disappear from `agent.log`

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -138,6 +138,8 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
 
         Runtime.getRuntime()
                 .addShutdownHook(new Thread(ServiceManager.INSTANCE::shutdown, "skywalking service shutdown thread"));
+
+        SnifferConfigInitializer.pluginInitCompleted();
```
wu-sheng, line 142: Also, `Executors.newSingleThreadScheduledExecutor` is not just used in **FileWriter**, others like `JVMService` service used them too. I think put `SnifferConfigInitializer.pluginInitCompleted();` here would make more logs disappeared. You just need this at L131, right after `installOn(instrumentation)`

test/plugin/scenarios/jdk-threadpool-scenario/bin/startup.sh
wu-sheng: `plugin-tests-*.yaml` in https://github.com/apache/skywalking-java/tree/main/.github/workflows. Last step in the doc, https://github.com/apache/skywalking-java/blob/main/docs/en/setup/service-agent/java-agent/Plugin-test.md#local-test-and-pull-request-to-the-upstream

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
Cool-Coding: Yeah, Before plugins are loaded completed, logs will be printed by system.out.

wu-sheng: Also, if you need a new flag of plugin initialization, it looks like the flag should be in `PluginFinder`? [SnifferConfigInitializer.java](https://github.com/apache/skywalking-java/pull/146/files/22e85142ff56493cee90adf49d123b71ff3e0089#diff-3f4c9264a89197350b7eb2ca0e2f2daa494306980a00142edef7f5abfc941951) seems a wrong place to add.

test/plugin/scenarios/jdk-threadpool-scenario/bin/startup.sh
Cool-Coding: > `plugin-tests-*.yaml` in https://github.com/apache/skywalking-java/tree/main/.github/workflows. Last step in the doc, https://github.com/apache/skywalking-java/blob/main/docs/en/setup/service-agent/java-agent/Plugin-test.md#local-test-and-pull-request-to-the-upstream

Ok, I will see it. Thank you!

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+        Object argument = allArguments[0];
+
+        Object updateResult = updateOnlyParam(argument);
+        if (updateResult != null) {
+            allArguments[0] = updateResult;
+        }
+    }
+
+    public abstract Object updateOnlyParam(Object param);
```
wu-sheng, line 46: What does this method name mean? `onlyParam` means?

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+    private ContextSnapshot contextSnapshot;
+
+    public SwRunnableWrapper(Runnable runnable, ContextSnapshot contextSnapshot) {
+        this.runnable = runnable;
+        this.contextSnapshot = contextSnapshot;
+    }
+
+    @Override
+    public void run() {
+        AbstractSpan span = ContextManager.createLocalSpan(getOperationName());
```
wu-sheng, line 39: Are you considering `spin-tracing`? The agent is going to trace its own internal methods? And meanwhile, this method could print logs. You will face endless executions.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
Cool-Coding: > 

it will change the param of `execute` and `submit` method wrapping it to `SWCallableWrapper` or `SWRunnableWrapper`.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
Cool-Coding: > 

I  have considered this case. the instructed point are  just`execute` and `submit` methods  in `ThreadPoolExecutor` and its subclasses.  `FileWriter` and `JVMService` use `scheduleAtFixedRate` method.   

If the `execute` method in `ThreadPoolExecutor` subclass calls its super class `execute` method, this would result in `Runnable` will be wrapped twice, but it won't form a `spin-tracing`.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
Cool-Coding: > Also, `Executors.newSingleThreadScheduledExecutor` is not just used in **FileWriter**, others like `JVMService` service used them too. I think put `SnifferConfigInitializer.pluginInitCompleted();` here would make more logs disappeared. You just need this at L131, right after `installOn(instrumentation)`

Agree. I will try and verify it.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
wu-sheng: But `scheduleAtFixedRate` is also could be used by users, which would be missed by this plugin.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
wu-sheng: Then we should name it `wrapExecutor`?

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
wu-sheng: I think the better way is to identify the caller and exclude the agent package. WDYT?
How to identify could be discussed later.

Cool-Coding: > 

To be honest, My focus is `ThreadPoolExecutor`, not all kinds of 'TheadPool`s

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
Cool-Coding: > 

it doesn't wrap executor. it just wraps the `Runnable` or `Callable`.  we could name it `wrapCommand` or `wrapTask`. WDYT?

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
Cool-Coding: > I think the better way is to identify the caller and exclude the agent package. WDYT? How to identify could be discussed later.

Ok, I will think another better way. Keep in touch!

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
wu-sheng: Just `wrap` should be enough, I think.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
wu-sheng: 2 general ways.
1. Ask SkyWalking core to use a specific interface(extends Runnable/Callable) only. Then the plugin is easy to resolve the caller.
2. Get the stack and try to find `org.apache.skywalking`. 

The (2) is easier but it costs performance due to getting stack elements.

Cool-Coding: > 

Yes, `ScheduledThreadPool` is inherited from `ThreadPoolExecutor`, also has `execute` and `submit` methods. so I take all the subclasses of `ThreadPoolExecutor` into account, but only `execute` and `submit` methods. I'll reconsider it all

Cool-Coding: > Yes, `ScheduledThreadPool` is inherited from `ThreadPoolExecutor`, also has `execute` and `submit` methods. so I take all the subclasses of `ThreadPoolExecutor` into account, but only `execute` and `submit` methods. I'll reconsider it all

Yes, ScheduledThreadPool is inherited from ThreadPoolExecutor, also has execute and submit methods. so I take all the subclasses of ThreadPoolExecutor into account, but only execute and submit methods. Do you think we just take `scheduleAtFixedRate` and `scheduleWithFixedDelay` into account in addition to `execute` and `submit`, or cover all kinds of threadpools?

wu-sheng: Once this is `ThreadPoolExecutor` plugin, it should cover all known cases, which is why we should detect caller one way or another.

wu-sheng: And, have you known there are meter plugins for thread pool, such as `UndertowWorkerThreadPoolConstructorIntercept`. I think if this plugin covers the thread pool case too, it should report the thread pool metrics too.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
Cool-Coding: > Just `wrap` should be enough, I think.

ok

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
wu-sheng: image

This is available on v9 `General Service/Instance/JVM` panel.

Cool-Coding: > And, have you known there are meter plugins for thread pool, such as `UndertowWorkerThreadPoolConstructorIntercept`. I think if this plugin covers the thread pool case too, it should report the thread pool metrics too.

I've seen the issue before. I seen it again just now. it also instructs `ThreadPoolExecutor` with  the match rule that package name is started with "org.xnio". If my `ThreadPoolExecutor` plugin covers  this, the `Undertow ThreadPool` plugin is redundant I think. I will take it into account. 

Please take a look at [the reply](#discussion_r845092153) and give me a reply. Thank you!

wu-sheng: > My focus is ThreadPoolExecutor, not all kinds of TheadPools.

This makes perfect sense to me. We don't need to cover all threads.

> Do you think we just take scheduleAtFixedRate and scheduleWithFixedDelay into account in addition to execute and submit, or cover all kinds of threadpools?

Not all thread pool, just these two popular scheduling ways.

Cool-Coding: @wu-sheng  `ScheduledThreadPoolExecutor.scheduleAtFixedRate` has been supported. but  the trace is getting longer and longer over the time. I think it is a bad sense. WDYT?
image

Besides, `ScheduledThreadPoolExecutor.scheduleAtFixedRate` in `JVMService` doesn't show in the trace ui. I don't identify the caller.  I don't know the reason.

wu-sheng: OK, I think we should not support this for now. It makes the trace strange.

Cool-Coding: > OK, I think we should not support this for now. It makes the trace strange.

Agree.

Cool-Coding: > > And, have you known there are meter plugins for thread pool, such as `UndertowWorkerThreadPoolConstructorIntercept`. I think if this plugin covers the thread pool case too, it should report the thread pool metrics too.
> 
> I've seen the issue before. I seen it again just now. it also instructs `ThreadPoolExecutor` with the match rule that package name is started with "org.xnio". If my `ThreadPoolExecutor` plugin covers this, the `Undertow ThreadPool` plugin is redundant I think. I will take it into account.
> 
> Please take a look at [reply1](#discussion_r845039466), [reply2 ](#discussion_r845092153). I want to know your point of view. Thank you!

@wu-sheng  I am adding `ThreadPoolExecutor` metrics. But How to distinguish different thread pools？That is, how to determine the thread pool name. so far, I haven't thought of a meaningful name yet. Do you have any good suggestions?

wu-sheng: Thread pool supports setting a custom name by users. AFAIK.

wu-sheng: If it is not set, I think we could use pool's object ID as the name? It is meaningless, yes, but it is better than nothing.
After all, the thread pool metric is at instance level only.

wu-sheng:
Are you going to create a new one?

Cool-Coding:
> Are you going to create a new one?

No, I will still use this one. There is  a new PR  before my PR is merged.  I reset my commit just now and pulled the new PR just now. `GitHub` closed my PR automatically.  I have reopened it.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
Cool-Coding: > Thread pool supports setting a custom name by users. AFAIK.

User can custom thread's name, not threadpool's name. They are not equivalent. I will use object ID or other name that seems more meaningful.

Cool-Coding: @wu-sheng Usually, an application includes many thread pools. If we collect all thread pool meters, but users don't pay attention to these indicators, then these collected meters are garbage data. I think we can let users configure thread pools they care about in `agent.config`. the rough format is  `class1:name1,name2;class2:name1,name2`. Then we can match it  and name `ThreadFactory` in `ThreadPoolExecutor` plugin through `thread stack`. Although it consumes a little performance when the application starts, I think it's worth it. WDYT?

wu-sheng:
It seems some test cases fail, please recheck.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
wu-sheng: @Cool-Coding I think this part of codes has not been polished yet.

Cool-Coding:
> It seems some test cases fail, please recheck.

I have rechecked it. It seems that it has nothing to do with my modification. I will recheck it again.  please take a look at [my reply](#discussion_r847191002).

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
wu-sheng: Got it, make sense. Yes, maybe we have to use the class name as the thread pool name. 

I think we could make trace supported in this PR and do meter collecting in another?

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/define/HierarchyThreadPoolExecutorInstrumentation.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.define;
+
+import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.HierarchyMatch;
+
+public class HierarchyThreadPoolExecutorInstrumentation extends AbstractThreadPoolExecutorInstrumentation {
```
wu-sheng, line 24: Consider to use `LogicalOrMatch` to replace this and `ThreadPoolExecutorInstrumentation`?

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
wu-sheng: One more scenario to discussion, 

The ThreadPoolExecutor could be used to run an endless loop task, like a timer. And `run` method would never end. 
As you mentioned using the class name to filter in meter collection, should we add this kind of mechanism in the tracing part too? (Have to back about caller detection, as user could use APIs rather than inherit from this)

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/define/HierarchyThreadPoolExecutorInstrumentation.java
Cool-Coding: > Consider to use `LogicalOrMatch` to replace this and `ThreadPoolExecutorInstrumentation`?

I have tried to use `LogicalMatchOperation.or` before. but it accepts `IndirectMatch` class type, however, `NameMatch` is not subclass of `IndirectMatch`. I  replaced `NameMatch` with `MultiClassNameMatch` and verified it just now ,it works.

`LogicalMatchOperation.or(HierarchyMatch.byHierarchyMatch(ENHANCE_CLASS), MultiClassNameMatch.byMultiClassMatch(ENHANCE_CLASS))`

wu-sheng: Make sense.

wu-sheng:
Many NPEs happen in the UT, I think you break something. Please recheck.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
Cool-Coding: > @Cool-Coding I think this part of codes has not been polished yet.

I have polished the `WriterFactoryTest` UT adding the code `        BDDMockito.given(SnifferConfigInitializer.isPluginInitCompleted()).willReturn(true)`. The UT passed.

And the statement `SnifferConfigInitializer.pluginInitCompleted()` has been moved after `installOn(instrumentation)`.

```
        agentBuilder.type(pluginFinder.buildMatch())
                    .transform(new Transformer(pluginFinder))
                    .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
                    .with(new RedefinitionListener())
                    .with(new Listener())
                    .installOn(instrumentation);

        SnifferConfigInitializer.pluginInitCompleted();

        try {
            ServiceManager.INSTANCE.boot();
        } catch (Exception e) {
            LOGGER.error(e, "Skywalking agent boot failure.");
        }
```

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
Cool-Coding: > Got it, make sense. Yes, maybe we have to use the class name as the thread pool name.
> 
> I think we could make trace supported in this PR and do meter collecting in another?

ok, agree!

Cool-Coding: > One more scenario to discussion,
> 
> The ThreadPoolExecutor could be used to run an endless loop task, like a timer. And `run` method would never end. As you mentioned using the class name to filter in meter collection, should we add this kind of mechanism in the tracing part too? (Have to back about caller detection, as user could use APIs rather than inherit from this)

Do you mean we do not trace some `Runnable` or `Callable` task ?

Cool-Coding:
> Many NPEs happen in the UT, I think you break something. Please recheck.

I have rechecked it again. There is still one failed that has nothing to do with my modification.
```
ERROR: 1 dead links found!
[✖] https://twitter.com/ASFSkyWalking → Status: 404
```

wu-sheng:
> > Many NPEs happen in the UT, I think you break something. Please recheck.
> 
> I have rechecked it again. There is still one failed that has nothing to do with my modification.
> 
> ```
> ERROR: 1 dead links found!
> [✖] https://twitter.com/ASFSkyWalking → Status: 404
> ```

Ignore this. Don't worry.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
wu-sheng: > Also, if you need a new flag of plugin initialization, it looks like the flag should be in PluginFinder? SnifferConfigInitializer.java seems a wrong place to add.

I mean which class should own the flag. `SnifferConfigInitializer` has nothing related to the plugin mechanism.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolExecuteMethodInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+public class ThreadPoolExecuteMethodInterceptor extends AbstractThreadingPoolInterceptor {
+
+    @Override
+    public Object wrap(Object param) {
+        if (param instanceof SwRunnableWrapper) {
+            return param;
+        }
+
+        if (param instanceof RunnableFuture) {
+            return null;
```
wu-sheng, line 34: Would not `return null` cause an NPE? Same as below.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolExecuteMethodInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        if (!(param instanceof Runnable)) {
+            return null;
```
wu-sheng, line 38: Same question about `return null`.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/define/ThreadPoolExecutorInstrumentation.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.ConstructorInterceptPoint;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.InstanceMethodsInterceptPoint;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.StaticMethodsInterceptPoint;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassEnhancePluginDefine;
+import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.HierarchyMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.MultiClassNameMatch;
+import org.apache.skywalking.apm.agent.core.plugin.match.logical.LogicalMatchOperation;
+
+public class ThreadPoolExecutorInstrumentation extends ClassEnhancePluginDefine {
```
wu-sheng, line 33:
```suggestion
public class ThreadPoolExecutorInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
```

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/define/ThreadPoolExecutorInstrumentation.java
```diff
@@ -0,0 +1,103 @@
...
+                        return true;
+                    }
+                }
+        };
+    }
+
+    @Override
+    public StaticMethodsInterceptPoint[] getStaticMethodsInterceptPoints() {
+        return new StaticMethodsInterceptPoint[0];
+    }
```
wu-sheng, line 101:
```suggestion
```

ClassInstanceMethodsEnhancePluginDefine has covered this.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolExecuteMethodInterceptor.java
Cool-Coding: it won't cause NPE. Let's see the caller code in 'beforeMethod'.
```
        Object updateResult = wrap(argument);
        if (updateResult != null) {
            allArguments[0] = updateResult;
        }
```

Cool-Coding: > OK, please add comments on `AbstractThreadingPoolInterceptor#wrap`.
> 
> Also consider to add more comments on codes too.
Ok!

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/wrapper/SwRunnableWrapper.java
wu-sheng: No need, please address others. This is good for me now.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
```diff
@@ -0,0 +1,59 @@
...
+    public abstract Object wrap(Object param);
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        return ret;
+    }
+
+    @Override
+    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Throwable t) {
+        if (ContextManager.isActive()) {
```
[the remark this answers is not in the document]
Cool-Coding: > 

Normally, There must be an context.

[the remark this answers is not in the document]
wu-sheng: Then, please remove all these checks. An exception is better than nothing.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -130,6 +130,8 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
                     .with(new Listener())
                     .installOn(instrumentation);
 
+        SnifferConfigInitializer.pluginInitCompleted();
```
wu-sheng, line 133:
```suggestion
        PluginFinder.pluginInitCompleted();
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/SnifferConfigInitializer.java
wu-sheng: Same. I am not arguing about the process flow, I am talking about where to hold this flag.

I mean moving this flag to https://github.com/apache/skywalking-java/pull/146#discussion_r848061270

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
Cool-Coding: I think `pluginInitCompleted` show belong to `SnifferConfigInitializer`.

the IS_PLUGIN_INIT_COMPLETED field is not a flag that indicates the plugins are loaded. It's the flag that ByteBuddy has loaded all plugins. Before this, ThreadPoolExecutor can't be created.

wu-sheng: No, you have a misunderstanding with the codes. PluginFinder is the one working with bytebuddy. SnifferConfigInitializer is just about `Config.class`.

Please read the codes again.

apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/AbstractThreadingPoolInterceptor.java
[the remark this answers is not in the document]
Cool-Coding: > Then, please remove all these checks. An exception is better than nothing.

Ok.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
wu-sheng: Putting it in the Agent makes sense.

Cool-Coding: > Member

Ok, Then keep it like this.

wu-sheng: I think you still don't follow all requirements. Rename is required, and move to `SkyWalkingAgent#IS_PLUGIN_INSTALLED_COMPLETED`.



> > Putting it in the Agent makes sense.
> 
> Ok, Then keep it like this.

I think I said putting it in the Agent class, not the way it is right now. You somehow misunderstood the point.

Cool-Coding: > Please move and rename this flag, I think others are good to me.

Ok, I understand it.  `SnifferConfigInitializer` is just responsible for configuration.  I will move it to the `PluginFinder`.

wu-sheng (approved):

PR #154, opened 2022-04-12
Support default database(not set through JDBC URL) in mysql-5.x plugin

xu1009:
### Fix <https://github.com/apache/skywalking/issues/8860>
  

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<8860>.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -37,7 +37,7 @@ Release Notes.
 * Fix the bug that maybe causing memory leak and repeated traceId when use gateway-2.1.x-plugin or gateway-3.x-plugin.
 * Fix Grpc 1.x plugin could leak context due to gRPC cancelled.
 * Add JDK ThreadPoolExecutor Plugin.
-* Fix mysql-5.x plugins NPE.
+* Support default database(not set through JDBC URL) in mysql-5.x plugin.
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,7 @@ Release Notes.
 * Make sure the parent endpoint in tracing context from existing first ENTRY span, rather than first span only.
 * Fix the bug that maybe causing memory leak and repeated traceId when use gateway-2.1.x-plugin or gateway-3.x-plugin.
 * Fix Grpc 1.x plugin could leak context due to gRPC cancelled.
+* Fix mysql-5.x plugins NPE.
 
 #### Documentation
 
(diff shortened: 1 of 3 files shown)


apm-sniffer/apm-sdk-plugin/mysql-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/mysql/v5/ConnectionCreate5xInterceptor.java
```diff
@@ -42,7 +43,8 @@ public void beforeMethod(Class clazz, Method method, Object[] allArguments, Clas
     public Object afterMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes,
         Object ret) {
         if (ret instanceof EnhancedInstance) {
-            ConnectionInfo connectionInfo = ConnectionCache.get(allArguments[0].toString(), allArguments[1].toString(), allArguments[3].toString());
+            String database = Objects.isNull(allArguments[3]) ? "" : allArguments[3].toString();
```
wu-sheng, line 46: URL could be NULL? Because? Users forgot to set?

xu1009: jdbc url like 
jdbc:mysql://ip:port?zeroDateTimeBehavior=convertToNull&useUnicode=true&characterEncoding=UTF8&connectTimeout=10000&socketTimeout=20000&useSSL=false&useTimezone=true&serverTimezone=GMT%2B8

the database not assign in jdbcurl, this can get null databse

CHANGES.md
```diff
@@ -37,6 +37,7 @@ Release Notes.
 * Fix the bug that maybe causing memory leak and repeated traceId when use gateway-2.1.x-plugin or gateway-3.x-plugin.
 * Fix Grpc 1.x plugin could leak context due to gRPC cancelled.
 * Add JDK ThreadPoolExecutor Plugin.
+* Fix mysql-5.x plugins NPE.
```
wu-sheng, line 40:
```suggestion
* Support default database(not set through JDBC URL) in mysql-5.x plugin.
```

wu-sheng (approved):

PR #157, opened 2022-04-15
Possible bug fix + logging grammar improvement

BFergerson:
I'm unsure if this is a legit bug but it looked strange that `Config.Agent.NAMESPACE` is being used for `namespace` and `cluster` when `Config.Agent.CLUSTER` also exists.

If that's not a bug I can remove it. Also improved logging grammar.


wu-sheng (approved):

BFergerson:
@wu-sheng, looks like it was a bug.

Just noticed I forgot to update changes. Should I open another PR?

wu-sheng:
Yes, it should be a bug.

PR #159, opened 2022-04-18
Add layer field to event when reporting

fgksgf:

**Note: There is a circular dependency issue in E2E tests, I will update E2E test cases after [this PR](https://github.com/apache/skywalking/pull/8881)  is merged.**

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Constants.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Constants.java
@@ -26,4 +26,8 @@ public class Constants {
     public static String EMPTY_STRING = "";
 
     public static char SERVICE_NAME_PART_CONNECTOR = '|';
+
+    // The name of the layer that represents agent-installed services,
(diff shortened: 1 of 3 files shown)


kezhenxu94 (dismissed):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Constants.java
```diff
@@ -26,4 +26,8 @@ public class Constants {
     public static String EMPTY_STRING = "";
 
     public static char SERVICE_NAME_PART_CONNECTOR = '|';
+
+    // The name of the layer that represents agent-installed services,
+    // which is defined at https://github.com/apache/skywalking/blob/master/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/analysis/Layer.java
```
wu-sheng, line 31: If  you want to put link in comments, please use perm link.

fgksgf: ok

kezhenxu94 (approved):

wu-sheng (approved):

PR #162, opened 2022-04-20
[Feature] [ShenYu]<Apache ShenYu (incubating) plugin support dubbo proxy>(#8869)

hutaishi:
Add skywaliking plugin for support Apache ShenYu (incubating) gateway.

### topology
image

### trace
image


wu-sheng:
2 questions
1. Which parts are pure Dubbo plugin enhancement? Tbgis part shoud be in another PR
2. Why you need a snapshot continue inside Dubbo plugin?You could use a local span to continue in Shenyu. Binding 2 plugins with a context key is not recommended, only when has to be done, I feel this is not the case.

hutaishi:
> 2 questions
> 
> 1. Which parts are pure Dubbo plugin enhancement? Tbgis part shoud be in another PR
> 2. Why you need a snapshot continue inside Dubbo plugin?You could use a local span to continue in Shenyu. Binding 2 plugins with a context key is not recommended, only when has to be done, I feel this is not the case.

Hello, I'd like to ask the second question. I tried to use localspan, but due to cross threading, abstracttracercontext is empty when closing localspan.

image




image

wu-sheng:
I  not suggesting local span in Dubbo plugin, I suggest you to add a new local span in your own codes and in the same thread of Dubbo filter. Then they are linked automatically.

hutaishi:
> I not suggesting local span in Dubbo plugin, I suggest you to add a new local span in your own codes and in the same thread of Dubbo filter. Then they are linked automatically.


 I understand in an instant. Thank you for your guidance.
successful.
image

I'll change it in your way later.

hutaishi:
> 2 questions
> 
> 1. Which parts are pure Dubbo plugin enhancement? Tbgis part shoud be in another PR
> 2. Why you need a snapshot continue inside Dubbo plugin?You could use a local span to continue in Shenyu. Binding 2 plugins with a context key is not recommended, only when has to be done, I feel this is not the case.

now:
Only the Shenyu plugin is modified and a localspan is created to concat Dubbo

apm-sniffer/optional-plugins/shenyu-2.4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/shenyu/v24x/ApacheDubboPluginInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.lang.reflect.Method;
+
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.plugin.shenyu.v24x.util.CommonUtil;
+
+/**
+ * shenyu AlibabaDubboPlugin Interceptor.
```
wu-sheng, line 29: Dubbo doesn't belong to Alibaba for a long time. Please fix relative comments.

apm-sniffer/optional-plugins/shenyu-2.4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/shenyu/v24x/define/AlibabaDubboPluginInstrumentation.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInstanceMethodsEnhancePluginDefine;
+import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static org.apache.skywalking.apm.agent.core.plugin.match.NameMatch.byName;
+
+public class AlibabaDubboPluginInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
+    @Override
+    protected ClassMatch enhanceClass() {
+        return byName("org.apache.shenyu.plugin.alibaba.dubbo.AlibabaDubboPlugin");
```
wu-sheng, line 34: Why does Shenyu require old Dubbo? As a proxy, this should be transparent, isn't it?

wu-sheng:
Besides Alibaba legacy Dubbo question, the plugin test should be added too.

apm-sniffer/optional-plugins/shenyu-2.4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/shenyu/v24x/define/AlibabaDubboPluginInstrumentation.java
hutaishi: I have tested that Dubbo 2.7x client can access the old version of Dubbo, but Shenyu is based on two plug-ins, not one.
I'll give a suggestion to Shenyu community.

hutaishi:
> Besides Alibaba legacy Dubbo question, the plugin test should be added too.

thinks, I will add test.

apm-sniffer/optional-plugins/shenyu-2.4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/shenyu/v24x/define/AlibabaDubboPluginInstrumentation.java
wu-sheng: Alibaba Dubbo is like 3 years old, I guess? And considering no one is maintaining it, and fewer people are going to use it, I would suggest we remove that support.
Even for SkyWalking's native Alibaba Dubbo plugin, once we received a report saying there is a bug, the official solution would be to remove it too.

hutaishi: > Alibaba Dubbo is like 3 years old, I guess? And considering no one is maintaining it, and fewer people are going to use it, I would suggest we remove that support. Even for SkyWalking's native Alibaba Dubbo plugin, once we received a report saying there is a bug, the official solution would be to remove it too.

Agree. In addition, the new version of Dubbo client can also call the old duboo service, although I haven't fully tested it

I will remove support for the old version.

wu-sheng: > Agree. In addition, the new version of Dubbo client can also call the old duboo service, although I haven't fully tested it

I think the protocol is not unchanged, RPC frameworks are keeping forward compatibility seriously, I assume Dubbo community should consider this already. 
If you or Shenyu community has concerns, you could submit a question to Dubbo to confirm, I think they would be happy to confirm and see 3 projects closing integration.

hutaishi:
> Besides Alibaba legacy Dubbo question, the plugin test should be added too.

Thanks to the mentor's guidance, it has been modified and passed the local test, the UI works fine，Please help review again.

test/plugin/scenarios/shenyu-2.4.x-dubbo-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,134 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+segmentItems:
+  - serviceName: shenyu-gateway
+    segmentSize: nq 0
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: test.apache.skywalking.apm.testcase.shenyu.http.services.OrderService.$invoke(String,String[],Object[])
```
wu-sheng, line 22: Is this `OrderService.$invoke` expected? Is this an auto-generated method?

hutaishi: Yes， skywalkinig dubbo plugin aotu-generated.    Generic service in dubbo.

hutaishi:
> General good, once you confirmed whether the operation names fit your expectation.


This operation name is generated by the skywalking Dubbo plugin, which does not support generalization calls well.

wu-sheng:
> > General good, once you confirmed whether the operation names fit your expectation.
> 
> This operation name is generated by the skywalking Dubbo plugin, which does not support generalization calls well.

A tip, our doc doesn't mention HTTP/gRPC or some other protocols(you supported) are/is supported or not, you may be better to document this on your side. @hutaishi

hutaishi:
> > > General good, once you confirmed whether the operation names fit your expectation.
> > 
> > 
> > This operation name is generated by the skywalking Dubbo plugin, which does not support generalization calls well.
> 
> A tip, our doc doesn't mention HTTP/gRPC or some other protocols(you supported) are/is supported or not, you may be better to document this on your side. @hutaishi

Thank you for your guidance and tip, in fact, I would like to propose an issue to support other rpc and improve the documentation, you see whether it is feasible

wu-sheng:
OK, generally about the details, Shenyu is better than us to host, because you would keep your eyes on the changes with time. Here (in SkyWalking), it will be easy to outdated, eventually, it will harm your end users.

PR #165, opened 2022-04-25
Change the dateformat of logs to yyyy-MM-dd HH:mm:ss.SSS

guoxiaod:
### change dateformat  to yyyy-MM-dd HH:mm:ss.SSS
ISO 8601 is an international standard for date and time formats.

In 4.2.2.4 of the standard, it says,
If necessary for a particular application a decimal fraction of hour, minute or second may be included. If a decimal fraction is included, lower order time elements (if any) shall be omitted and the decimal fraction shall be divided from the integer part by the decimal sign specified in ISO 31-0, i.e. the comma [,] or full stop [.]. Of these, the comma is the preferred sign. If the magnitude of the number is less than unity, the decimal sign shall be preceded by two zeros in accordance with 3.6.

So, we can use ',' or '.' as decimal sign.
Although the standard recommends  ',', but usually we use '.' in java.

https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf


wu-sheng:

wu-sheng:
Thanks for pointing this out. I think we missed this somehow.

wu-sheng (approved):

PR #166, opened 2022-04-27
[Polish] Code level polish (no logic change)

Oliverwqcwrw:


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/CustomizeExpression.java
```diff
@@ -37,7 +37,7 @@ public class CustomizeExpression {
     private static final ILog LOGGER = LogManager.getLogger(CustomizeExpression.class);
 
     public static Map<String, Object> evaluationContext(Object[] allArguments) {
-        Map<String, Object> context = new HashMap<>();
+        Map<String, Object> context = new HashMap<>(allArguments.length, 1);
```
wu-sheng, line 40: Why change the load factor to `1`?

Oliverwqcwrw: Using the default threshold, it will still expand when capacity is `0.75 * size()`, and I want to prevent it from expanding completely

wu-sheng: I think this can't avoid that case actually. Because factor is not determining this.

Oliverwqcwrw: I'm a little confused

>  if (++size > threshold)
            resize();

It is true that it is above the calculated threshold for expansion

If I'm wrong can you explain exactly what it is that determines expansion

lujiajing1126: If you think the load factor is critical, you may provide the report of microbenchmarks.

Oliverwqcwrw: > 
    @Test
    public void expandTest(){
        HashMap<String, String> map = new HashMap<>(20);
        final long start1 = System.nanoTime();
        for (int i = 0; i < 20; i++) {
            map.put("key" + i, "value");
        }
        System.out.println("Expansion takes nanoTime = " + (System.nanoTime() - start1));

        HashMap<String,String> map2 = new HashMap<>(20, 1);
        final long start2 = System.nanoTime();
        for (int i = 0; i < 20; i++) {
            map2.put("key" + i, "value");
        }
        System.out.println("Without expansion takes nanoTime = " + (System.nanoTime() - start2));
    }

> Expansion takes nanoTime = 31689
Without expansion takes nanoTime = 7528

wu-sheng: Is this performance sensitive? Bigger factor actually costs more memory. Generally, this is a tradeoff.

Oliverwqcwrw: I think performance should take precedence over memory consumption for `agent`

Do you think?

lujiajing1126: > > 
> 
> ```
> @Test
> public void expandTest(){
>     HashMap<String, String> map = new HashMap<>(20);
>     final long start1 = System.nanoTime();
>     for (int i = 0; i < 20; i++) {
>         map.put("key" + i, "value");
>     }
>     System.out.println("Expansion takes nanoTime = " + (System.nanoTime() - start1));
> 
>     HashMap<String,String> map2 = new HashMap<>(20, 1);
>     final long start2 = System.nanoTime();
>     for (int i = 0; i < 20; i++) {
>         map2.put("key" + i, "value");
>     }
>     System.out.println("Without expansion takes nanoTime = " + (System.nanoTime() - start2));
> }
> ```
> 
> > Expansion takes nanoTime = 31689
> > Without expansion takes nanoTime = 7528

I deeply doubt this test. This can hardly be called a microbenchmark.

If you swap two cases, the result will be the opposite!

```java
HashMap<String,String> map2 = new HashMap<>(20, 1);
final long start2 = System.nanoTime();
for (int i = 0; i < 20; i++) {
    map2.put("key" + i, "value");
}
System.out.println("Without expansion takes nanoTime = " + (System.nanoTime() - start2));

HashMap<String, String> map = new HashMap<>(20);
final long start1 = System.nanoTime();
for (int i = 0; i < 20; i++) {
    map.put("key" + i, "value");
}
System.out.println("Expansion takes nanoTime = " + (System.nanoTime() - start1));
```

this just generated (with extremely large fluctuation)

```
Without expansion takes nanoTime = 181705
Expansion takes nanoTime = 90335
```

lujiajing1126: > Is this performance sensitive? Bigger factor actually costs more memory. Generally, this is a tradeoff.

I wrote a simple microbenchmark to test the impact of the load factor, which can be found here,

https://gist.github.com/lujiajing1126/17c5372af0e940a21067a3e6f8554594

I scan the parameter space from 1 to 20 (not even), and the result shows no significant difference.

So I'm against this kind of modification.

Superskyyy: @Oliverwqcwrw your test is not measured correctly due to the JVM warmup. I also think this load factor should not be tempered here unless there's solid empirical/ mathematical evidence.

Oliverwqcwrw: > @Oliverwqcwrw your test is not measured correctly due to the JVM warmup. The first part will always take much longer unless you run few things before. I also think this load factor should not be tempered here unless there's solid empirical/ mathematical evidence.

Thanks for your solution, I do wonder why my test result is so strange, it seems that the future test will write a benchmark test

Oliverwqcwrw: > > Is this performance sensitive? Bigger factor actually costs more memory. Generally, this is a tradeoff.
> 
> I wrote a simple microbenchmark to test the impact of the load factor, which can be found here,
> 
> https://gist.github.com/lujiajing1126/17c5372af0e940a21067a3e6f8554594
> 
> I scan the parameter space from 1 to 20 (not even), and the result shows no significant difference.
> 
> So I'm against this kind of modification.

Thanks for your reply,I did have a problem with my test

wu-sheng: > > @Oliverwqcwrw your test is not measured correctly due to the JVM warmup. The first part will always take much longer unless you run few things before. I also think this load factor should not be tempered here unless there's solid empirical/ mathematical evidence.
> 
> 
> 
> Thanks for your solution, I do wonder why my test result is so strange, it seems that the future test will write a benchmark test

Learn JHM and warnup concept. JVM 's JIT are relative about this

Don't measure one method if it is too fast, try 1000 to 100000 as a unit.

Oliverwqcwrw: Thank you very much. Learned a lot. It's a very friendly community

wu-sheng:
Your codes cause NPE. You should run tests before submitting the changes to upstream.

Oliverwqcwrw:
Sorry, I forgot to execute test, let me fix it

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -127,7 +127,7 @@ public static class Agent {
          * The identifier of the instance
          */
         @Length(50)
-        public volatile static String INSTANCE_NAME = "";
+        public static volatile String INSTANCE_NAME = "";
```
wu-sheng, line 130: What is the point of doing this? Same question about `static final` or `final static`. They have no difference in compiled codes.

Oliverwqcwrw: I changed the order of modifiers in accordance with the Java specification

Oliverwqcwrw: The Java Language Specification recommends listing modifiers in the following order:

> Annotations
public
protected
private
abstract
static
final
transient
volatile
synchronized
native
default
strictfp

wu-sheng: This is not a university study or research. We just do enginerring meaningful.

Oliverwqcwrw: Okay, I'm going to undo the modification of the modifier order

Superskyyy: > Okay, I'm going to undo the modification of the modifier order

Just want to point out that this kind of change is not wrong and most likely positive in terms of code quality, but it goes in the direction of unnecessarily costing the reviewer's effort. We would appreciate it more if you could in the future work on functional changes and we are more than happy to see more pull requests from you like that.

Oliverwqcwrw: I agree with you

wu-sheng:
So, should I expect this is your last fix in the near future? Didn't see you fix more

Oliverwqcwrw:
Yes, I have nothing else to fix

wu-sheng (approved):
LGTM

PR #168, opened 2022-05-04
gRPC plugin support trace client generic call

hutaishi:
skywalking grpc plugin support trace client generic call. Fix https://github.com/apache/skywalking/issues/8936

### topology
image

### trace
image

#### only trace grpc generic call
Determine if the request parameter is a DynamicMessage
```java
allArguments[1] instanceof com.google.protobuf.DynamicMessage
```

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/client/ClientCallImplGenericCallInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/client/ClientCallImplGenericCallInterceptor.java
@@ -71,6 +71,7 @@ public class ClientCallImplGenericCallInterceptor
         if (StringUtil.isBlank(asyncCallMethod)) {
             return;
         }
+        ContextManager.getRuntimeContext().remove(GENERIC_CALL_METHOD);
 
         Listener<?> observer = (Listener<?>) allArguments[0];
         Metadata headers = (Metadata) allArguments[1];
@@ -113,6 +114,7 @@ public class ClientCallImplGenericCallInterceptor
         AbstractSpan span = (AbstractSpan) objInst.getSkyWalkingDynamicField();
         // Scenario of specifying IP + port
         String remotePeer = (String) ContextManager.getRuntimeContext().get(CLIENT_STREAM_PEER);
+        ContextManager.getRuntimeContext().remove(CLIENT_STREAM_PEER);
         span.setPeer(remotePeer);
         Arrays.stream(objInst.getClass().getDeclaredMethods()).filter(m -> m.getName().equals("getAttributes"))
                 .findFirst()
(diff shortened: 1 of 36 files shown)


apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/define/NettyClientStreamInstrumentation.java
```diff
@@ -0,0 +1,71 @@
...
+import net.bytebuddy.matcher.ElementMatcher;
+
+/**
+ * Intercept constructor to obtain server IP.
+ * grpc Low version(eg:1.6.0): use io.grpc.netty.NettyClientStream.
+ * grpc High version(eg:1.33.1): use io.grpc.netty.shaded.io.grpc.netty.NettyClientStream.
+ */
+public class NettyClientStreamInstrumentation extends ClassInstanceMethodsEnhancePluginDefine {
+
+    private static final String ENHANCE_CLASS = "io.grpc.netty.NettyClientStream";
```
wu-sheng, line 39: Intercepting netty class in the grpc plugin seems not a good idea. User or other RPC framework could use `NettyClientStream` too. Also, in `NettyClientStreamInterceptor`, all conditions are netty native relative.

This is the major concern for this plugin.

apm-sniffer/apm-sdk-plugin/grpc-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grpc/v1/client/ClientCallImplGenericCallInterceptor.java
```diff
@@ -0,0 +1,220 @@
...
+    @Override
+    public void onConstruct(EnhancedInstance objInst, Object[] allArguments) throws Throwable {
+        MethodDescriptor methodDescriptor = (MethodDescriptor) allArguments[0];
+        objInst.setSkyWalkingDynamicField(methodDescriptor);
+    }
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+            MethodInterceptResult result) throws Throwable {
+        String asyncCallMethod = (String) ContextManager.getRuntimeContext().get(GENERIC_CALL_METHOD);
```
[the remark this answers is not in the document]
hutaishi: I will follow this good practice to modify it.

test/plugin/scenarios/grpc-generic-call-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,105 @@
...
+            skipAnalysis: false
+            refs:
+              - { parentEndpoint: org.apache.skywalking.apm.testcase.grpc.proto.TestServiceDynamic.unary,
+                  networkAddress: '', refType: CrossThread,
+                  parentSpanId: 0, parentTraceSegmentId: not null,
+                  parentServiceInstance: not null, parentService: grpc-provider,
+                  traceId: not null }
+      - segmentId: not null
+        spans:
+          - operationName: org.apache.skywalking.apm.testcase.grpc.proto.TestServiceDynamic.unary
```
wu-sheng, line 54: Could you list the operation name format rules? I prefer we could keep consistent with gRPC service name(package + service name). Java package name seems a little strange as prefix.

hutaishi: I reused method of `OperationNameFormatUtil#formatOperationName`.
consistent: package + service name + method.
My integration test case package name is the same as the Java package name.
Java package names are not actually used.

wu-sheng (approved):

PR #171, opened 2022-05-08
Add log endpoint name

BFergerson:
### <Feature description>
- [x] If this is non-trivial feature, paste the links/URLs to the design doc.

---

As discussed in https://github.com/apache/skywalking/discussions/9007


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
```diff
@@ -410,6 +410,11 @@ public CorrelationContext getCorrelationContext() {
         return this.correlationContext;
     }
 
+    @Override
+    public String getPrimaryEndpointName() {
+        return (primaryEndpoint != null) ? primaryEndpoint.getName() : null;
```
wu-sheng, line 415: **primaryEndpoint** should not be NULL in any case. I don't think we need null-check.

apm-sniffer/apm-toolkit-activation/apm-toolkit-log4j-1.x-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/log/log4j/v1/x/log/GRPCLogAppenderInterceptor.java
```diff
@@ -96,6 +96,9 @@ private LogData transform(final AppenderSkeleton appender, LoggingEvent event) {
                         .build())
                 .setBody(LogDataBody.newBuilder().setType(LogDataBody.ContentCase.TEXT.name())
                                     .setText(TextLog.newBuilder().setText(transformLogText(appender, event)).build()).build());
+        if (ContextManager.getPrimaryEndpointName() != null) {
+            builder.setEndpoint(ContextManager.getPrimaryEndpointName());
```
wu-sheng, line 100: `getPrimaryEndpointName` doesn't have to be invoked twice. It is better `get name` and do null check?

wu-sheng (commented):
Check this, https://github.com/apache/skywalking-java/runs/6341901459?check_suite_focus=true, this change would affect e2e tests.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java
BFergerson: What about if it's called before the first push?

https://github.com/apache/skywalking-java/blob/bd7912106aede5f79dbcfcdafe5c13c4378d8629/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/TracingContext.java#L529

wu-sheng: How a tracing context existing without a span created? I think that would cause more issues.

wu-sheng: Take a deep look at how this method is being called, `ContextManager` can't create a TracingContext(AbstractTracerContext) without creating a span.

BFergerson: Looked into it further. You're right. Removed null check.

wu-sheng (approved):

PR #174, opened 2022-05-13
Add `plugin.jdbc.trace_sql_parameters` into Configuration Discovery S…

zhyyu:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<[9028](https://github.com/apache/skywalking/issues/9028)>.


wu-sheng:
I think we still need some discussions, why do you need adding a new method to `ConfigurationDiscoveryService`? This breaks many things and brings the plugin logic into the agent core.

You should change the `ConfigurationDiscoveryService` to support multiple watchers to watch one key, and implement the collecting checker inside the plugin only.

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/PSSetterDefinitionOfJDBCInstrumentation.java
```diff
@@ -40,11 +40,10 @@ public PSSetterDefinitionOfJDBCInstrumentation(boolean ignorable) {
     public ElementMatcher<MethodDescription> getMethodsMatcher() {
         ElementMatcher.Junction<MethodDescription> matcher = none();
 
-        if (JDBCPluginConfig.Plugin.JDBC.TRACE_SQL_PARAMETERS) {
-            final Set<String> setters = ignorable ? PS_IGNORABLE_SETTERS : PS_SETTERS;
-            for (String setter : setters) {
-                matcher = matcher.or(named(setter));
-            }
+        // remove TRACE_SQL_PARAMETERS judgement for dynamic config
```
wu-sheng, line 43: Static should be kept, dynamic should only override the default value.

zhyyu: There is a problem, `org.apache.skywalking.apm.plugin.jdbc.PSSetterDefinitionOfJDBCInstrumentation#getMethodsMatcher` only invoked once when instrument occur. if dynamic change later, it won't be affected, and `org.apache.skywalking.apm.plugin.jdbc.JDBCPreparedStatementSetterInterceptor#beforeMethod` will not be active even when the `TRACE_SQL_PARAMETERS` is true.

wu-sheng: Watcher should be initialized in the driver's interceptor, not in the instrumentation class, no matter in any case. I think the class loader issues you faced are because of this.

zhyyu: image

`org.apache.skywalking.apm.plugin.jdbc.PSSetterDefinitionOfJDBCInstrumentation#getMethodsMatcher` only occur once.  Whose interceptor  `org.apache.skywalking.apm.plugin.jdbc.JDBCPreparedStatementSetterInterceptor#beforeMethod` put sql parameter into `StatementEnhanceInfos `. If do not enhance these class, even though the dynamic config changed from false to true. `org.apache.skywalking.apm.plugin.jdbc.mysql.PreparedStatementExecuteMethodsInterceptor#beforeMethod` cannot get parameter in line `final Object[] parameters = cacheObject.getParameters();`

wu-sheng: Don't do anything in `PSSetterDefinitionOfJDBCInstrumentation`, which is in another class loader and only executed in the booting stage.

Try to initial watcher in `JDBCPreparedStatementSetterInterceptor`

zhyyu: > Don't do anything in `PSSetterDefinitionOfJDBCInstrumentation`, which is in another class loader and only executed in the booting stage.
> 
> Try to initial watcher in `JDBCPreparedStatementSetterInterceptor`

Yes, just because of `PSSetterDefinitionOfJDBCInstrumentation` only executed in the booting stage, it just executed once. But the `getMethodsMatcher` in  `PSSetterDefinitionOfJDBCInstrumentation` in the booting stage check the `JDBCPluginConfig.Plugin.JDBC.TRACE_SQL_PARAMETERS`. if it is false, `JDBCPreparedStatementSetterInterceptor` won't be active because no set method in `PreparedStatement` will be matched.

image

wu-sheng: You are asking a dynamic configuration, of course, you have to always instrument the target class. `PSSetterDefinitionOfJDBCInstrumentation` has to be changed. But this should not be the reason you are adding more things to the core level.
The core is unexpected to change in this case.

zhyyu:
Also , now `org.apache.skywalking.apm.plugin.jdbc.JDBCPluginConfig.Plugin.JDBC` is loaded by one classloader.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/dynamic/ConfigurationDiscoveryService.java
```diff
@@ -121,7 +121,8 @@ public void shutdown() throws Throwable {
     public void registerAgentConfigChangeWatcher(AgentConfigChangeWatcher watcher) {
         WatcherHolder holder = new WatcherHolder(watcher);
         if (register.containsKey(holder.getKey())) {
-            throw new IllegalStateException("Duplicate register, watcher=" + watcher);
+            LOGGER.debug("Duplicate register, watcher={}", watcher);
+            return;
```
wu-sheng, line 125: I think we need to enhance this logic. Duplicate watchers should not be always legal and ignore the second. We should check the input class and existing watcher class (including class loader level), if they are the same, we should be able to ignore the duplicate.
Think in this way, one config class could be loaded by different Agent ClassLoader objects as it is being used in two classloaders of service's codes.
If they are not matched, we should not ignore them, we should support more than one watcher instance in the `ConfigurationDiscoveryService`.

In your current implementation, the dynamic config updating could work on one class loader's SQL parameter collecting, but not for others, and randomly. This would be a very confused bug, and hard to locate when others face this.

zhyyu: > Think in this way, one config class could be loaded by different Agent ClassLoader objects as it is being used in two classloaders of service's codes.

But most time, we just register watcher in one place, so do not exist watcher is loaded by two classloader. Is it cost-effective to fit the situation we rarely use?

wu-sheng: Remember, you are changing agent core, so, some extra list is definitely acceptable as they are only created once.
But making core having known bugs seems not a good practice. After all, multiple class loaders are common in some runtime env.

wu-sheng: I was writing this kernel with much smaller cases, if I said users will not face those cases, people like you will be in a chaos.

zhyyu: Ok, can I change `ConfigurationDiscoveryService.Register#register` field from `Map<String, WatcherHolder> register` to `Map<String, List<WatcherHolder>> register` for supporting multi same key watcher?

wu-sheng: Yes, you should follow this way.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/dynamic/ConfigurationDiscoveryService.java
```diff
@@ -121,8 +122,13 @@ public void shutdown() throws Throwable {
...
         WatcherHolder holder = new WatcherHolder(watcher);
         if (register.containsKey(holder.getKey())) {
-            LOGGER.debug("Duplicate register, watcher={}", watcher);
-            return;
+            List<WatcherHolder> watcherHolderList = register.get(holder.getKey());
+            for (WatcherHolder watcherHolder : watcherHolderList) {
+                if (watcherHolder.getWatcher().getClass().equals(watcher.getClass())) {
+                    LOGGER.debug("Duplicate register, watcher={}", watcher);
+                    return;
+                }
```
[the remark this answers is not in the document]
zhyyu: the `else` logic is the same same as below `register.put(holder.getKey(), holder);

wu-sheng:
GitHub seems down, https://www.githubstatus.com/. I will be back when it is recovered.

wu-sheng:
GHA is back, but CI fails, please check.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/dynamic/ConfigurationDiscoveryService.java
```diff
@@ -238,17 +247,25 @@ private void getAgentDynamicConfig() {
...
-        private final Map<String, WatcherHolder> register = new HashMap<>();
+        private final Map<String, List<WatcherHolder>> register = new HashMap<>();
 
         private boolean containsKey(String key) {
             return register.containsKey(key);
         }
 
         private void put(String key, WatcherHolder holder) {
-            register.put(key, holder);
+            List<WatcherHolder> watcherHolderList = register.get(key);
```
wu-sheng, line 257: As the watcher list for one key is supported, I think we have to add a sync to this method.

zhyyu: > As the watcher list for one key is supported, I think we have to add a sync to this method.

I thought this sync problem, but if add syn,  `registerAgentConfigChangeWatcher` is synchronized, and `JDBCPreparedStatementSetterInterceptor` call the `registerAgentConfigChangeWatcher` method register the watcher  every time when prepared sql occur, I worry about it will effect the sql performance. And if we do not and syn to it, even if it register multi same wachers, it will not effect the `JDBCPreparedStatementSetterInterceptor` register case. Can we just add comment on `registerAgentConfigChangeWatcher` to announce it is thread unsafe?

wu-sheng: Then that is the issue of `JDBCPreparedStatementSetterInterceptor`. You should not register the watcher repeated, no matter with or without `sync`, it causes a huge perf. impact.

wu-sheng: Watcher should only be registered in the JDBC Driver initializing stage. Consider `JDBCDriverInterceptor` to initial this watcher.

zhyyu: > Then that is the issue of `JDBCPreparedStatementSetterInterceptor`. You should not register the watcher repeated, no matter with or without `sync`, it causes a huge perf. impact.

Can I use static code block or constructor code block in `JDBCPreparedStatementSetterInterceptor` to register the watcher once?

wu-sheng: > Can we just add comment on registerAgentConfigChangeWatcher to announce it is thread unsafe?

It is pointless, you are using this at this concurrency status.  Do you.

wu-sheng: > Can I use static code block or constructor code block in JDBCPreparedStatementSetterInterceptor to register the watcher once?

Why do you have to do that? That is really bad for the agent in practice. Potentially, this could cause a lot of class loader issues.

zhyyu: > > Can we just add comment on registerAgentConfigChangeWatcher to announce it is thread unsafe?
> 
> It is pointless, you are using this at this concurrency status. Do you.

I mean I add sync on `registerAgentConfigChangeWatcher` and remove the register logic out of `beforeMethod` in `JDBCPreparedStatementSetterInterceptor` using static or constructor code block. So it will not generate perf impact.

wu-sheng: `JDBCDriverInterceptor` is there, what do you concern?

wu-sheng: > > > Can we just add comment on registerAgentConfigChangeWatcher to announce it is thread unsafe?
> > 
> > 
> > It is pointless, you are using this at this concurrency status. Do you.
> 
> I mean I add sync on `registerAgentConfigChangeWatcher` and remove the register logic out of `beforeMethod` in `JDBCPreparedStatementSetterInterceptor` using static or constructor code block. So it will not generate perf impact.

I have answered this question.

**That is really bad for the agent in practice. Potentially, this could cause a lot of class loader issues.**

wu-sheng: `static` is never recommended in the interceptor, and all static blocks in the agent would be reviewed carefully. It is acceptable when it really has to be, and 100% correct. In the interceptor, I could jump to the conclusion, that `static` is not allowed.

zhyyu: Ok, I have removed the register logic to `JDBCDriverInterceptor` and add `synchronized` on `registerAgentConfigChangeWatcher` method.

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/TraceSqlParametersWatcher.java
```diff
@@ -0,0 +1,46 @@
...
+        super(propertyKey);
+        defaultValue = JDBCPluginConfig.Plugin.JDBC.TRACE_SQL_PARAMETERS;
+    }
+
+    @Override
+    public void notify(ConfigChangeEvent value) {
+        if (EventType.DELETE.equals(value.getEventType())) {
+            JDBCPluginConfig.Plugin.JDBC.TRACE_SQL_PARAMETERS = defaultValue;
+        } else {
+            JDBCPluginConfig.Plugin.JDBC.TRACE_SQL_PARAMETERS = Boolean.parseBoolean(value.getNewValue());
```
wu-sheng, line 37: You miss a `try/catch` and fall back to defaultValue logic here. Once user doesn't set `boolean` to this config value, partical dynamic configurations would be lost.

Only `ConfigurationDiscoveryCommandExecutor#execute` will print the final exception.

This is unexpected case.

zhyyu: > You miss a `try/catch` and fall back to defaultValue logic here. Once user doesn't set `boolean` to this config value, partical dynamic configurations would be lost.
> 
> Only `ConfigurationDiscoveryCommandExecutor#execute` will print the final exception.
> 
> This is unexpected case.

```java
public static boolean parseBoolean(String s) {
        return ((s != null) && s.equalsIgnoreCase("true"));
    }
```

I check the `Boolean.parseBoolean`, it won't throw any exception. If user doesn't set `boolean` to this config, the config will be set to false.

wu-sheng (approved):

wu-sheng (approved):

PR #175, opened 2022-05-13
Fix Span not finished in gateway plugin when the gateway request time out

wuwen5:
- [x] fix https://github.com/apache/skywalking/issues/9061.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Release Notes.
 * Grpc plugin support trace client async generic call(without grpc stubs), support Method type: `UNARY`、`SERVER_STREAMING`.
 * Enhance Apache ShenYu (incubating) plugin: support trace `grpc`,`sofarpc`,`motan`,`tars` rpc proxy.
 * Add primary endpoint name to log events.
+* Fix Span not finished in gateway plugin when the gateway request timeout.
 
 #### Documentation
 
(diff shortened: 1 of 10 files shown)


CHANGES.md
```diff
@@ -16,6 +16,7 @@ Release Notes.
 * Grpc plugin support trace client async generic call(without grpc stubs), support Method type: `UNARY`、`SERVER_STREAMING`.
 * Enhance Apache ShenYu (incubating) plugin: support trace `grpc`,`sofarpc`,`motan`,`tars` rpc proxy.
 * Add primary endpoint name to log events.
+* Fix Span not finished in gateway plugin when the gateway request timeout.
```
wu-sheng, line 19: Is this a Gateway 3.x plugin fix only? Or others may have similar issues?

test/plugin/scenarios/gateway-3.x-scenario/gateway-projectA-scenario/src/main/resources/application.yml
```diff
@@ -20,8 +20,14 @@ server:
...
   cloud:
     gateway:
+      httpclient:
+        connect-timeout: 2000
       routes:
       - id: provider_route
         uri: http://localhost:18070
         predicates:
         - Path=/provider/b/*
+      - id: provider_timeout
```
wu-sheng, line 30: You are adding a `timeout` calling, please update the expected data too.

CHANGES.md
wuwen5: 2.1.x\3.x

wu-sheng (approved):
Thanks for fixing.

PR #181, opened 2022-05-21
Fix bug log4j2 grpc report missing traceId

andotorg:
==== 🐛 Remove this line WHEN AND ONLY WHEN you're fixing a bug, follow the checklist 👇 ====
### Fix https://github.com/apache/skywalking/discussions/9113
- [✅] Fix the problem that traceid cannot be reported when log4j2 global is modified to asynchronous log
==== 🐛 Remove this line WHEN AND ONLY WHEN you're fixing a bug, follow the checklist 👆 ====

- [✅] Update the [`CHANGES` log](https://github.com/apache/skywalking-java/blob/main/CHANGES.md).

PR #182, opened 2022-05-22
Support  log4j2 contextSelector as AsyncLoggerContextSelector  in gRPC log report

andotorg:
==== 🐛 Remove this line WHEN AND ONLY WHEN you're fixing a bug, follow the checklist 👇 ====

Fix https://github.com/apache/skywalking/discussions/9113
[✅] Fix the problem that traceid cannot be reported when log4j2 global is modified to asynchronous log
==== 🐛 Remove this line WHEN AND ONLY WHEN you're fixing a bug, follow the checklist 👆 ====

[✅] Update the [CHANGES log](https://github.com/apache/skywalking-java/blob/main/CHANGES.md).
[✅] Update the [Log4j2 log](`application-toolkit-log4j-2.x.md`).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@ Release Notes.
 * Enhance Apache ShenYu (incubating) plugin: support trace `grpc`,`sofarpc`,`motan`,`tars` rpc proxy.
 * Add primary endpoint name to log events.
 * Fix Span not finished in gateway plugin when the gateway request timeout.
+* Support `-Dlog4j2.contextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector` in gRPC log report.
 
 #### Documentation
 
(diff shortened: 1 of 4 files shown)


andotorg:
image
image

andotorg:
@wu-sheng hi， I don't quite understand the reason for this failure. What should I do to solve this failed task？

andotorg:
Mark it https://github.com/apache/skywalking-java/pull/181

wu-sheng:
> @wu-sheng hi， I don't quite understand the reason for this failure. What should I do to solve this failed task？

I just checked, I think this is not relative to you. `https://logback.qos.ch/manual/appenders.html#AsyncAppender` is in the doc, but somehow, qos.ch's CA fails(maybe expired). 
You don't have to worry about this.

wu-sheng (approved):

PR #191, opened 2022-06-02
string "/eureka" should match pattern "/eureka/*" or "/eureka/**" 

xjs1919:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@ Release Notes.
 * Support `-Dlog4j2.contextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector` in gRPC log report.
 * Fix tcnative libraries relocation for aarch64.
 * Add `plugin.jdbc.trace_sql_parameters` into Configuration Discovery Service.
+* Enhance TracePathMatcher, "/eureka/*" or "/eureka/**" can match "/eureka".
 
 #### Documentation
 
--- a/apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
+++ b/apm-sniffer/optional-plugins/trace-ignore-plugin/src/main/java/org/apache/skywalking/apm/plugin/trace/ignore/matcher/FastPathMatcher.java
@@ -73,12 +73,12 @@ public class FastPathMatcher implements TracePathMatcher {
             // pattern: a/*   a/**
             //          ↓     ↓
             //string:   a     a
-            if(pc=='/' && sc == 0){
+            if (pc == '/' && sc == 0) {
                 p++;
                 if (safeCharAt(pat, p) == '*') {
                     p++;
                     return multiWildcardMatch(pat, p, str, s);
-                }else{
+                } else {
                     return false;
                 }
             }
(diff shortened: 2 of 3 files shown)


devkanro:
After my test, AntPathMatcher do not match this rule.

```Kotlin
AntPathMatcherAntPathMatcher().match("/eruka/*", "/eruka") // false
```

wu-sheng:
If this does follow ant path match, I have concerns about this, whether this us reasonable.

devkanro:
> If this does follow ant path match, I have concerns about this, whether this is reasonable.

@wu-sheng `/eureka/**` matches `/eureka`, but `/eureka/*` not.

wu-sheng:
OK, got your point. Both should support.

wu-sheng:
I just want to ask, whether 2 configs make more sense than supporting this kind of config? After all, /* means at least having a /.

My concern is about this.

wu-sheng:
Is there other tools supporting in this way?

devkanro:
@wu-sheng 
No, I means the current behavior of AntPathMatcher is `/eureka/**` matches `/eureka`, but `/eureka/*` not.

devkanro:
I am don't sure my FastPathMatcher's behavior, I will get test for it.

devkanro:
There are my test result, use 'AntPathMatcher/FastPathMatcher'

```Kotlin
matches("/eureka/*", "/eureka") = false/false
matches("/eureka/*", "/eureka/") = true/true
matches("/eureka/*", "/eureka/test/") = false/true
matches("/eureka/*", "/eureka/test") = true/true
matches("/eureka/*/", "/eureka") = false/false
matches("/eureka/*/", "/eureka/") = true/false
matches("/eureka/*/", "/eureka/test") = false/false
matches("/eureka/*/", "/eureka/test/") = true/true
matches("/eureka/**", "/eureka") = true/false
matches("/eureka/**", "/eureka/") = true/true
matches("/eureka/**", "/eureka/test/") = true/true
matches("/eureka/**", "/eureka/test") = true/true
matches("/eureka/**/", "/eureka") = true/false
matches("/eureka/**/", "/eureka/") = true/true
matches("/eureka/**/", "/eureka/test") = true/true
matches("/eureka/**/", "/eureka/test/") = true/true
```

devkanro:
Highlights for diff

```
matches("/eureka/*", "/eureka/test/") = false/true
matches("/eureka/*/", "/eureka/") = true/false
matches("/eureka/**", "/eureka") = true/false
matches("/eureka/**/", "/eureka") = true/false
```

wu-sheng:
Are these diff from previous or PR version?

devkanro:
The diff is about Spring's AntPathMatcher between my FastPathMatcher in current production version of skywalking.

I am not compare the PR version.

devkanro:
By comparison, FastPathMatcher seems more intuitive?

Case 1:
`/eureka/*` vs `/eureka/test/`
* means zero or one path part, but a path part may have ends with / or not.

Case 2:
`/eureka/*/` vs `/eureka/`
* means zero or one path part, but if it means zero, the path should be `/eureka//`, can we omit the double /?

Case 3:
`/eureka/**` vs `/eureka`
/** means at least one /

Case 4:
`/eureka/**/` vs `/eureka`
**/ means path should ends with /

wu-sheng:
I feel like the current version is good. I am not sure why Spring works in a different way.

devkanro:
Spring's AntPathMatcher is copy from Apache Ant.

wu-sheng:
@xjs1919 Please update with us, why do you think this behavior makes sense?

xjs1919:
> @xjs1919 Please update with us, why do you think this behavior makes sense?

i want to ignore path contains "actuator", so i have to config like this :

trace.ignore_path=${SW_AGENT_TRACE_IGNORE_PATH:**/actuator/**,GET:/actuator}

it works, however a little redundant and not graceful，and Spring's AntPathMatcher can match "/eureka" with pattern "/eureka/**"

wu-sheng:
I am not sure why AntPathMatcher works like that, but to be honest, I don't think it is the standard. 
`/eureka/**` ignoring `/eureka` seems unreasonable. Maybe Spring team has their reasons, but logically, this is right.

wu-sheng:
Closing due to support @devkanro 's opinion(original author)

PR #193, opened 2022-06-08
Fix ArgumentTypeNameMatch doesn't work for array argument

ForestWang123:
I found ArgumentTypeNameMatch doesn't work for array argument (e.g. the STRING_ARRAY_ARGUMENT_TYPE in org.apache.skywalking.apm.plugin.jdbc.postgresql.define.ConnectionInstrumentation )

Fix <ArgumentTypeNameMatch doesn't work for array argument>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bytebuddy/ArgumentTypeNameMatch.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bytebuddy/ArgumentTypeNameMatch.java
@@ -60,7 +60,7 @@ public class ArgumentTypeNameMatch implements ElementMatcher<MethodDescription> 
     public boolean matches(MethodDescription target) {
         ParameterList<?> parameters = target.getParameters();
         if (parameters.size() > index) {
-            return parameters.get(index).getType().asErasure().getName().equals(argumentTypeName);
+            return parameters.get(index).getType().getActualName().equals(argumentTypeName);
         }
 
         return false;
--- a/docs/en/contribution/compiling.md
+++ b/docs/en/contribution/compiling.md
@@ -7,6 +7,8 @@ Prepare JDK 8+.
 ```shell
 git clone https://github.com/apache/skywalking-java.git
 cd skywalking-java
+git submodule init
+git submodule update
 ./mvnw clean package -Pall
 ```
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@ Release Notes.
 * Support `-Dlog4j2.contextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector` in gRPC log report.
 * Fix tcnative libraries relocation for aarch64.
 * Add `plugin.jdbc.trace_sql_parameters` into Configuration Discovery Service.
+* Fix ArgumentTypeNameMatch doesn't work for array argument
 
 #### Documentation
 
--- a/apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/bytebuddy/ArgumentTypeNameMatchTest.java
+++ b/apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/plugin/bytebuddy/ArgumentTypeNameMatchTest.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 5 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bytebuddy/ArgumentTypeNameMatch.java
```diff
@@ -60,7 +60,7 @@ private ArgumentTypeNameMatch(int index, String argumentTypeName) {
     public boolean matches(MethodDescription target) {
         ParameterList<?> parameters = target.getParameters();
         if (parameters.size() > index) {
-            return parameters.get(index).getType().asErasure().getName().equals(argumentTypeName);
```
wu-sheng, line 63: What does Array return in this case?

ForestWang123:
> 
I have checked, https://nvd.nist.gov/vuln/detail/CVE-2019-17638 is not a dead link.
Is anything wrong?

docs/en/contribution/compiling.md
```diff
@@ -7,6 +7,8 @@ Prepare JDK 8+.
 ```shell
 git clone https://github.com/apache/skywalking-java.git
 cd skywalking-java
+git submodule init
+git submodule update
```
wu-sheng, line 11: These two steps are not necessary. There is `-Pall` including these

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/bytebuddy/ArgumentTypeNameMatch.java
wu-sheng: When you read the Java class file spec, you would notice `[Lorg.apache.skywalking.apm.agent.core.plugin.bytebuddy.Person` is actually the correct expression for `Person[]`

You could learn this from Java specification. This is official.

ForestWang123: Yes, I know that.

But parameters.get(index).getType().asErasure().getName() will cause some Instrumentation classes fail, e.g. org.apache.skywalking.apm.plugin.jdbc.postgresql.define.ConnectionInstrumentation, and org.apache.skywalking.apm.plugin.jdbc.postgresql.define.PgStatementInstrumentation.

Of course,  we can let plugin authors write in "[Lorg.apache.skywalking.apm.agent.core.plugin.bytebuddy.Person". But don't you think "Person[]" is better for readability?

wu-sheng: > But parameters.get(index).getType().asErasure().getName() will cause some Instrumentation classes fail, e.g. org.apache.skywalking.apm.plugin.jdbc.postgresql.define.ConnectionInstrumentation, and org.apache.skywalking.apm.plugin.jdbc.postgresql.define.PgStatementInstrumentation.

What does this `but` mean? They are working before your change.

ForestWang123: No,they cannot define the InterceptPoint correctly. 
  Because java.lang.String[]  is not equal to [java.lang.String in org.apache.skywalking.apm.plugin.jdbc.postgresql.define.ConnectionInstrumentation

wu-sheng: OK, maybe tests miss those cases, so don't detect it. 

I think you should fix those rather than changing this API implementation. Also, I would recommend to updating plugin dev doc to highlight this.

ForestWang123: java.lang.String[] is more friendly and lower knowledge requirement for plugin authors than [java.lang.String, isn't it?

wu-sheng: > java.lang.String[] is more friendly and lower knowledge requirement for plugin authors than [java.lang.String, isn't it?

This is a kind of instrumentation, so, we should be closing to Java class file format, rather than Java source codes. 
Let's fix those wrong instrumentation definitions, to use `[L`. Also, we should add a parameter check for `ArgumentTypeNameMatch`. If the given value is using `[]` suffix, the kernel should report an exception with a clear message saying, please use `[L` to define an Array type, and ref to Java specification.

Would you like to open another pull request to fix these?

wu-sheng: Another similar story in the backend OAL engine, as it is on the top of Javassist, another binary code manipulator, we can't use `int-Integer` boxing and unboxing, as it doesn't exist in the binary code level at all.
So, we don't want to hide this tech details, we could add verifications to avoid the bugs you mentioned above.

wu-sheng:
I just labeled this as `rejected`, because we are going to expect another new PR. 
Keep this open until the new PR opened, tracking there are several issues to fix.

@ForestWang123 Thanks for reporting this.

wu-sheng:
Fix through https://github.com/apache/skywalking-java/pull/194

PR #194, opened 2022-06-09
Fix argument type name in postgresql plugin and add type name checking

ForestWang123:
1. fix argument type name of Array in postgresql-8.x-plugin from java.lang.String[] to [Ljava.lang.String; 
2. add type name checking in ArgumentTypeNameMatch and ReturnTypeNameMatch
3. highlight the type naming rule in docs/en/setup/service-agent/java-agent/Java-Plugin-Development-Guide.md

### Fix <argument type name of Array in postgresql-8.x-plugin>

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/postgresql-8.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/postgresql/define/PgStatementInstrumentation.java
+++ b/apm-sniffer/apm-sdk-plugin/postgresql-8.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/postgresql/define/PgStatementInstrumentation.java
@@ -26,6 +26,7 @@ import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInst
 import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
 
 import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
 import static org.apache.skywalking.apm.agent.core.plugin.bytebuddy.ArgumentTypeNameMatch.takesArgumentWithType;
 import static org.apache.skywalking.apm.agent.core.plugin.match.NameMatch.byName;
 import static org.apache.skywalking.apm.plugin.jdbc.postgresql.Variables.PG_STATEMENT_EXECUTE_METHOD_INTERCEPTOR;
@@ -43,8 +44,11 @@ public class PgStatementInstrumentation extends ClassInstanceMethodsEnhancePlugi
             new InstanceMethodsInterceptPoint() {
                 @Override
                 public ElementMatcher<MethodDescription> getMethodsMatcher() {
-                    return (named("execute").or(named("executeQuery")).or(named("executeUpdate")).or(named("executeLargeUpdate")))
-                            .and(takesArgumentWithType(0, "java.lang.String"));
+                    return named("execute").and(takesArgumentWithType(0, "java.lang.String")).and(takesArguments(1))
+                            .or(named("execute").and(takesArgumentWithType(1, "[Ljava.lang.String;")))
+                            .or(named("executeQuery"))
+                            .or(named("executeUpdate").and(takesArgumentWithType(0, "java.lang.String")).and(takesArguments(1)))
+                            .or(named("executeUpdate").and(takesArgumentWithType(1, "[Ljava.lang.String;")));
                 }
 
                 @Override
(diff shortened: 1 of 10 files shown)


apm-sniffer/apm-sdk-plugin/postgresql-8.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/postgresql/define/PgStatementInstrumentation.java
```diff
@@ -43,12 +43,8 @@ public InstanceMethodsInterceptPoint[] getInstanceMethodsInterceptPoints() {
             new InstanceMethodsInterceptPoint() {
                 @Override
                 public ElementMatcher<MethodDescription> getMethodsMatcher() {
-                    return named("execute").and(takesArgumentWithType(0, "java.lang.String"))
-                                           .or(named("execute").and(takesArgumentWithType(0, "java.lang.String[]")))
```
wu-sheng, line 47: Is this removed on purpose?

docs/en/setup/service-agent/java-agent/Java-Plugin-Development-Guide.md
```diff
@@ -297,7 +297,16 @@ public interface InstanceMethodsInterceptPoint {
...
 ```
 You may also use `Matcher` to set the target methods. Return **true** in `isOverrideArgs`, if you want to change the argument
-ref in interceptor.
+ref in interceptor. 
+Please refer to [bytebuddy](https://bytebuddy.net/#/) for details of defining `ElementMatcher`.
+
+In Skywalking, we provide 3 classes to facilitate `ElementMatcher` definition:
+* `AnnotationTypeNameMatch`: Depends on whether there are certain annotations in the target methods.
+* `ReturnTypeNameMatch`: Based on return type name (package name + `.` + class name) of the target methods.
+* `ArgumentTypeNameMatch`: Based on argument index and type name (package name + `.` + class name) of the target methods.
```
wu-sheng, line 306:
```suggestion
* `AnnotationTypeNameMatch`: Check on whether there is a certain annotation in the target method.
* `ReturnTypeNameMatch`: Check the return type name (package name + `.` + class name) of the target method.
* `ArgumentTypeNameMatch`: Check on the argument index and the type name (package name + `.` + class name) of the target method.
```

docs/en/setup/service-agent/java-agent/Java-Plugin-Development-Guide.md
```diff
@@ -297,7 +297,16 @@ public interface InstanceMethodsInterceptPoint {
...
+
+**Attention**:
+* For `ReturnTypeNameMatch` and `ArgumentTypeNameMatch`, please use [Lxxx; (Java file format defined in [JVM Specification](https://docs.oracle.com/javase/specs/jvms/se8/html/)) to define an Array type. For example, you should write [Ljava.lang.String; for `java.lang.String[]`.
```
wu-sheng, line 309:
```suggestion
* In case of using `ReturnTypeNameMatch` and `ArgumentTypeNameMatch`, use [Lxxx; (Java file format defined in [JVM Specification](https://docs.oracle.com/javase/specs/jvms/se8/html/)) to define an Array type. For example, you should write `[Ljava.lang.String;` for `java.lang.String[]`.
```

wu-sheng:
This PR is generally good, thanks for fixing quickly. Please polish based on review comments.

apm-sniffer/apm-sdk-plugin/postgresql-8.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdbc/postgresql/define/PgStatementInstrumentation.java
wu-sheng: I noticed in `org.postgresql.jdbc.PgStatement`, there is no `String[]` at index 0 for `execute` and `executeUpdate` method.

wu-sheng: Meanwhile `public int executeUpdate(String sql, String[] columnNames)` seems having a different execution path. Is this another bug? (should be `takesArgumentWithType(1...)?`

ForestWang123: OK, There are some problems of my implementation here.  
I have tightened the matching rule for execute and executeUpdate; delete the executeLargeUpdate methods which have not been implemented in PgStatement.

Please check, thank you

wu-sheng (approved):

PR #206, opened 2022-06-15
Optimize gRPC Log reporter to set service name for the first element

wu-sheng:

It was setting the service name for every element, which cause unnecessary resources and networks.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
@@ -124,6 +124,7 @@ public class LogReportServiceClient implements BootService, GRPCChannelListener,
                     // https://github.com/apache/skywalking-data-collect-protocol/blob/master/logging/Logging.proto
                     // Log collecting protocol defines LogData#service is required in the first element only.
                     logData.setService(Config.Agent.SERVICE_NAME);
+                    isFirst = false;
                 }
                 logDataStreamObserver.onNext(logData.build());
             }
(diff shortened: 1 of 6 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/remote/LogReportServiceClient.java
```diff
@@ -95,39 +94,46 @@ public void consume(final List<LogData> dataList) {
...
+                        }
+                    });
+
+            boolean isFirst = true;
+            for (final LogData.Builder logData : dataList) {
+                if (isFirst) {
+                    // Only set service name of the first element in one stream
+                    // https://github.com/apache/skywalking-data-collect-protocol/blob/master/logging/Logging.proto
+                    // Log collecting protocol defines LogData#service is required in the first element only.
+                    logData.setService(Config.Agent.SERVICE_NAME);
```
kezhenxu94, line 126:
```suggestion
                    logData.setService(Config.Agent.SERVICE_NAME);
                    isFirst = false;
```

kezhenxu94 (dismissed):

mrproliu (dismissed):

mrproliu (approved):

wu-sheng:
Follow https://github.com/apache/skywalking/blob/d3e38ec69780f2a5c111ee557a056f0f736a838b/oap-server/server-receiver-plugin/skywalking-log-recevier-plugin/src/main/java/org/apache/skywalking/oap/server/recevier/log/provider/handler/grpc/LogReportServiceGrpcHandler.java#L69-L78, which made @mrproliu attentions due to https://github.com/apache/skywalking/issues/9230

PR #215, opened 2022-06-20
[Bug #9250]shenyu plugin fix IgnoredTracerContext traceId NPE

hutaishi:
### Fix <Apache Shenyu plugin NPE #9250>  

When the implementation class of AbstractTracerContext is IgnoredTracerContext, the DistributedTraceId obtained from the context is empty, resulting in a null pointer exception for obtaining the traceId

Fix https://github.com/apache/skywalking/issues/9250

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@ Release Notes.
 
 8.12.0
 ------------------
-* Fix `Shenyu plugin` IgnoredTracerContext traceId NPE(Use with the apm-trace-ignore-plugin plugin will trigger)
+* Fix `Shenyu plugin`'s NPE in reading read trace ID when IgnoredTracerContext is used in the context.
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@ Release Notes.
 
 8.12.0
 ------------------
-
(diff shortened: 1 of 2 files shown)


wu-sheng:
Please update the changelog, 8.11.0 is out today, so, we need this fix in 8.12.0 changes.

hutaishi:
> Please update the changelog, 8.11.0 is out today, so, we need this fix in 8.12.0 changes.

Thanks for the reminder. Now updated

CHANGES.md
```diff
@@ -4,7 +4,7 @@ Release Notes.
 
 8.12.0
 ------------------
-
+* Fix `Shenyu plugin` IgnoredTracerContext traceId NPE(Use with the apm-trace-ignore-plugin plugin will trigger)
```
wu-sheng, line 7:
```suggestion
* Fix `Shenyu plugin`'s NPE in reading read trace ID when IgnoredTracerContext is used in the context.
```

wu-sheng (approved):

CHANGES.md
wu-sheng: IgnoredTracerContext could be activated in many cases, such as ignore-plugin, sampling policy activated, and agent disconnected with the OAP server.

hutaishi: > IgnoredTracerContext could be activated in many cases, such as ignore-plugin, sampling policy activated, and agent disconnected with the OAP server.

Understood, not enough consideration at the time

PR #228, opened 2022-06-22
update witness class in elasticsearch-6.x-plugin

xzyJavaX:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes [#9258](https://github.com/apache/skywalking/issues/9258)


apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/define/RestHighLevelClientInstrumentation.java
```diff
@@ -260,6 +260,6 @@ public StaticMethodsInterceptPoint[] getStaticMethodsInterceptPoints() {
 
     @Override
     protected String[] witnessClasses() {
-        return new String[] {Constants.TASK_TRANSPORT_CHANNEL_WITNESS_CLASSES};
+        return new String[] {Constants.CREATE_INDEX_RESPONSE_WITNESS_CLASSES};
```
wu-sheng, line 263: So, **TASK_TRANSPORT_CHANNEL_WITNESS_CLASSES** as witness class is working for other interceptors?

wu-sheng:
I think you break the CI, please recheck.

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/define/RestHighLevelClientInstrumentation.java
xzyJavaX: There are no other problems for the time being.

wu-sheng: What do you mean? Why does this matter and others are not?

I can see `org.elasticsearch.transport.TaskTransportChannel` is used for 6 instrumentation definitions, if this witness class is changed, but others are not, I would assume, there are still other interceptors are going to run?

wu-sheng: @xzyJavaX We need feedback about this before we merge this change.

xzyJavaX: We confirm that TASK_TRANSPORT_CHANNEL_WITNESS_CLASSES is working for other interceptors.And the version of transport-client can be changed from 6.7.1-6.8.4 to 6.2.0-6.8.4.Test cases will be added later.

test/plugin/scenarios/elasticsearch-rest-high-level-6.x-scenario/config/expectedData.yaml
```diff
@@ -307,124 +307,6 @@ segmentItems:
       - {key: db.type, value: Elasticsearch}
       - {key: db.instance, value: not null}
       skipAnalysis: 'false'
-    - operationName: Elasticsearch/IndexRequest
```
wu-sheng, line 310: Why do these get removed?

xzyJavaX: Moved to elasticsearch-transport-6.x-scenario

xzyJavaX: The version of rest-high-level-client changes too much, cannot be cover with a test case, in order not to affect the test of transport, so there are two cases.

wu-sheng: It is fine you don't add more versions to support.
Please update doc according to the latest status.

xzyJavaX: Supported-list.md updated.

wu-sheng (approved):

PR #229, opened 2022-06-23
[ISSUE #9264] fix onHalfClose span operation name

drgnchan:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/9264

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Release Notes.
 8.12.0
 ------------------
 * Fix `Shenyu plugin`'s NPE in reading read trace ID when IgnoredTracerContext is used in the context.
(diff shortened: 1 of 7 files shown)


wu-sheng:
Please update changes.md in the root.

wu-sheng:
I think you forgot to change gRPC relative tests? The plugin tests fail.

drgnchan:
> I think you forgot to change gRPC relative tests? The plugin tests fail.

Yes.I think i know how to fix it.I am sorry that i cannot fix it right now without a pc.I will start to fix it in 2 hours later.

wu-sheng:
Don't hurry. Fix it when you have time.

wu-sheng (approved):

PR #230, opened 2022-06-23
Remove --batch-mode in CI

wu-sheng:

Remove `--batch-mode` as our UTs seem not stable in this case.


kezhenxu94 (commented):
`--batch-mode` basically means run maven command in non-interactive (and disable colors, download progress, etc.), it doesn't mean run the tasks in parallel or in batch.

wu-sheng:
OK, but this error really looks like a race condition issue. So, I thought this is it.

PR #258, opened 2022-06-24
Save http status code regardless of it's value

kezhenxu94:

Previously in most of the plugins we only save the status code when it's `>=400`, and in some plugins we save the status code regardless it's value, this patch make the behaviors consistent in all plugins, so that users can configure `OAL` to compute metrics of status codes.


kezhenxu94:
Finally, all tests passed

wu-sheng (approved):

PR #266, opened 2022-07-01
trace-ignore-plugin can also ignore LocalSpan and ExitSpan

xjs1919:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Release Notes.
 * Fix `onHalfClose` using span operation name `/Request/onComplete` instead of the worng name `/Request/onHalfClose`.
 * Add plugin to support RESTeasy 4.x.
 * Add plugin to support hutool-http 5.x.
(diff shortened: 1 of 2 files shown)


wu-sheng:
Sorry, this is not what sampling means in SkyWalking. Also, this causes half noop span included in trace, which also breaks a trace.

xjs1919:
how can i ignore the local span and the exit span？

wu-sheng:
You could remove relative plugin, or change source codes.
This is not just concept conflict, also cause unpredictable bugs.

xjs1919:
”remove relative plugin" can not satisfy this kind of situation：
i want to see the detailed trace spans only when occurs an error in production environment，when there is no error，i just want to ignore these spans。 skywalking provides dynamic configurations， when combined with the dynamic configuration， i can resolve this situation。

wu-sheng:
You can do what you feel right. You have source codes, don't need our permission.
The review is from upstream perspective, we are clear sampling focuses on segment, not span, very clearly.

PR #267, opened 2022-07-03
fix bug

JoeKerouac:
### Fix span was created but never stop, in spring-cloud-gateway plugin(both v2.1.x and v3.x)
- [x] stop span in afterMethod


wu-sheng:
We need your scenario and explanation of why span isn't stopped correctly. There are test cases verified span are reported as expected. So you are fixing the cases we didn't test.

Also, you should add these cases into the plugin test scenarios to verify.

JoeKerouac:
@wu-sheng when you import apm-spring-cloud-gateway-2.1.x-plugin or apm-spring-cloud-gateway-3.x-plugin, and import spring-cloud-gateway, spring-cloud-gateway-x.x.x-plugin will work, and LocalSpan will be created in `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v3x.NettyRoutingFilterInterceptor#beforeMethod` (or `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v21x.NettyRoutingFilterInterceptor#beforeMethod`), but not release it in `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v3x.NettyRoutingFilterInterceptor#afterMethod` (or `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v21x.NettyRoutingFilterInterceptor#afterMethod`);

**This will cause memory leak**;

wu-sheng:
I can get the idea `NETTY_ROUTING_FILTER_TRACED_ATTR` isn't being removed, but you mentioned span doesn't stop, which makes me confused. If the span is not stopped, plugin test can't be passed. 

https://github.com/apache/skywalking-java/blob/6eeacf0959cf3837a63acb1b674291182d90bbc7/test/plugin/scenarios/gateway-3.x-scenario/config/expectedData.yaml#L96
In this expected file, this span has been reported to collector.

wu-sheng:
@wallezhang Could you take a look at this?

JoeKerouac:
@wu-sheng In fact it is, If you're talking about unit tests, unit tests for this plugin are inherently problematic; spans are manually closed in unit tests;

wu-sheng:
No, I am talking about this real use case. 

https://github.com/apache/skywalking-java/tree/6eeacf0959cf3837a63acb1b674291182d90bbc7/test/plugin/scenarios/gateway-3.x-scenario

JoeKerouac:
@wu-sheng oh, in this use case, just test `apm-spring-cloud-gateway-2.0.x-plugin`, and it's no problem, but i'm talk aboult `apm-spring-cloud-gateway-3.x-plugin` and `apm-spring-cloud-gateway-2.1.x-plugin`, neither version has been use case;

wu-sheng:
I am not following. In this case, it depends on 3.x 

https://github.com/apache/skywalking-java/blob/6eeacf0959/test/plugin/scenarios/gateway-3.x-scenario/gateway-projectA-scenario/pom.xml#L35

JoeKerouac:
@wu-sheng It should be that the test case does not cover this scenario

wu-sheng:
This is your PR, you should not the one saying `should` or `should not`. As you are contributing a code change, you should verify what is exactly the issue.

wu-sheng:
We can't accept a PR without a clean explanation, as by that, we will face unexpected situations.

JoeKerouac:
@wu-sheng I have verified it on my project, I will write a demo to verify it later;

wu-sheng:
You should check what is missing in the existing test case, through **Plugin automatic test framework**.(https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/plugin-test/)

After that, you need to add a new or adjust existing cases to include your cases in it. Then the next time, when another contributor tries to enhance this plugin again, he would know he didn't break other things. 
That is why we have so many cases in the CI(130+ frameworks testing).

JoeKerouac:
ok, i will

JoeKerouac:
@wu-sheng sorry, my fault, I didn't notice the span created in `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v3x.NettyRoutingFilterInterceptor` is closed in `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v3x.HttpClientFinalizerSendInterceptor`, in fact, The problem is due to a version conflict, at runtime,  Different versions of `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v3x.NettyRoutingFilterInterceptor` and `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v21x.HttpClientFinalizerSendInterceptor` take effect, In the end, the span was not released due to a type conversion exception in `org.apache.skywalking.apm.plugin.spring.cloud.gateway.v21x.HttpClientFinalizerSendInterceptor`;

thank you for your patience;

wallezhang:
> @wallezhang Could you take a look at this?

Sorry, I just saw this discussion. It seems you already have a conclusion now.

wu-sheng:
Oh, you should not put two plugins of gateway in the same time. They didn't have a very strict witness class to disable mismatched versions, as many codes use this class.

PR #292, opened 2022-08-09
Force the injected high-priority classes in order to avoid NoClassDefFoundError.

leihuazhe:
Fix when enhance higher priority classes which is put by hashmap causing NoClassDefFoundError

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@ Release Notes.
 * Upgrade gson to 2.8.9.
 * Upgrade netty-codec-http2 to 4.1.79.Final.
 * Fix race condition causing agent to not reconnect after network error
-* Fix enhance high priority classes which is put by hashmap causing NoClassDefFoundError
+* Force the injected high-priority classes in order to avoid NoClassDefFoundError.
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,11 +15,13 @@ Release Notes.
 * Upgrade gson to 2.8.9.
 * Upgrade netty-codec-http2 to 4.1.79.Final.
 * Fix race condition causing agent to not reconnect after network error
+* Fix enhance high priority classes which is put by hashmap causing NoClassDefFoundError
 
 #### Documentation
 
 * Add a section in `Bootstrap-plugins` doc, introducing HttpURLConnection Plugin compatibility.
-* Update `Plugin automatic test framework` ,fix inconsistent description about configuration.yml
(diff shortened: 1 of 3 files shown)


leihuazhe:
I  change the  skywalking-java-agent  package for our company private using. But, when I launch the agent, here is the phenomenon：
wecom-temp-826069dacbd3a3d8ba1c34b5602ba5ad

The MethodInvocationContext is before MethodInterceptResult, so when load the MethodInvocationContext , will cause  java.lang.NoClassDefFoundError. 

Although the order of high loader classes is not out of order  in the skywalking project.
But I think we should ensure the load sequence should order by the define in the class.

leihuazhe:
image

the MethodInvocationContext is the subclass of MethodInterceptResult.
if classloader load the MethodInvocationContext first, it couldn't find the MethodInterceptResult .so exception below:

java.lang. IllegalStateException Create breakpoint: java.lang.NoClassDefFoundError: com/maple/shine/apm/agent/core/plugin/interceptor/enhance/MethodInterceptResult

this is my customized class.

wu-sheng:
I am not sure this is the case. The class files could be cross-reference. Such as A is using B, and B is using A.

If Byte-buddy is asking for a sorted class map, it should be clear on the method's comment, which I can't find it. Have you checked with byte-buddy author about your case? Because you are using our codebases for private purposes, I can't confirm your change, unless the bytebuddy side could confirm the sequence matters.

leihuazhe:
resonable,I will check the bytebuddy's api, take some expriment.

leihuazhe:
The bytebuddy author's replay:  
_**If there is a dependency, you need to inject the classes in the required order.**_

So maybe we should ensure the sequence when using the ClassInjector.injectRaw

 
https://github.com/raphw/byte-buddy/issues/1299

wu-sheng:
Thanks for the update. You should update changes.md about this fix.

leihuazhe:
finished

CHANGES.md
```diff
@@ -15,6 +15,7 @@ Release Notes.
 * Upgrade gson to 2.8.9.
 * Upgrade netty-codec-http2 to 4.1.79.Final.
 * Fix race condition causing agent to not reconnect after network error
+* Fix enhance high priority classes which is put by hashmap causing NoClassDefFoundError
```
wu-sheng, line 18:
```suggestion
* Force the injected high-priority classes in order to avoid NoClassDefFoundError.
```

wu-sheng (approved):

PR #300, opened 2022-08-31
Upgrade jedis plugin to support 3.x(stream),4.x

pg-yang:

Closes [#9534](https://github.com/apache/skywalking/issues/9534)

I  didn't  create  common module  , Because  there  is  nothing  to  share except  a config  class  . I  enhanced  connection  ,  it has  better  compatibility than enhancing declared method  in `Jedis/Pipeline`


test/plugin/scenarios/jedis-2.x-scenario/support-version.list
```diff
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# 3.0.1 not support stream
+3.0.1
```
wu-sheng, line 18: What does this mean?

wu-sheng: And is 3.0.1 the only 3.x version?

apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/define/AbstractWitnessInstrumentation.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    protected String[] witnessClasses() {
+        return new String[0];
+    }
+
+    @Override
+    protected List<WitnessMethod> witnessMethods() {
+        return Collections.singletonList(new WitnessMethod(
+                "redis.clients.jedis.Pipeline",
```
wu-sheng, line 39: So, `redis.clients.jedis.Client` doesn't exist in v4, but `redis.clients.jedis.Pipeline` exist in v3 or v2?

wu-sheng (commented):
The `supported-list` should be updated.

test/plugin/scenarios/jedis-2.x-scenario/support-version.list
pg-yang: Stream supported since Jedis 3.1.0 , So I put `3.0.1` to `  jedis-2.x-scenario` , This  test-scenario not  include  stream  test  
https://github.com/redis/jedis/releases/tag/jedis-3.1.0

wu-sheng: I see. Thanks for sharing.

apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/define/AbstractWitnessInstrumentation.java
pg-yang: Yes

wu-sheng (approved):

PR #303, opened 2022-09-05
Update guava-cache,jedis,memcached,ehcache plugins to adopt uniform tags

pg-yang:

First , Add db .type  for  guava , encache which  miss the  tag 

related  issue  https://github.com/apache/skywalking/issues/9556


apm-sniffer/optional-plugins/ehcache-2.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/ehcache/v2/EhcacheLockInterceptor.java
```diff
@@ -49,6 +49,7 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
         if (element != null) {
             Tags.DB_STATEMENT.set(span, element.toString());
         }
+        Tags.DB_TYPE.set(span, "Ehcache");
```
wu-sheng, line 52: How about we add a `Tags.CACHE_TYPE`? As we have agreed that, database and cache are different.

pg-yang: Must   modify  all cache  plugin , such  jedis . 

And  `DB_STATEMENT`  ?

pg-yang: I found  the  comment  about  DB _TYPE 
```

    /**
     * DB_TYPE records database type, such as sql, redis, cassandra and so on.
     */
    public static final StringTag DB_TYPE = new StringTag(3, "db.type");

```

wu-sheng: I think we didn't analyze spans from those plugins, so, changing new tags for analyzing is not an issue. I noticed plugins like Jedis using database `statement` and `type`.

Also, how about we add a new `cache.key`(key for reading and writing) and `cache.op`(read/write/general) tags for those plugins too?

pg-yang: We could modify  these  tag , since they haven't  be  used.
We  don't  have  clear way  to  extract cache.key , cache.op .
We can't  guarantee developers must write code  to  provide  correct `op ` ,`key ` at least .
in  order  to get  correct  `op ` ,`key `,Plugins must know java method which is  involved by user , it's  very   difficult ,and  ugly

wu-sheng: What do you mean knowing method? Redis get API always requires a key

pg-yang: -> cache.key 

Some  command  don't  have  key  , or  the  key  can't  be  serialized 

-> cache.op 

For  example  , simulate  java  stack that  call  redis server    , like  this
```
get -> exe 
put -> exe 
```
We  intercept  exe method  usually , in  order  to  get  op . we need  know  the method  which is  invoked   by  user (get /put ) .
Of  course , We  also  could  detect op  from  method  argument  ,  But  sometimes  the  method  don't  have  appropriate  argument .
We  also need  a  relationship for  mapping  something to op  ， I think it need too many work， and it very  depends on developer 's  idea or perception .

pg-yang: We  could  provide  a  mapping  mechanism .
The  plugin only  send  statement  to  OAP , and OAP provide a  mechanism  to  parse  the  statement to op/key

pg-yang: I'm very  understand  that  you want  more  metric . But  I  don't  agree  create fat-plugin

wu-sheng: Yes, it is not easy. I don't mean it has to be done.
As you asked for something to do? I provide as many ideas as possible.
This could cause several PR, several months to do. This happened many times. SkyWalking kernel had been written 4 times
 😁This is hoe oss evaluation. We try to do long term plan step by step.

pg-yang: > We could provide a mapping mechanism in OAP The plugin only send statement to OAP , and OAP could parse the statement to op/key or other flag . So , user could know how it work clearly

How about this . I like

wu-sheng: > I  understand  that  you want  more  metric . But  I  don't  agree  with creating fat-plugin  

I don't want more metrics. SkyWalking costs a lots of energe to reduce resource costs. 
My point is simple, if we cost timr to build a thing, we build a useful tool. 
SkyWalking is not chasing a function on the menu, website or in the news. We try to do a right thing, which is a good APM tool to help ppl addressing perf issues.

OSS community is not like a daily work, we are not urgent.
Many details could be discussed during the process.

SkyWalking existing features are all being discussed many times, they are not done by an easy way for anyone.

wu-sheng: > > We could provide a mapping mechanism in OAP The plugin only send statement to OAP , and OAP could parse the statement to op/key or other flag . So , user could know how it work clearly
> 
> 
> 
> How about this . I like
> 
> BTW , fat-plugin shoudn't exist

If you searched closed issues, you will find a SQL parser PoC, powered by Apache ShardingSphere. It failed due to perf impact, can't analyze SQL fast enough to get table read/write metrics. 

I don't think we could win this for cache. SQL is much better formated than No SQL or cache statements.

pg-yang: I see , I don't work for a company .HaHa .
Le's talk about providing mapping  mechanism in OAP for op/key

pg-yang: > > > We could provide a mapping mechanism in OAP The plugin only send statement to OAP , and OAP could parse the statement to op/key or other flag . So , user could know how it work clearly
> > 
> > 
> > How about this . I like
> > BTW , fat-plugin shoudn't exist
> 
> If you searched closed issues, you will find a SQL parser PoC, powered by Apache ShardingSphere. It failed due to perf impact, can't analyze SQL fast enough to get table read/write metrics.
> 
> I don't think we could win this for cache. SQL is much better formated than No SQL or cache statements.

Yes , it maybe slow .  Let me think about this problem.

pg-yang: At first I wanted to try something   , but  I  always  want  to  finish  it  as  possible  as  soon , even  if no one is urging me .   Sorry  for  my mood .

wu-sheng: At least in SkyWalking, we don't push contributors. You are already self driven, so we can have you as volunteers to build this. It is not fair we asked a volunteer to promise something must be done.

We have plans to release every 2 or 3 months, you could set a plan to finish this in one or two milestones, which means you could have half a year buffer.

So, don't hurry, and don't worry. And more ,Git is a time machine, we could revert halfly done things to keep release work, and revert the revert back to continue.

wu-sheng:
Let's start from listed all relative plugins, which are cache plugins.
Then we could check tags usage and adjust them one by one.
- Make cache tags to replace database tags
- Add tags for keys
- Define op tags(read/write/enqueue/etc.) and implement them in plugins.

pg-yang:
redis commands : https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/Protocol.java#L218
memcached : https://github.com/memcached/memcached/wiki/Commands
redis is the  most complex  ,  I'll  try it  first .

pg-yang:
Look like this  
```java
public class OperatorParser {

    private static final Map<Command, String> MAPPING = new HashMap<>();

    static {
        MAPPING.put(Command.GET, "READ");
        MAPPING.put(Command.HGET, "READ");
        MAPPING.put(Command.MGET, "READ");
        MAPPING.put(Command.ZCARD, "READ");
        MAPPING.put(Command.ZCOUNT, "READ");
        // .... 
        // ....
    }

    public static String parse(Command command) {
        return Optional.ofNullable(command).map(MAPPING::get).orElse("Other");
    }
}

```
Do  you have  better way ?

pg-yang:
We could provide a  tag  named `cache.command`  ,   user can config which command should be  analyzed in OAP .

pg-yang:
https://github.com/pg-yang/skywalking-java/blob/feature/virtual-cache-v2/apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/OperationParser.java

It's  my code  for  converting command  to `op `.

wu-sheng:
To easy the configurations, I think we only need read/write and default(cmd in your configuration?). Then the worst case when we missed something, is that, we don't treat it as a read/write command.

wu-sheng:
> We could provide a tag named `cache.command` , user can config which command should be analyzed in OAP . Or mapping command to op(write/read/etc) , It's simple string match , such as
> 
> ```
> Write:
>   - GET
>   - HGET
> Read:
>  - SET
>  - SETNX
>  - HSET
> ```
> 
> Actually , business application just use a few command in cache system

Yes, we just need READ or WRITE tag if we could identify them.

pg-yang:
How about  we  provide mapping mechanism from  command  to  op  in OAP . we support two  group  (read/write) analysis .
This mapping  could be  configured  by  user  .  User  make that  which  command  is write  or  read

pg-yang:
So, We put the mapping  configure to  plugin side .

pg-yang:
I have done  a  demo  for  plugin .  I  add  a  module named  `virtual-cache-helper`.
Cache plugin must call `HelperInitiator#register` , this method  will return a `Function`  that convert command  to  op

wu-sheng:
Where does this class place? I think this is plugin specific only?

pg-yang:
`HelperInitiator` will  be  put  to  a  new  module named  `virtual-cache-helper` , this  module is  submodule  of `apm-sdk-plugin` , and  it's  responsible for parsing the configuration , providing  uniform convert function 
All  cache plugin  will  depends  this plugin  .

wu-sheng:
Why we need to share? I think the command is tech stack relative. 
Do you want to redis clients sharing?

pg-yang:
> Why we need to share? I think the command is tech stack relative. Do you want to redis clients sharing?
jedis  , guava-cache,  memercached  , all cache plugin need to  depend it

wu-sheng:
Why? Their command should be different, I think we could have their own? It is better we don't set plugin dependencies. This would be more friendly.

pg-yang:
Every  cache plugin  should  have  own  config for   Write/Read  mapping , even  if  their comma is  different . This module only math String command  to operation .
If  we  don't  want  adding  dependency to  cache  plugin , we  could  use   `maven shade plugin `  for  `merging ` common  class  to  every cache plugin .

wu-sheng:
Every plugin commands are different, I think copying the codes into various plugins under different package names is fine.

Jedis has nothing to share with Ehcache commands, right?

wu-sheng:
> https://github.com/pg-yang/skywalking-java/blob/feature/virtual-cache-v2/apm-sniffer/apm-sdk-plugin/virtual-cache-helper/src/main/java/org/apache/skywalking/apm/plugin/vcache/helper/HelperInitiator.java

I didn't read this from my cellphone at this night. I just checked I think you made things too complex. 
As a mapping, I think you just need a map like your [previous version](https://github.com/pg-yang/skywalking-java/blob/feature/virtual-cache-v2/apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/OperationParser.java). Simple, fast and clear

pg-yang:
No problem , Just  need  a  config .

pg-yang:
Update  ,define following  tag for all cache plugin 

1. `cache.type` , represent  which cache system is used 
2. `cache.op` , represent the command for write or  read 
3. `cache.cmd` represent command which  send to cache system 
4. `cache.key`  represent key which  send to cache system , nullable 


Provide  following config  for  all  cache plugin (Java  filed format)

1. `Set<String> OPERATION_MAPPING_WRITE` , which  command should be  mapped to write 
2. `Set<String> OPERATION_MAPPING_READ` ,which  command should be  mapped to read

wu-sheng:
Make sense to me, please go ahead. We need to find a place to doc these tags and meanings in Java agent and backend.

I will back to this thread later.

wu-sheng:
Could you send a pull request to here https://skywalking.apache.org/docs/skywalking-java/latest/en/setup/service-agent/java-agent/java-plugin-development-guide/#advanced-apis (after the whole `Advanced APIs` section)?

We should add reserved tags, the new added tags and database relative tags should be defined here.
Database relative tags could refer from codes and the [backend doc](https://skywalking.apache.org/docs/main/latest/en/setup/service-agent/virtual-database/)

A new backend doc for `virtual-cache` should be added later

pg-yang:
OK.

pg-yang:
Let's finish `virtual-cache` first , then coding ,  we  could agree through the `doc`.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
```diff
@@ -107,6 +107,28 @@ public static final class HTTP {
...
+
+    /**
+     * CACHE_TYPE records the cache command
+     */
+    public static final StringTag CACHE_CMD = new StringTag(17, "cache.cmd");
+
+    /**
+     * CACHE_TYPE records the cache key
+     */
+    public static final StringTag CACHE_KEY = new StringTag(18, "cache.key");
```
wu-sheng, line 131: @apache/skywalking-committers Please notice, these new tags would be significant for virtual cache analysis, which would be added in 9.3.0.

If you want your cache framework relative plugins would work like these Java plugin, please follow the tag format protocol.

pg-yang:
I  think  this PR would not  be  merged util the work in OAP finished , it's better wait I finish the test of  OAP and plugin .

wu-sheng:
I will review the change, I think this PR doesn't have side effects. Users could use it for previous OAP, as we don't analysis anything before.

We could add v8.13.0 Java agent and required tags in the OAP doc.

wu-sheng (approved):
LGTM

pg-yang:
I'm  disheartened  to write doc .  Thanks , You  are   patient

wu-sheng:
> I'm disheartened to write doc . Thanks , You are patient

I would like to help with docs writing if you have any trouble.

PR #304, opened 2022-09-07
Support set/hashset/treeSet/arraylist/linkedlist types in the agent and plugin configurations

pg-yang:

Set is faster than List for detecting value if exist . and could remove  duplicate  value  too

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Release Notes.
 8.13.0
 ------------------
 
+* Support set-type in the agent or plugin configurations
 
 #### Documentation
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -5,6 +5,7 @@ Release Notes.
 8.13.0
 ------------------
 
+* Config support Set parameter
```
wu-sheng, line 8:
```suggestion
* support set-type in the agent or plugin configurations.
```

wu-sheng:
Set is a generic type, I noticed you only support String Set? 
You should read the generic type and convert element type too.

wu-sheng:
Set is a generic type, I noticed you only support String Set? 
You should read the generic type and convert element type too.

pg-yang:
Try to  support Set<String> , String<Integer>  , etc .

wu-sheng:
I mean the codes are converting to set string forcely, no matter what generic type declared.

pg-yang:
Are  you  mean  Java-generics ? like  Set<String> , Set<Integer> ....

wu-sheng:
Yes, based on your codes, I believe you still would add string in to int set.
We should convert configuration to the expected type.

pg-yang:
OK, I'll support all `java.util.Collection` . It's better

pg-yang:
PING

wu-sheng (approved):

PR #305, opened 2022-09-08
Optimize ConfigInitializer warnning mesaage,fix override and merge de…

pg-yang:

 I use System.err for printing warnning message , because there is no log dependence in this module , the warning message such as 
```
The config value will be truncated , because the length max than 1 : agent.service_name -> Your_ApplicationName

```

Closes https://github.com/apache/skywalking/issues/9577

### DIFF
--- a/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
+++ b/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
@@ -52,19 +52,21 @@ public class ConfigInitializer {
             if (Modifier.isPublic(field.getModifiers()) && Modifier.isStatic(field.getModifiers())) {
                 String configKey = (parentDesc + "." + field.getName()).toLowerCase();
                 Class<?> type = field.getType();
-
                 if (Map.class.isAssignableFrom(type)) {
                     /*
                      * Map config format is, config_key[map_key]=map_value, such as plugin.opgroup.resttemplate.rule[abc]=/url/path
-                     * "config_key[]=" will generate a bleak Map , user could use this mechanism to set a blank Map
+                     * "config_key[]=" will generate an empty Map , user could use this mechanism to set an empty Map
                      */
                     ParameterizedType genericType = (ParameterizedType) field.getGenericType();
                     Type[] argumentTypes = genericType.getActualTypeArguments();
                     Type keyType = argumentTypes[0];
                     Type valueType = argumentTypes[1];
-                    // A chance to set a blank map
+                    // A chance to set an empty map
                     if (properties.containsKey(configKey + "[]")) {
-                        field.set(null, initBlankMap(type));
+                        Map currentValue = (Map) field.get(null);
+                        if (currentValue != null && !currentValue.isEmpty()) {
+                            field.set(null, initEmptyMap(type));
+                        }
                     } else {
                         // Set the map from config key and properties
                         Map map = readMapType(type, configKey, properties, keyType, valueType);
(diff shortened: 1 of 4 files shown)


docs/en/setup/service-agent/java-agent/configurations.md
```diff
@@ -114,5 +114,10 @@ This is the properties list supported in `agent/config/agent.config`.
...
 | `plugin.micronauthttpserver.collect_http_params`                | This config item controls that whether the Micronaut http server plugin should collect the parameters of the request. Also, activate implicitly in the profiled trace.                                                                                                                                                                                                                                                                                                                                                                                 | SW_PLUGIN_MICRONAUTHTTPSERVER_COLLECT_HTTP_PARAMS                | `false`                                                                 |
 
+# How to override default value to blank Collection/Map
+
+* Collection type config such as : `plugin.kafka.topics=a,b,c,d` , In order set the value to blank . just omit the value:` plugin.kafka.topics=` 
+* Map type config such as : `plugin.kafka.producer_config[key]=X` ，In order set the value to blank , just omit the key and value :` plugin.kafka.producer_config[]=`
```
wu-sheng, line 120:
```suggestion
# Reset Collection/Map type configurations as empty collection.

* Collection type config, e.g.  using ` plugin.kafka.topics=` 
to override default `plugin.kafka.topics=a,b,c,d` 
* Map type config, e.g. using `plugin.kafka.producer_config[]=` to override default `plugin.kafka.producer_config[key]=value`
```

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
```diff
@@ -50,43 +53,45 @@ private static void initNextLevel(Properties properties, Class<?> recentConfigTy
                 String configKey = (parentDesc + "." + field.getName()).toLowerCase();
                 Class<?> type = field.getType();
 
-                if (type.equals(Map.class)) {
+                if (Map.class.isAssignableFrom(type)) {
                     /*
-                     * Map config format is, config_key[map_key]=map_value
-                     * Such as plugin.opgroup.resttemplate.rule[abc]=/url/path
+                     * Map config format is, config_key[map_key]=map_value, such as plugin.opgroup.resttemplate.rule[abc]=/url/path
+                     * "config_key[]=" will generate a bleak Map , user could use this mechanism to set a blank Map
```
wu-sheng, line 59: Like the docs I changed, there is no blank map/list. It is an empty map/list.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
```diff
@@ -50,43 +53,45 @@ private static void initNextLevel(Properties properties, Class<?> recentConfigTy
...
-                    Type valueType = null;
-                    if (argumentTypes != null && argumentTypes.length == 2) {
-                        // Get key type and value type of the map
-                        keyType = argumentTypes[0];
-                        valueType = argumentTypes[1];
+                    Type keyType = argumentTypes[0];
+                    Type valueType = argumentTypes[1];
+                    // A chance to set a blank map
+                    if (properties.containsKey(configKey + "[]")) {
+                        field.set(null, initBlankMap(type));
```
wu-sheng, line 67: If it has been initialized as empty, you should break and jump out of the loop. Otherwise, you would provide a sequential re-init

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
```diff
@@ -50,43 +53,45 @@ private static void initNextLevel(Properties properties, Class<?> recentConfigTy
...
                     }
-                    Map map = (Map) field.get(null);
-                    // Set the map from config key and properties
-                    setForMapType(configKey, map, properties, keyType, valueType);
-                } else if (Collection.class.isAssignableFrom(type)) {
+                    continue;
+                }
+                if (!properties.containsKey(configKey)) {
+                    continue;
+                }
```
wu-sheng, line 79: I think it is better if you could use `else if` here.

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
```diff
@@ -50,43 +53,45 @@ private static void initNextLevel(Properties properties, Class<?> recentConfigTy
...
+                //In order to guarantee user could override the default value to blank , we parse the value even if it's blank
```
wu-sheng, line 80: I don't think `override xxx to` is correct grammar. You could say `use xxx to override` or `xx is reset as yyy`

pg-yang: In order to guarantee the default value could be reset as empty/blank , we parse the value even if it's blank

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
```diff
@@ -52,50 +52,50 @@ private static void initNextLevel(Properties properties, Class<?> recentConfigTy
...
                     Type[] argumentTypes = genericType.getActualTypeArguments();
                     Type keyType = argumentTypes[0];
                     Type valueType = argumentTypes[1];
-                    // A chance to set a blank map
+                    // A chance to set an empty map
                     if (properties.containsKey(configKey + "[]")) {
-                        field.set(null, initBlankMap(type));
+                        Map currentValue = (Map) field.get(null);
+                        if (currentValue != null && !currentValue.isEmpty()) {
+                            field.set(null, initBlankMap(type));
```
wu-sheng, line 68: This method is still not changed.

pg-yang: Sorry , I don't  understand .
If  the default value of Map isn't empty , I should  create  an empty Map .

wu-sheng: `initBlankMap`->`initEmptyMap`. A map can't be blank, a string can.

pg-yang: Sorry .

wu-sheng (dismissed):

wu-sheng (approved):

PR #309, opened 2022-09-14
[Bug #9598] Apache ShenYu agent Tid is null

lishuo5263:
### Fix <Apache ShenYu agent Tid is null #9598>

When the implementation class of GlobalPluginExecuteMethodInterceptor#beforeMethod, The traceId result perhaps return empty string value, '.orElse("")' method should be set global traceId to default value.

Fix https://github.com/apache/skywalking/issues/9598

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Release Notes.
 * Support to set the value of Map/List field to an empty map/list.
 * Add plugin to support [Impala JDBC](https://www.cloudera.com/downloads/connectors/impala/jdbc/2-6-29.html) 2.6.x.
 * Update guava-cache, jedis, memcached, ehcache plugins to adopt uniform tags.
-* Fix Apache ShenYu plugin traceId empty string value. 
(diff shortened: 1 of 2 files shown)


wu-sheng:
To be reviewed by Shenyu contributor @hutaishi

CHANGES.md
```diff
@@ -11,7 +11,7 @@ Release Notes.
 * Support to set the value of Map/List field to an empty map/list.
 * Add plugin to support [Impala JDBC](https://www.cloudera.com/downloads/connectors/impala/jdbc/2-6-29.html) 2.6.x.
 * Update guava-cache, jedis, memcached, ehcache plugins to adopt uniform tags.
-
+* Fix `Apache ShenYu` plugin traceId empty string value.
```
wu-sheng, line 14:
```suggestion
* Fix `Apache ShenYu` plugin traceId empty string value. 

```

wu-sheng (approved):
As the contributor is from Shenyu committer team(ref https://shenyu.apache.org/team), I would give this PR a pass.

PR #347, opened 2022-10-13
Transport maybe is not a EnhancedInstance.

xiaolongzuo:
### Fix <Transport maybe is not a EnhancedInstance. For example, a TFrameTransport encapsulating TSocket.>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Release Notes.
 * Update guava-cache, jedis, memcached, ehcache plugins to adopt uniform tags.
 * Fix `Apache ShenYu` plugin traceId empty string value. 
 * Add plugin to support [brpc-java-3.x](https://github.com/baidu/starlight/tree/brpc-java-v3)
+* Fix the problem of abnormal type conversion when the thrift plug-in intercepts TServiceClient.
 
 #### Documentation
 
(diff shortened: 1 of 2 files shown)


wu-sheng:
You should use `instanceof` rather than a `catch`.

wu-sheng:
> Fix <Transport maybe is not a EnhancedInstance. For example, a TFrameTransport encapsulating TSocket.>

So, in `TFrameTransport` case, `UNKNOWN` is good enough?

wu-sheng:
Are you going to explain what you are proposing to change?

wu-sheng:
This PR has no update in half month. I have to close this.
If you want to continue on this, let's know.

xiaolongzuo:
> > Fix <Transport maybe is not a EnhancedInstance. For example, a TFrameTransport encapsulating TSocket.>
> 
> So, in `TFrameTransport` case, `UNKNOWN` is good enough?

The best way is to enumerate all the situations and take out the correct peers for these situations. However, I'm not sure of any other situations. TFrameTransport is just one of the ones I encounter in my own use.

wu-sheng:
If you still want to finish this PR, please follow the comments of the review.

About the UNKNOWN, I just want to make sure which field is going to be set as UNKNOW and how it looks like on the UI?

xiaolongzuo:
> If you still want to finish this PR, please follow the comments of the review.

1、Which field is going to be set as UNKNOW？ 
A：Peer of exit span, likes below.
image
2、How it looks like on the UI?
A：I think it should be UNKNOWN in the peer, but I haven't tested this.

wu-sheng:
Peer as UNKNOWN? Then the node of topology would be unknown, this seems not right.

As on the client side, you should be able to get remote service logic name, or IP/DNS address, right?

PR #348, opened 2022-10-16
Update `compose-start-script.template` to make compatible with new version docker compose

pg-yang:

The latest docker compose use `-` as separator rather than `_`when  generating container name . 
Use `COMPOSE_COMPATIBILITY` env  to keep compatibility  

Refer  to  https://github.com/docker/compose/commit/150fd4b8cfa39726e68c12fc0fd45ac4e0f20c5f

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Release Notes.
 * Update guava-cache, jedis, memcached, ehcache plugins to adopt uniform tags.
 * Fix `Apache ShenYu` plugin traceId empty string value. 
 * Add plugin to support [brpc-java-3.x](https://github.com/baidu/starlight/tree/brpc-java-v3)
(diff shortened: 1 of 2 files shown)


kezhenxu94:
I would suggest modifying infra e2e to adapt this.

pg-yang:
Maybe  the wrong title mislead you ?
Actually ,  it's the  problem about java agent plugins test  scenario . test  scenario use  shell to involve `docker  compose ` .  
Then  invoke `docker ps -qf "name=${container_name}`(the  value of $container_name use  `_` as   separator) to get container id , Because  the  latest docker compose use `-` as separator , so the shell couldn't  find any container

kezhenxu94:
> Maybe  the wrong title mislead you ?
> 

Yes. Plugin tests are different from e2e tests. Please also update the changelog.

kezhenxu94 (approved):

PR #349, opened 2022-10-17
Fix dead link

kezhenxu94:


wu-sheng (approved):

kezhenxu94:
https://kotlinlang.org is using cookie and if there is no cookie it just redirect to itself, browsers have cookie and can display this correct but the link checker doesn't support cookie so we have to ignore links from https://kotlinlang.org for now

wu-sheng (approved):

PR #362, opened 2022-10-21
Polish up rabbitmq-5.x plugin to fix missing broker tag on consumer side

pg-yang:

There is an issue about `NotNullAssertor` in agent  plugin test when checking tag with null value 

https://github.com/apache/skywalking-java/blob/6d5bbe8e5f7edec0cb10e3772266b6f632a04a34/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/util/TagValuePair.java#L41-L46
Like  above code , if the value of tag is null , we wouldn't set anything to value , and the value would be a blank string, , The   blank string is initiated by `KeyStringValuePair.Builder` which compiled by  protobuf . So the  result of  `NotNullAssertor` alway is true .

This is  `KeyStringValuePair.Builder` : 

image

This is  why the value of   tag `mq.broker`  is  null   ,but  the test sill passed 

https://github.com/apache/skywalking-java/blob/6d5bbe8e5f7edec0cb10e3772266b6f632a04a34/test/plugin/scenarios/rabbitmq-scenario/config/expectedData.yaml#L33


wu-sheng (approved):
LGTM, wait for plugin tests to verify this doesn't break existing cases.

pg-yang:
@wu-sheng  Did you look the above comment ?  
In a word  ,  `NotNullAssertor `  maybe not accurate  with tag which value is null .
We  could  provide  `NotBlankAssertor` in  test  tool project  , or  set the value to  keyValueBuilder even if it's null

wu-sheng:
> We could provide NotBlankAssertor in test tool project , or set the value to keyValueBuilder even if it's null

Usually, we don't suggest post tag with empty value, so, we only have `not null` for now. If you feel necessary, `not blank` should be fine, but only for string field or tag.

pg-yang:
Yes , I  think it's  necessary  , I previously confirmed  virtual mq related tag through   `expectedData.yaml` in  test case , and it  seems not accuracy for now  . 
Please wait until I finish `not blank` . It  could help us to check whether others mq plugin's  tag is right ?

wu-sheng:
Make sense.

pg-yang:
OK . Got  it

wu-sheng:
So, I could merge this?

PR #363, opened 2022-10-23
Polish up activemq plugin to fix missing broker tag on consumer side

pg-yang:


CHANGES.md
```diff
@@ -19,6 +19,7 @@ Release Notes.
 * Correct the duration of the transaction span for Neo4J 4.x.
 * Plugin-test configuration.yml dependencies support docker service command field
 * Polish up rabbitmq-5.x plugin to fix missing broker tag on consumer side
+* Polish up activemq plugin to fix missing broker tag on consumer side
```
wu-sheng, line 22:
```suggestion
* Polish up activemq plugin to fix missing broker tag on consumer side
* Enhance MQ plugin relative tests to check key tags not blank.
```

I think we should add this in the change log?

wu-sheng (approved):

PR #366, opened 2022-10-25
dcl+volatile

Alexhuangqing:


wu-sheng:
Could you explain why you need this? Are you really facing any case about concurrency case for this default class loader?

Alexhuangqing:
> Could you explain why you need this? Are you really facing any case about concurrency case for this default class loader?

no ,I just think by my habit

wu-sheng:
This should be initialized before all codes, and no concurrent case.
We never faced issues in real scenarios.

PR #371, opened 2022-11-02
Support mannual propagation of tracing context to next stream for webflux.

wuwen5:
Currently, the webflux plug-in does not support tracking downstream component calls.

e.g.
```java
@GetMapping("/testcase/annotation/mono/onnext")
    public Mono<String> monoOnNext(@RequestBody(required = false) String body) {
        return Mono.subscriberContext()
                .flatMap(ctx -> {
                    visit("http://localhost:" + serverPort + "/testcase/success");
                    //e.g. redis、dubbo and other components calls cannot be traced
                    return Mono.just("Hello World");
                });
    }
```

Related issue: https://github.com/apache/skywalking/issues/8644

This implementation has learned from `spring-cloud-sleuth`'s manual mode (`WebFluxSleuthOperators`)


apm-sniffer/apm-toolkit-activation/apm-toolkit-kafka-activation/src/main/resources/skywalking-plugin.def
```diff
@@ -14,4 +14,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-toolkit-kafka=org.apache.skywalking.apm.toolkit.activation.kafka.define.KafkaOnMessageAnnotationInstrumentation
\ No newline at end of file
+toolkit-kafka=org.apache.skywalking.apm.toolkit.activation.webflux.WebFluxSkyWalkingOperatorsActivation
```
wu-sheng, line 17: I think this should not be changed.

apm-application-toolkit/apm-toolkit-webflux/src/main/java/org/apache/skywalking/apm/toolkit/webflux/WebFluxSkyWalkingOperators.java
```diff
@@ -0,0 +1,120 @@
...
+        } catch (Exception e) {
+            throw new IllegalStateException(e);
+        } 
+    }
+
+    public static <T> T withSpanInScope(ServerWebExchange serverWebExchange, Callable<T> callable) {
+        try {
+            return callable.call();
+        } catch (Exception e) {
+            throw new IllegalStateException(e);
```
wu-sheng, line 113: Why do we through an IllegalStateException from any caught exception? This seems wrong.

wu-sheng:
A doc should be added under `Manual APIs` menu. 

Also, as you mentioned, I noticed Kafka API, which doesn't have a document too. Could you also add a doc for it (including a change log)? See the original PR https://github.com/apache/skywalking/pull/5304

wu-sheng:
And please fix CI.

apm-application-toolkit/apm-toolkit-webflux/src/main/java/org/apache/skywalking/apm/toolkit/webflux/WebFluxSkyWalkingOperators.java
wuwen5: If throwing directly will cause trouble for the use of lambda, I have changed to the `sneakyThrow `

wu-sheng: I mean, you don't need to catch? As a wrapper, it should declare the same `throw` as the original.

docs/menu.yml
```diff
@@ -46,6 +46,10 @@ catalog:
             path: "/en/setup/service-agent/java-agent/application-toolkit-trace-cross-thread"
           - name: "MicroMeter Registry"
             path: "/en/setup/service-agent/java-agent/application-toolkit-micrometer"
+          - name: "Webflux"
+            path: "/en/setup/service-agent/java-agent/application-toolkit-webflux"
+          - name: "Kafka"
+            path: "/en/setup/service-agent/java-agent/application-toolkit-kafka"
```
wu-sheng, line 52:
```suggestion
          - name: "Webflux Tracing Assistant APIs"
            path: "/en/setup/service-agent/java-agent/application-toolkit-webflux"
          - name: "Kafka Tracing Assistant APIs"
            path: "/en/setup/service-agent/java-agent/application-toolkit-kafka"  
```

wu-sheng (dismissed):

CHANGES.md
```diff
@@ -24,13 +24,15 @@ Release Notes.
...
 * Add RocketMQ test scenarios for version 4.3 - 4.9. No 4.0 - 4.2 release images for testing.
+* Support mannual propagation of tracing context to next operators for webflux.
 
 #### Documentation
 
 * Update `configuration` doc about overriding default value as empty map/list accordingly.
 * Update plugin dev tags for cache relative tags.
 * Add plugin dev docs for virtual database tags.
 * Add plugin dev docs for virtual MQ tags.
+* Add doc about kafka plugin Manual APIs.
```
wu-sheng, line 35:
```suggestion
* Add doc about kafka plugin Manual APIs.

```

wu-sheng (dismissed):

wu-sheng:
One new plugin is being added, please update plugin-list.md

test/plugin/scenarios/webflux-scenario/webflux-projectB-scenario/src/main/java/test/apache/skywalking/apm/testcase/sc/webflux/projectB/controller/TestAnnotationController.java
```diff
@@ -59,4 +66,18 @@ public Mono<String> urlPattern(@PathVariable("test") String var) {
...
+
+    @RequestMapping("/testcase/success")
+    public String downstreamMock() {
+        return "1";
+    }
+
+    @GetMapping("/testcase/annotation/mono/onnext")
+    public Mono<String> monoOnNext(@RequestBody(required = false) String body) {
+        return Mono.subscriberContext()
+                .flatMap(ctx -> WebFluxSkyWalkingOperators.withSpanInScope(ctx, () -> {
```
wu-sheng, line 78: One question is `WebFluxSkyWalkingOperators#withSpanInScope` being called in the thread having to trace context, capture the snapshot, and continue this snapshot in the callback of another thread, right?

If I got this scenario correctly, I would recommend this API be renamed to `WebFluxSkyWalkingOperators#continueTracing`.

wu-sheng (approved):
LGTM

PR #375, opened 2022-11-05
Polish up pulsar plugins to remove unnecessary dynamic value , set peer at consumer side 

pg-yang:

1. Set peer at consumer side 
2. remove unnecessary dynamic value , instead to get field through consumer/producer(this) object  .

In  pulsar `serviceUrl` means a cluster , and  pulsar support  ` cluster-level failover` , so It's inappropriate to hold `serviceUrl` as  it might be updated , Refer to  https://pulsar.apache.org/docs/client-libraries-java . 

I'm  sure that my  code format style is correct .I  formatted a little code follow current style  ,   Thanks


wu-sheng:
@codelipenghui Could you recheck this? We are formatting the peer and MQ tags for this new dashboard. 

https://skywalking.apache.org/docs/main/next/en/setup/service-agent/virtual-mq/

wu-sheng:
> In pulsar serviceUrl means a cluster , and pulsar support cluster-level failover , so It's inappropriate to hold serviceUrl as it might be updated , Refer to https://pulsar.apache.org/docs/client-libraries-java .

As it is updated in time, so, we may have different cluster addresses when Pulsar cluster scales out

pg-yang:
I think `cluster-level failover ` is not   scales out . 
For  `cluster-level failover ` , we organize each cluster to a service , it's clear .
For  a cluster , It's inevitable if the cluster changed  whether  automatically or manually .

wu-sheng (approved):

PR #380, opened 2022-11-10
correct toolkit naming

BFergerson:
`toolit` -> `toolkit`


wu-sheng (dismissed):

wu-sheng (approved):
You forgot to update the changelog. Did for you

PR #381, opened 2022-11-12
Rename`JDBI` to `JDBC`

nisiyong:
Many spans named `JDBI` are confusing and I think we should correct that typo.

### DIFF
--- a/.github/workflows/plugins-test.3.yaml
+++ b/.github/workflows/plugins-test.3.yaml
@@ -121,7 +121,7 @@ jobs:
         run: |
           mkdir -p skywalking-agent/plugins
           curl -O https://skyapm.github.io/ci-assist/jars/ojdbc14-10.2.0.4.0.jar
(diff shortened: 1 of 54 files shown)


wu-sheng:
This is a legacy thing, rather than a typo.
We were following Open Tracing tag, which declare this. I think it is fine to correct them, but you need to make sure all tests passed

nisiyong:
> This is a legacy thing, rather than a typo.

Thanks for your mention, and I have found that [JDBI](https://jdbi.org/) meaning. `JDBC` still will be easier to understand. All tests are passed, please take a look.

wu-sheng:
Oracle case is still failing.

nisiyong:
I checked the test case failed message, and I think this error is not related to this PR. Could you rerun the `oracle-scenario`?
image

wu-sheng:
I had re-runed twice. It keeps failure.

nisiyong:
I think I should create a PR at repo [SkyAPM/java-plugin-extensions](https://github.com/SkyAPM/java-plugin-extensions) first because the oracle plugin is out of the official repo.

wu-sheng:
A new release of the Oracle plugin repo

https://github.com/SkyAPM/java-plugin-extensions/releases/tag/2.3.0

nisiyong:
The `oracle-scenario` test case runs successfully now.

wu-sheng (approved):

PR #387, opened 2022-11-17
Upgrade agent-test-tools and polish dubbo plugin

nisiyong:
* Upgrade skywalking-agent-test-tools
* Update docs and fix wrong spelling
* Relate https://github.com/apache/skywalking-java/pull/382
* Relate https://github.com/apache/skywalking-agent-test-tool/pull/47

### DIFF
--- a/test/plugin/agent-test-tools/pom.xml
+++ b/test/plugin/agent-test-tools/pom.xml
@@ -35,7 +35,7 @@
     <packaging>pom</packaging>
 
     <properties>
-        <agent-test-tools.version>b2c68c3274a19d40056f109f8145109225c697fa</agent-test-tools.version>
+        <agent-test-tools.version>8a09b6f16b0b3fb535ca515437ef107042aaae45</agent-test-tools.version>
         <agent-test-tools.workingDirectory>${project.basedir}/target/agent-test-tools</agent-test-tools.workingDirectory>
         <agent-test-tools.repos>https://github.com/nisiyong/skywalking-agent-test-tool.git</agent-test-tools.repos>
     </properties>
--- a/test/plugin/agent-test-tools/pom.xml
+++ b/test/plugin/agent-test-tools/pom.xml
@@ -35,7 +35,7 @@
     <packaging>pom</packaging>
 
     <properties>
-        <agent-test-tools.version>7f20775e0631356c4823d9372b09d653db0e6540</agent-test-tools.version>
+        <agent-test-tools.version>2900625c6af6379fa16543e42478974de608f662</agent-test-tools.version>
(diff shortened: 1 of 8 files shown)


wu-sheng:
@nisiyong It seems the tool change breaks.

nisiyong:
> @nisiyong It seems the tool change breaks.

I will take a look.

test/plugin/agent-test-tools/pom.xml
```diff
@@ -35,9 +35,9 @@
     <packaging>pom</packaging>
 
     <properties>
-        <agent-test-tools.version>7f20775e0631356c4823d9372b09d653db0e6540</agent-test-tools.version>
+        <agent-test-tools.version>8a09b6f16b0b3fb535ca515437ef107042aaae45</agent-test-tools.version>
         <agent-test-tools.workingDirectory>${project.basedir}/target/agent-test-tools</agent-test-tools.workingDirectory>
-        <agent-test-tools.repos>https://github.com/apache/skywalking-agent-test-tool.git</agent-test-tools.repos>
+        <agent-test-tools.repos>https://github.com/nisiyong/skywalking-agent-test-tool.git</agent-test-tools.repos>
```
nisiyong, line 40: @wu-sheng After https://github.com/apache/skywalking-agent-test-tool/pull/49 merge, I will change the repo URL.

nisiyong:
Finally, the agent-test-tools has fixed the prefix bug. And  at the same time, I found  `start with` or `not blank` couldn't use at the MeterItems tag value, because the MeterId will compare with the MeterItems tag value.

[MeterId.java](https://github.com/apache/skywalking-agent-test-tool/blob/master/validator/src/main/java/org/apache/skywalking/plugin/test/agent/tool/validator/entity/MeterId.java)

```java
public class MeterId {
    ...
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        MeterId meterId = (MeterId) o;
        return Objects.equals(name, meterId.name) &&
            Objects.equals(tags, meterId.tags);
    }
}
```

So I am not going to change the dubbo thread pool name, but the agent-test-tools still support the new operators `start with` and `end with` in other cases.

wu-sheng:
I am not very clear why `eq` works, but `start with` not?

nisiyong:
> I am not very clear why `eq` works, but `start with` not?

All the operator could work well at the expected value. But in the meter scenario, the test-tools will find the same meter first, then compare the meter value. The same meters must have the same tag-keys and the same tag-value, that's why `MeterId#equals` compare with tags and `MeterId.Tag#equals` compared with tag-key and tag-value.

nisiyong:
The agent-test -tools repo URL and revision has be updated. All things are done, @wu-sheng please take a look.

wu-sheng (approved):

PR #404, opened 2022-12-02
Add tags `mq.message.keys` and `mq.message.tags` for RocketMQ producer span

nisiyong:
RocketMQ message has `tags` and `keys` concept. 
See https://rocketmq.apache.org/docs/4.x/producer/04concept1#message

* Add span tag `message.tags` to help developers identify messages.
* Add span tag `message.keys` to help developers search messages by key in the RocketMQ console.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/rocketMQ-4.x-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/rocketMQ-4.x-plugin/pom.xml
@@ -30,7 +30,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <rocketmq-client.version>4.3.1</rocketmq-client.version>
+        <rocketmq-client.version>4.1.0-incubating</rocketmq-client.version>
     </properties>
 
     <dependencies>
--- a/apm-sniffer/apm-sdk-plugin/rocketMQ-4.x-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/rocketMQ-4.x-plugin/pom.xml
@@ -30,7 +30,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <rocketmq-client.version>4.1.0-incubating</rocketmq-client.version>
+        <rocketmq-client.version>4.3.1</rocketmq-client.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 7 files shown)


wu-sheng:
Isn't keys enough to locate the message? The tracing should take as less messages as possible

nisiyong:
> Isn't keys enough to locate the message? 

The `keys` is indeed enough to locate the message,  which is convenient for locating the problem of message loss.
The `tags` is used to distinguish interrelated messages under the same topic, it is convenient for identifying what type of messages are produced, and the consumer side will filter message with tags, which is useful to help developers know the subscription between provider and consumer, not only topic name. For more explanations see: https://rocketmq.apache.org/docs/4.x/producer/04concept1/#when-to-use-topictag

> The tracing should take as less messages as possible

It should follow this principle, now these tags will be ignored when they are blank.

apm-sniffer/apm-sdk-plugin/rocketMQ-4.x-plugin/pom.xml
```diff
@@ -30,7 +30,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <rocketmq-client.version>4.1.0-incubating</rocketmq-client.version>
+        <rocketmq-client.version>4.3.1</rocketmq-client.version>
```
wu-sheng, line 33: Why bump up this version? Does this mean you break the support list?

nisiyong: It could be reverted, just want to use the release version, rather than the `incubation` version.

nisiyong: https://mvnrepository.com/artifact/org.apache.rocketmq/rocketmq-client

apm-sniffer/apm-sdk-plugin/rocketMQ-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/v4/MessageSendInterceptor.java
```diff
@@ -59,6 +59,15 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
...
         Tags.MQ_BROKER.set(span, (String) allArguments[0]);
         Tags.MQ_TOPIC.set(span, message.getTopic());
+        String keys = message.getKeys();
+        if (StringUtil.isNotBlank(keys)) {
+            span.tag(Tags.ofKey("message.keys"), keys);
+        }
+        String tags = message.getTags();
+        if (StringUtil.isNotBlank(tags)) {
+            span.tag(Tags.ofKey("message.tags"), tags);
+        }
```
wu-sheng, line 69: please add a prefix for tag keys. `mq.message.keys` and `mq.message.tags`. If they are widely known concepts in the queue ecosystem, consider moving them to `Tags`. If not, you don't have to.

nisiyong: Span tags' names have been changed. Keeping RocketMQ's own concepts under itself plugin will be better.

wu-sheng (approved):

PR #405, opened 2022-12-07
[fix][plugin][pulsar] Clean message context after message is recycled to prevent memory leak.

tjiuming:
Currently, we set `new MessageEnhanceRequiredInfo()` into `MessageImpl` to store the tracing context.
But Pulsar has a Recycle mechanism, see: https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java#L126
The `MessageImpl` will not GC by JVM, so the `new MessageEnhanceRequiredInfo()` settled into `MessageImpl` won't be GC too, it will lead to memory leak.
The PR will clean the `new MessageEnhanceRequiredInfo()` settled into `MessageImpl` to prevent memory leak.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/MessageRecycleMethodInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/MessageRecycleMethodInterceptor.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
 package org.apache.skywalking.apm.plugin.pulsar.common;
 
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
--- a/apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/MessageRecycleMethodInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/MessageRecycleMethodInterceptor.java
@@ -1,0 +1,25 @@
+package org.apache.skywalking.apm.plugin.pulsar.common;
+
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+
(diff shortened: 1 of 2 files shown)


apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/MessageRecycleMethodInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class MessageRecycleMethodInterceptor implements InstanceMethodsAroundInterceptor {
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+        // NOOP
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        objInst.setSkyWalkingDynamicField(null);
```
wu-sheng, line 35: I am feeling `reset` in the before is better.

https://github.com/apache/pulsar/blob/68ca60cea7665e6a4bd9f07518b038c17893fe02/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MessageImpl.java#L684

I can see the `recycle` had been done in the method. Is reset this field is good after that?

wu-sheng:
> The MessageImpl will not GC by JVM, so the new MessageEnhanceRequiredInfo() settled into MessageImpl won't be GC too, it will lead to memory leak.

If you mean `MessageImpl` objects are pooled, then the current codes could lead to dirty data. But leak? I doubt that.

@tisonkun @codelipenghui Could any of you confirm this?

apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/MessageRecycleMethodInterceptor.java
tjiuming: it makes sense, reset in the before is better. reset in the after may lead to concurrent problems.

tjiuming:
> > The MessageImpl will not GC by JVM, so the new MessageEnhanceRequiredInfo() settled into MessageImpl won't be GC too, it will lead to memory leak.
> 
> If you mean `MessageImpl` objects are pooled, then the current codes could lead to dirty data. But leak? I doubt that.

in my opinion, I just feeling that release the tracing context as soon as the message recycled will be helpful to saving the heap memory. 
My words are wrong, it's not memory leak, because the MessageImpl is reachable and usable.

wu-sheng:
> we enhance ConsumerImpl#messageProccessed to generate Consumer/Receive event, but usually, the duration of executing ConsumerImpl#messageProccessed is veeeery short.
uses are using Consumer#receive()/receive(long,TimeUnit)/receiveAsync()/batchReceive()/batchReceiveAsync() cannot get the actual duration

About how to get the duration accurate, I think you are better than me. Pulsar plugin mostly was added in 2019, and there was no `async`-span concept back then. 
Now, if you want to make a span duration accurate, we could open the span, and declare it is async and finish it in the callback or another thread when you need. The tracing context would be held by the agent kernel until you finish the span async. 

Ref doc, https://skywalking.apache.org/docs/skywalking-java/next/en/setup/service-agent/java-agent/java-plugin-development-guide/#advanced-apis

wu-sheng (approved):

tjiuming:
There is one point I need to explain:
I searched the `new MessageEnhanceRequiredInfo()` settled into MessageImpl usage,  in `PulsarConsumerInterceptor`, it checked if the result of `(MessageEnhanceRequiredInfo) msg
                .getSkyWalkingDynamicField()` is null. see: https://github.com/apache/skywalking-java/blob/main/apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/PulsarConsumerInterceptor.java#L87
it won't lead to NPE

wu-sheng:
This field is initialized as null. Basically, this should be fine.

wu-sheng:
@tjiuming Oops, you seem missed the changes.md update. Please submit another PR to add one.

tjiuming:
> @tjiuming Oops, you seem missed the changes.md update. Please submit another PR to add one.

OK

PR #427, opened 2022-12-22
[fix] In the higher version of mysql-connector-java 8x, there is an error in the value of db.instance.

lvxiao1:
Starting from MySQL 8.0.20, the `databaseTerm` feature allows you to use the `setCatalog` and `setSchema` methods to switch databases

The code is from 8.0.20 `ConnectionImpl`
image

image

- [ ]  In the higher version of `mysql-connector-java`8x, there is an error in the value of db.instance.

PR #428, opened 2022-12-22
Support to customize the collect period of JVM relative metrics.

dylanforest:
### Support to customize the collect period of JVM relative metrics.
- [] If this is non-trivial feature, paste the links/URLs to the design doc.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/10179.


nisiyong:
It seems the OAP side will calculate GC time per second. This change should consider more.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -273,6 +273,10 @@ public static class Jvm {
          * The buffer size of collected JVM info.
          */
         public static int BUFFER_SIZE = 60 * 10;
+        /**
+         * The period in seconds of JVM metrics collection.
+         */
+        public static int METRICS_COLLECT_PERIOD = 10;
```
wu-sheng, line 279: You should not change the default value.

wu-sheng: It is rarely to see thousands threads for a health service.

apm-sniffer/config/agent.config
```diff
@@ -150,6 +150,8 @@ correlation.value_max_length=${SW_CORRELATION_VALUE_MAX_LENGTH:128}
 correlation.auto_tag_keys=${SW_CORRELATION_AUTO_TAG_KEYS:}
 # The buffer size of collected JVM info.
 jvm.buffer_size=${SW_JVM_BUFFER_SIZE:600}
+# The buffer size of collected JVM info.
+jvm.metrics_collect_period=${SW_JVM_METRICS_COLLECT_PERIOD:10}
```
wu-sheng, line 154: configurations.md should be updated too.

wu-sheng:
> It seems the OAP side will calculate GC time per second. This change should consider more.

Could you list which metrics?

dylanforest:
> It seems the OAP side will calculate GC time per second. This change should consider more.

I don't see any necessary gc metrics calculation for every second. 
The OAP use sum() method in OAL to aggregate gc metrcis, including time and count of different gc phase.
After that, it will generat the final gc metrics of 1min.

On the other hand, Agent record lastest time or count of diff gc phase, and use them to create new metrics in next time collection.
So, I think it's fine if anyone want to change their JVM metrics collect period

dylanforest:
> You should not change the default value.

My bad

dylanforest:
> You should not change the default value.

The default value stays and the configuration.md is updated.
Excuse me for my first pr. ＾-＾

wu-sheng:
Let's wait for @nisiyong feedback.  And the CI would fail,

https://github.com/apache/skywalking-java/pull/427#issuecomment-1362516834

I need to wait for both things.

nisiyong:
I may have just got COVID-19, I will check it when I get better😿

wu-sheng:
> I may have just got COVID-19, I will check it when I get better😿

Take care. I am suffering symptoms too. 4 days fever, but it us going down.

dylanforest:
Tack care guys.
I'm lucky so far, but we really out of weds here.

CHANGES.md
```diff
@@ -20,7 +20,8 @@ Release Notes.
 * Add tags `mq.message.keys` and `mq.message.tags` for RocketMQ producer span
 * Clean the trace context which injected into Pulsar MessageImpl after the instance recycled
 * Fix In the higher version of mysql-connector-java 8x, there is an error in the value of db.instance.
-* Add support for KafkaClients 3.2.x
+* Add support for KafkaClients 3.x.
```
wu-sheng, line 23: @lujiajing1126 I made this. As #429 missed to correct versions.

lujiajing1126: Thx, I forgot to fix it there

nisiyong:
> It seems the OAP side will calculate GC time per second. This change should consider more.

@dylanforest @wu-sheng Sorry for my late reply, this PR looks good to me. 

And I should correct what I said before, the agent side will calculate the GC time rather than the OAP side.

https://github.com/apache/skywalking-java/blob/cd84ba2167fd2ca7ccd8531b035d390fadd582a9/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/GCModule.java#L56

Currently, the metrics are stored at a minute level in the backend, so I think this is fine.

wu-sheng:
> > It seems the OAP side will calculate GC time per second. This change should consider more.
> 
> @dylanforest @wu-sheng Sorry for my late reply, this PR looks good to me.
> 
> And I should correct what I said before, the agent side will calculate the GC time rather than the OAP side.
> 
> https://github.com/apache/skywalking-java/blob/cd84ba2167fd2ca7ccd8531b035d390fadd582a9/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/GCModule.java#L56
> 
> Currently, the metrics are stored at a minute level in the backend, so I think this is fine.

Thanks for the clarification.  
> Currently, the metrics are stored at a minute level in the backend

About this, I want to add a little more. The metrics are `SUM`ed in minute/hour/day. From the visualization part, it shows `VM GC Time (ms)` and `JVM GC Count`  in the time buckets of the range. 

____

I agree, this change should be fine.

wu-sheng:
BTW @nisiyong Are you feeling better now?

docs/en/setup/service-agent/java-agent/configurations.md
```diff
@@ -47,6 +47,7 @@ This is the properties list supported in `agent/config/agent.config`.
...
+| `jvm.metrics_collect_period`                              | The period in seconds of JVM metrics collection.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       | SW_JVM_METRICS_COLLECT_PERIOD                                    | `1`                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
```
wu-sheng, line 50:
```suggestion
| `jvm.metrics_collect_period`                              | The period in seconds of JVM metrics collection.  Unit is second.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     | SW_JVM_METRICS_COLLECT_PERIOD                                    | `1`                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
```

apm-sniffer/config/agent.config
```diff
@@ -150,6 +150,8 @@ correlation.value_max_length=${SW_CORRELATION_VALUE_MAX_LENGTH:128}
 correlation.auto_tag_keys=${SW_CORRELATION_AUTO_TAG_KEYS:}
 # The buffer size of collected JVM info.
 jvm.buffer_size=${SW_JVM_BUFFER_SIZE:600}
+# The period in seconds of JVM metrics collection.
```
wu-sheng, line 153:
```suggestion
# The period in seconds of JVM metrics collection. Unit is second.
```

wu-sheng (approved):
LGTM.

@dylanforest I added `Unit is second.` to the doc and config, to make the new config more clear.

nisiyong:
> BTW @nisiyong Are you feeling better now?

Thanks for your concern, I am much better now. Had a fever for 3 days and it was really painful.

wu-sheng:
> > BTW @nisiyong Are you feeling better now?
> 
> Thanks for your concern, I am much better now. Had a fever for 3 days and it was really painful.

Same here. 3 days fever. Take enough rest. This doesn't go away immediately. The fever could be back.

PR #429, opened 2022-12-23
Add kafka-clients 3.x support

lujiajing1126:
### Fix signature change of the intercepted method in kafka-clients 3.2.x

Due to the upstream PR https://github.com/apache/kafka/pull/11046, kafka-clients 3.2.x has changed the signature of `pollForFetches`. The return type is `org.apache.kafka.clients.consumer.internals.Fetch` now.

This will cause an exception in SW agent 8.13.x

image


wu-sheng:
Clickhouse CI seems failing due to env failing.
I ask @kezhenxu94 to check whether we could use latest ck client lib on maven central.

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/KafkaConsumerInterceptor.java
```diff
@@ -100,6 +100,11 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
         return ret;
     }
 
+    @SuppressWarnings({"unchecked"})
+    protected Map<TopicPartition, List<ConsumerRecord<?, ?>>> fetchRecords(Object retObj) {
+        return (Map<TopicPartition, List<ConsumerRecord<?, ?>>>) retObj;
+    }
+
```
wu-sheng, line 107: What is the point of this method?

test/plugin/scenarios/kafka-scenario/support-version.list
```diff
@@ -25,3 +25,6 @@
 2.6.1
 2.7.0
 2.8.0
+3.0.2
+3.1.2
+3.2.3
```
wu-sheng, line 30: It looks like you are adding supports to 3.x, not just 3.2.x. Please update the PR title.

lujiajing1126: Fixed

lujiajing1126:
> Clickhouse CI seems failing due to env failing. I ask @kezhenxu94 to check whether we could use latest ck client lib on maven central.

It seems the artifact issue only exists in 0.3.0, but not 0.3.1

wu-sheng:
> > Clickhouse CI seems failing due to env failing. I ask @kezhenxu94 to check whether we could use latest ck client lib on maven central.
> 
> It seems the artifact issue only exists in 0.3.0, but not 0.3.1

The issue is, clickhouse lib has dependencies hosts on https://jitpack.io/ somehow, which is down for 2 days.

wu-sheng (approved):

PR #444, opened 2023-01-29
Add plugin for Spring 6, set development baseline to JDK 17, remove Powermock entirely

kezhenxu94:


wu-sheng:
**Notice, SkyWalking Java Agent requires JDK17 for compiling due to support Spring 6. Meanwhile, JRE for the runtime is kept as same as before. Only Spring 6 plugin is compiled in Java 17.**

@kezhenxu94 please check whether this statement is accurate, and add the final statement as project level chang log for this version.
And we should send a mail list and  a short blog to be advance.

kezhenxu94:
For all @apache/skywalking-committers , I've removed Powermock completely from skywalking-java repository and there is no reason we would bring it back in the future, I'll also try to remove it in the skywalking main repo. This is a future-prove work because Powermock would block us from upgrading the JDK baseline to Java8+, or upgrading the Junit framework to Junit 5. So be careful when you review contributors' pull requests, if they bring Powermock back, don't allow to merge.

pom.xml
```diff
@@ -345,24 +361,28 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven-compiler-plugin.version}</version>
                 <configuration>
-                    <source>${compiler.version}</source>
+                    <source>
```
lujiajing1126, line 364: For JDK17, as I understand we should switch to `--release` argument of `javac` instead of keep using `source` and `target`. This can give us better guarantee of backward-compatibility with such cross-compilation.

- https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html
- https://stackoverflow.com/questions/43102787/what-is-the-release-flag-in-the-java-9-compiler/43103038#43103038

apm-sniffer/optional-plugins/optional-spring-plugins/mvc-annotation-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v6/define/AbstractSpring5Instrumentation.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.skywalking.apm.plugin.spring.mvc.v6.define;
+
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.ClassInstanceMethodsEnhancePluginDefine;
+
+public abstract class AbstractSpring5Instrumentation extends ClassInstanceMethodsEnhancePluginDefine {
```
wankai123, line 22: should be `AbstractSpring6Instrumentation` ?

lujiajing1126 (commented):
Also I found many unexpected line breaks. Are these changes deliberate?

wankai123 (commented):
LGTM, a nit.

apm-sniffer/optional-plugins/optional-spring-plugins/mvc-annotation-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/v6/define/AbstractSpring5Instrumentation.java
kezhenxu94: Fixed

pom.xml
kezhenxu94: Sure, changed to `<release>` configuration, in plugin tests it's still `<source>` and `<target>` as the plugin tests are still built and run in Java 8

kezhenxu94:
> Also I found many unexpected line breaks. Are these changes deliberate?

The culprit is the IDE that does the auto format when I modify the file, have reverted those.

pom.xml
kezhenxu94: I reverted the `<release>` back to `<source>` and `<target>` as that won't compile in our case, and because the plugin tests are still run in Java 8 that's OK enough to ensure we don't use APIs from > Java 8

wu-sheng:
Do you update supported list doc? I can't find it.

kezhenxu94:
> Do you update supported list doc? I can't find it.

Updated

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -8,7 +8,7 @@ metrics based on the tracing data.
   * [Tomcat](https://github.com/apache/tomcat) 9
   * [Tomcat](https://github.com/apache/tomcat) 10
   * [Spring Boot](https://github.com/spring-projects/spring-boot) Web 4.x
-  * Spring MVC 3.x, 4.x 5.x with servlet 3.x
+  * Spring MVC 3.x, 4.x 5.x with servlet 3.x, 6.x
```
wu-sheng, line 11: It is better to be in a new line with optional flag.

wu-sheng: Opt 1?

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -26,7 +26,7 @@ metrics based on the tracing data.
   * [Netflix Spring Cloud Feign](https://github.com/spring-cloud/spring-cloud-openfeign) 1.1.x -> 2.x
   * [Okhttp](https://github.com/square/okhttp) 2.x -> 3.x -> 4.x
   * [Apache httpcomponent HttpClient](http://hc.apache.org/) 2.0 -> 3.1, 4.2, 4.3, 5.0, 5.1
-  * [Spring RestTemplate](https://github.com/spring-projects/spring-framework) 4.x
+  * [Spring RestTemplate](https://github.com/spring-projects/spring-framework) 4.x -> 6.x
```
wu-sheng, line 29: RestTemplate 6 plugin is in opt1 optional too?

kezhenxu94: > RestTemplate 6 plugin is in opt1 optional too?
Should be opt2 actually

> ¹Due to license incompatibilities/restrictions these plugins are hosted and released in 3rd part repository,
> go to [SkyAPM java plugin extension repository](https://github.com/SkyAPM/java-plugin-extensions) to get these.

> ²These plugins affect the performance or must be used under some conditions, from experiences. So only released in `/optional-plugins` or `/bootstrap-plugins`, copy to `/plugins` in order to make them work.

wu-sheng (approved):

lujiajing1126 (approved):

pom.xml
lujiajing1126: > I reverted the `<release>` back to `<source>` and `<target>` as that won't compile in our case, and because the plugin tests are still run in Java 8 that's OK enough to ensure we don't use APIs from > Java 8

I see. The `sun.*` is a disaster...

PR #445, opened 2023-01-30
refactor pipeline in jedis-plugin

xzyJavaX:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #[10311](https://github.com/apache/skywalking/issues/10311)


test/plugin/scenarios/jedis-3.1.x-plus-scenario/config/expectedData.yaml
```diff
@@ -101,44 +101,8 @@ segmentItems:
...
-            peer: redis-server:6379
-            skipAnalysis: false
-            tags:
-              - {key: cache.type, value: Redis}
-              - {key: cache.cmd, value: hdel}
-              - {key: cache.key, value: a}
-              - {key: cache.op, value: write}
-          - operationName: Jedis/set
+              - {key: cache.cmd, value: BATCH_EXECUTE}
+          - operationName: Jedis/exec
```
pg-yang, line 105: The spanId should be `6`

pg-yang:
`Pipeline` and `Transaction` have two main methods such as `exec/sync` or  `discard`. We'd better enhance these methods as well, especially for `Transaction`.
BTW, don't you plan to refactor jedis 4.x plugin?

lujiajing1126:
> BTW, are you going to refactor jedis 4.x plugin?

I found that in the current jedis 4.x plugin, the interception point seems to be weird. 

`Connection.sendCommand` only appends command to the `RedisOutputStream` which is a buffer(by default `8 KiB`). The real round-trip happens when `flush()` is called. So as I understand, we are now measuring the TCP connection time to the Redis server rather than the execution time of the real command. Can you pls make a confirmation?

xzyJavaX:
Yes, I plan to refactor jedis 4.x plugin.

wu-sheng:
As https://github.com/apache/skywalking-java/pull/444 changes a lot from project level, let's hold this PR an while. That PR will be merged soon.

wu-sheng:
Updated with latest main branch.

xzyJavaX:
@pg-yang @wu-sheng Please take a look.

pg-yang:
To make the plugins more stable, I suggest adding test cases for `discard`, etc.

PR #452, opened 2023-02-07
Fix a step in trace-ignore-plugin doc

Alceatraz:
Doc outdated, File `/agent/optional-plugins/apm-trace-ignore-plugin/apm-trace-ignore-plugin.config` not exist anymore, But still reference it in this doc.


wu-sheng (approved):
LGTM

PR #453, opened 2023-02-10
Refactor kotlin coroutine plugin with CoroutineContext

devkanro:
# What's Changed?
1. Using ThreadCoroutineContext to make distribution of spans and segments more readable.
2. Modify the Instrumentation point to the newCoroutineContext function, which is used to add TracingCoroutineContext to all coroutine domains.

# TODO
- [ ] The coroutine stack for exception coroutine, I haven't found a good way to implement it yet, but I will continue to try


wu-sheng:
Do you propose to change how to propogate cross coroutines?

devkanro:
> Do you propose to change how to propogate cross coroutines?

Do you mean to create something for cross-coroutine similar to ContextSnapshot for cross-thread or ContextCarrier for corss-process?

wu-sheng:
No, I am confirming what you are doing.

devkanro:
I have added some document for my code, you can check it

apm-sniffer/optional-plugins/kotlin-coroutine-plugin/src/main/java/org/apache/skywalking/apm/plugin/kotlin/coroutine/NewCoroutineContextInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public void beforeMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes, MethodInvocationContext context) {
+
+    }
+
+    @Override
+    public Object afterMethod(Class clazz, Method method, Object[] allArguments, Class<?>[] parameterTypes, Object ret, MethodInvocationContext context) {
+        if (ContextManager.isActive()) {
+            // We just need create TracingCoroutineContext for in-tracing threads.
+            // Kotlin coroutine plugin does not automatically activate tracing.
```
wu-sheng, line 39:
```suggestion
            // Create TracingCoroutineContext when have been in tracing context already.
            // Kotlin coroutine plugin does not automatically activate tracing.
```

Don't use `we/you/they` in the docs and comments. They are confusing readers.

wu-sheng (commented):
TBD due to `The coroutine stack for exception coroutine, I haven't found a good way to implement it yet, but I will continue to try`. Otherwise, this means #451 is going to be reverted.

devkanro:
Yes, at present we have three ways to go.
1. Merge this PR directly, revert #451
2. Use SPI to register `CoroutineExceptionHandler﻿`, we can handle uncaught exceptions in coroutines
3. In the same way as #451, intercept the `getExceptionResult` of `DispatchedTask`

devkanro:
`CoroutineExceptionHandler﻿` can only handle uncaught exceptions, and when other `CoroutineExceptionHandler﻿` is registered in the coroutine context, the `CoroutineExceptionHandler﻿` provided by SPI will not be executed.

The way of intercepting the `DispatchedTask.getExceptionResult` method is relatively hacky. We just refactored the plugin to be more in line with the design of kotlin coroutine, which somewhat undermines the purpose of this refactoring.

devkanro:
For me, I may adopt method 2. It is the highest priority for me to make the plug-in more in line with the kotlin design specification, which means that our plug-in is stable, efficient, and more in line with expectations.

Additionally, as much information as possible can be provided.

wu-sheng:
> `CoroutineExceptionHandler﻿` can only handle uncaught exceptions, and when other `CoroutineExceptionHandler﻿` is registered in the coroutine context, the `CoroutineExceptionHandler﻿` provided by SPI will not be executed.
> 
> The way of intercepting the `DispatchedTask.getExceptionResult` method is relatively hacky. We just refactored the plugin to be more in line with the design of kotlin coroutine, which somewhat undermines the purpose of this refactoring.

How are you going to register an SPI in agent codes? I never did this before.

devkanro:
SPI is not available in agent code because of their classloader hierarchy.

Maybe we should drop this feature.

wu-sheng:
Please update changlog to revert previous change.

CHANGES.md
```diff
@@ -12,8 +12,9 @@ Release Notes.
 * Remove Powermock entirely from the test cases.
 * Fix H2 instrumentation point
 * Refactor pipeline in jedis-plugin.
-* Enhance kotlin coroutine plugin for stack tracing.
 * Add plugin to support ClickHouse JDBC driver (0.3.2.*).
+* Refactor kotlin coroutine plugin with CoroutineContext.
+* Revert enhance kotlin coroutine plugin for stack tracing.
```
wu-sheng, line 17: I remember this feature never released. You don't need to add this `revert`.

```suggestion
```

wu-sheng, line 17:
```suggestion
```

devkanro:
Should I delete prev PR change log?

wu-sheng (dismissed):

wu-sheng:
Please resolve conflicts.

wu-sheng (approved):

PR #463, opened 2023-02-24
Fix Spring 6 test case

kezhenxu94:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/10440.


wu-sheng:
It looks like we need another plugin? Why did this test pass? Are there some dependencies changed?

kezhenxu94:
> It looks like we need another plugin?

no new plugin, just split the resttemplate plugins and add specific witness classes for different major version.

>  Why did this test pass? Are there some dependencies changed?

I don't know, I can't even reproduce the failure locally

wu-sheng (approved):

PR #466, opened 2023-02-28
fix tomcat-10x-plugin and add test case to support tomcat7.x-8.x-9.x

xzyJavaX:

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/tomcat-10x-plugin/src/main/java/org/apache/skywalking/apm/plugin/tomcat10x/TomcatInvokeInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/tomcat-10x-plugin/src/main/java/org/apache/skywalking/apm/plugin/tomcat10x/TomcatInvokeInterceptor.java
@@ -31,7 +31,6 @@ import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedI
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
 import org.apache.skywalking.apm.agent.core.util.CollectionUtil;
-import org.apache.skywalking.apm.agent.core.util.MethodUtil;
 import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
 import org.apache.skywalking.apm.util.StringUtil;
 import org.apache.tomcat.util.http.Parameters;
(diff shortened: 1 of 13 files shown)


xzyJavaX:
https://github.com/apache/skywalking-java/pull/286/files#r1119753555

wu-sheng:
Somehow Spring 6 case fails? Could you recheck?

lujiajing1126:
> Somehow Spring 6 case fails? Could you recheck?

The only issue I could find is that we expected `POST:/create` call returns status code 200, but in the actual data is 201.

But I don't think this PR is relevant since for `EntrySpan`, the tag is set when `stackDepth == currentMaxDepth`. This means only the Spring plugin can set tags.

@kezhenxu94 @wu-sheng any idea?

kezhenxu94:
> > Somehow Spring 6 case fails? Could you recheck?
> 
> The only issue I could find is that we expected `POST:/create` call returns status code 200, but in the actual data is 201.
> 
> But I don't think this PR is relevant. Since for `EntrySpan`, the tag is set when `stackDepth == currentMaxDepth`. This means only the Spring plugin can set tags.


This is somewhat relevant, I tested Spring 6 and expected the status code to be 201 but it was always 200, I was sure it was not an issue of Spring plugin but didn't dig deeper to the cause, so this PR accidentally fixes the bug too, so, please update the expected data for Spring 6, change status code from 200 to 201.

apm-sniffer/apm-sdk-plugin/tomcat-10x-plugin/src/main/java/org/apache/skywalking/apm/plugin/tomcat10x/TomcatInvokeInterceptor.java
```diff
@@ -82,9 +75,9 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
         HttpServletResponse response = (HttpServletResponse) allArguments[1];
 
         AbstractSpan span = ContextManager.activeSpan();
-        if (IS_SERVLET_GET_STATUS_METHOD_EXIST && response.getStatus() >= 400) {
+        Tags.HTTP_RESPONSE_STATUS_CODE.set(span, response.getStatus());
+        if (response.getStatus() >= 400) {
```
kezhenxu94, line 79: I think this fixes the issue I mentioned

lujiajing1126: > I think this fixes the issue I mentioned

But in the `EntrySpan`,

```java
    @Override
    public EntrySpan tag(String key, String value) {
        if (stackDepth == currentMaxDepth || isInAsyncMode) {
            super.tag(key, value);
        }
        return this;
    }
```

I suppose in the Tomcat plugin, the condition cannot be true. Do I miss something?

wu-sheng: It is better to debug this logic.

wu-sheng:
What do these 201 and 204 mean actually?

wu-sheng (approved):
I am fine with Tomcat 10 part. Wait for @kezhenxu94

kezhenxu94 (approved):
lgtm

apm-sniffer/apm-sdk-plugin/tomcat-10x-plugin/src/main/java/org/apache/skywalking/apm/plugin/tomcat10x/TomcatInvokeInterceptor.java
lujiajing1126: image

After debugging, it seems that the above method is not called. Instead, `public AbstractTracingSpan tag(AbstractTag<?> tag, String value)` is called while `EntrySpan` does not override this method (But `ExitSpan` does!). Then the `http.status_code` tag is set.

I am not sure if it is a feature or a bug? @wu-sheng @kezhenxu94

lujiajing1126: Another question is why Tomcat and Spring set different status code...

PR #468, opened 2023-03-03
Fix clickhouse0.3.2 plugin executeBatch method

wulangcode:
The executeBatch method of the current plugin does not record sql


wu-sheng:
Batch SQLs? It should be a long SQL list.

wulangcode:
> Batch SQLs? It should be a long SQL list.

Shouldn't Batch SQLs be logged? Or to limit the length?

wu-sheng:
It could be very huge, 1000+ statements are common. Usually, we don't recommend tracing all SQLs.

PR #469, opened 2023-03-06
Override tag(AbstractTag, String) method in Entry Span

xzyJavaX:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #[10480](https://github.com/apache/skywalking/issues/10480)

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/EntrySpan.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/EntrySpan.java
@@ -64,7 +64,7 @@ public class EntrySpan extends StackBasedTracingSpan {
 
     @Override
     public AbstractTracingSpan tag(AbstractTag<?> tag, String value) {
-        if (stackDepth == currentMaxDepth || tag.isCanOverwrite() || isInAsyncMode) {
+        if (stackDepth == currentMaxDepth || isInAsyncMode) {
             super.tag(tag, value);
         }
         return this;
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/EntrySpan.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/EntrySpan.java
@@ -20,6 +20,7 @@ package org.apache.skywalking.apm.agent.core.context.trace;
 
 import org.apache.skywalking.apm.agent.core.conf.Constants;
 import org.apache.skywalking.apm.agent.core.context.TracingContext;
+import org.apache.skywalking.apm.agent.core.context.tag.AbstractTag;
 import org.apache.skywalking.apm.network.trace.component.Component;
 
 /**
@@ -57,6 +58,14 @@ public class EntrySpan extends StackBasedTracingSpan {
     public EntrySpan tag(String key, String value) {
         if (stackDepth == currentMaxDepth || isInAsyncMode) {
             super.tag(key, value);
+        }
+        return this;
+    }
+
+    @Override
+    public AbstractTracingSpan tag(AbstractTag<?> tag, String value) {
+        if (stackDepth == currentMaxDepth || tag.isCanOverwrite() || isInAsyncMode) {
+            super.tag(tag, value);
         }
         return this;
     }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@ Release Notes.
 * Adapt Armeria's plugins to the latest version 1.22.x
 * Fix tomcat-10x-plugin and add test case to support tomcat7.x-8.x-9.x.
 * Fix thrift plugin generate duplicate traceid when `sendBase` error occurs
+* Override tag(AbstractTag, String) method in Entry Span.
 
 #### Documentation
 * Update docs of Tracing APIs, reorganize the API docs into six parts.


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/EntrySpan.java
```diff
@@ -61,6 +62,14 @@ public EntrySpan tag(String key, String value) {
         return this;
     }
 
+    @Override
+    public AbstractTracingSpan tag(AbstractTag<?> tag, String value) {
+        if (stackDepth == currentMaxDepth || tag.isCanOverwrite() || isInAsyncMode) {
```
kezhenxu94, line 67: Maybe 👇 ?

```suggestion
        if (tag.isCanOverwrite() && (stackDepth == currentMaxDepth || isInAsyncMode)) {
```

wu-sheng: super#tag includes `isCanOverwrite` already. I think we don't need it.

`stackDepth == currentMaxDepth || isInAsyncMode` are the only required conditions.

lujiajing1126: > super#tag includes `isCanOverwrite` already. I think we don't need it.
> 
> `stackDepth == currentMaxDepth || isInAsyncMode` are the only required conditions.

I suppose we still need `isCanOverwrite` check in the `EntrySpan`. `http.status_code` will hit this condition.

kezhenxu94: > > super#tag includes `isCanOverwrite` already. I think we don't need it.
> > `stackDepth == currentMaxDepth || isInAsyncMode` are the only required conditions.
> 
> I suppose we still need `isCanOverwrite` check in the `EntrySpan`. `http.status_code` will hit this condition.

The parent class `AbstractTracingSpan.java#tag` already checks `isCanOverwrite` so we don't need to check again here.

https://github.com/apache/skywalking-java/blob/f04866cf240ace47fa039fd5b0f548f68fb7bee6/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java#L127-L134

lujiajing1126: But in `ExitSpan`

https://github.com/apache/skywalking-java/blob/f04866cf240ace47fa039fd5b0f548f68fb7bee6/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/ExitSpan.java#L67-L72

Is this also redundent?

wu-sheng: redundant should not represent any bug. Otherwise, it is not redundant at all.
Could you explain your point?

wu-sheng: Let's leave the bug of what it is.
In this PR, we only need to fix tag set doesn't follow stack depth.

lujiajing1126: > Let's leave the bug of what it is. In this PR, we only need to fix tag set doesn't follow stack depth.

Make sense.

> redundant should not represent any bug. Otherwise, it is not redundant at all. Could you explain your point?

I do not pretty understand the original protocol. But I suppose the critical point is the priority of the `overwritable` property and the `stack depth` of the `Span`. Specifically,

1. For `EntrySpan`, does the "overwritable" tag can be overwritten even if we are not at the top of the stack?
2. For `ExitSpan`, does the "overwritable" tag can be overwritten even if we are not at the bottom of the stack?

wu-sheng: Think in this way
About entry Span, which means an incoming RPC, we could have Tomcat -> Spring. We need the information on SpringMVC provided, that is why we asked for the deepest. Even it will have more, Tomcat -> Spring -> xxx. Then before xxx happens, Spring will override Tomcat, but eventually, xxx will override Spring.
When we move to the client side(exit span), it is reverted. Then we need the first nested span. (For example), RPC(dubbo) client -> HTTPClient -> netty, we need RPC framework's information.

lujiajing1126: > Think in this way About entry Span, which means an incoming RPC, we could have Tomcat -> Spring. We need the information on SpringMVC provided, that is why we asked for the deepest. Even it will have more, Tomcat -> Spring -> xxx. Then before xxx happens, Spring will override Tomcat, but eventually, xxx will override Spring. When we move to the client side(exit span), it is reverted. Then we need the first nested span. (For example), RPC(dubbo) client -> HTTPClient -> netty, we need RPC framework's information.

Yes. I understand how the stack works. But suppose we have an interceptor between Spring and Tomat,

```
Tomcat(before) -> Spring (before) -> [business code] -> Spring (after) -> [user-defined interceptor] -> Tomcat (after)
                                                             ^
                                                             |
                                                             |
                                                             |
                                                    set http.staus_code
```

It is possible that in the `[user-defined interceptor]`, e.g. could be a [`GenericFilterBean`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/filter/GenericFilterBean.html) in this case, a new response code is set. So I guess that is where `tag.isCanOverwrite()` works.

So I think we should keep `tag.isCanOverwrite()` in both `EntrySpan` and `ExitSpan`.

wu-sheng: That is right. But you didn't follow what I said. I mean this logic is included when you call `super#tag`. I didn't mean this is wrong. Read `AbstractTracingSpan#tag`

lujiajing1126: > That is right. But you didn't follow what I said. I mean this logic is included when you call `super#tag`. I didn't mean this is wrong. Read `AbstractTracingSpan#tag`

I'm confused

> stackDepth == currentMaxDepth || isInAsyncMode are the only required conditions.

If this condition does not meet, how could the super method been called?

wu-sheng: An override tag doesn't mean you could override without depth. That is not from the design perspective. 
Override tag means Spring could override Tomcat. It isn't reverted.

wu-sheng: @lujiajing1126 BTW, why are we talking about this here? I don't think this is relative to this bug at all.

lujiajing1126: > @lujiajing1126 BTW, why are we talking about this here? I don't think this is relative to this bug at all.

I just want to confirm the original design. But for the `ExitSpan` case, we always allow overwrite? Is that a design-oriented decision as well?

Though I suppose there are still limitations, such as the `GenericBeanFilter` case I mentioned above...

wu-sheng:
Any update about this?

lujiajing1126:
> Any update about this?

Too many test failures. I suppose we may address https://github.com/apache/skywalking/issues/10479 first. WDYT

wu-sheng:
Make sense to me.

wu-sheng:
Close for now. No update in a month.

PR #479, opened 2023-03-14
Support keep trace profiling when cross-thread

mrproliu:
Following https://github.com/apache/skywalking/issues/10373, support collecting cross-thread trace profiling snapshots.
The basic principle is to save the profiling state of the current thread when using `ContextManager#capture`, and determine whether to add the current thread to the profiling task according to the profiling state in the current snapshot when `ContextManager#continued`.
Theoretically, as long as Java Agent can analyze the segment generated by 'cross-thread', it supports the profiling operations.


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionContext.java
```diff
@@ -72,39 +72,57 @@ public void stopProfiling() {
...
+
+    public boolean continueProfiling(TracingContext tracingContext, String traceSegmentId) {
+        // must smaller than max parallels monitoring count
+        if (currentProfilingCount.addAndGet(1) >= Config.Profile.MAX_PARALLEL) {
+            currentProfilingCount.addAndGet(-1);
+            return false;
+        }
+
+        return addProfilingThread(tracingContext, traceSegmentId) != null;
+    }
```
wu-sheng, line 114: So, we still have the threshold to control this for limiting the payload, right?

What is the threshold for UI, and what is for agent by default? Are they reasonably matched?

mrproliu: > What is the threshold for UI?

For now, the UI only limits the count of the first segment. The sampling count is not verified when continue profiling the cross-thread segment. 
I think the user only should only care about the profiling limit count of the first segment. The cross-thread segment should be kept profiling. 

> what is for agent by default?

The agent doesn't provide the sampling count, only receives the sampling count from OAP. At present, the agent only provides `profile.max_parallel` to limit the number of agent parallel analyses, so as to prevent dump data from affecting the execution of the business system.

> Are they reasonably matched?

The only thing that may be worried is that the value of `profile.max_parallel` is too small, which may cause the limited cross-thread not to be collected. However, this trade-off point, if the `profile.max_parallel` data is too large, may affect the execution of the business during profiling. If it is too small, it may lead to a lack of collected data.

wu-sheng: I just want to make sure `profile.max_parallel` is big enough for most cases. Such as for undertow(a typical requirement for this new feature).

mrproliu: `Undertow` may be an individual case, because it main thread quickly completes execution and closes, and the sub-thread will execute according to business time. In this case, the `profile.max_parallel` may be enough, because the main-thread slot will empty quickly. 
However, for the usual cross-thread execution, it may not be enough, because the main thread may continue to wait for the sub-thread execution results and return.
So I think maybe we can increase the default value of `profile.max_parallel`. Assuming that a thread will generate an additional sub-thread by default, the existing `5` can be raised to `10`.

wu-sheng: > So I think maybe we can increase the default value of profile.max_parallel. Assuming that a thread will generate an additional sub-thread by default, the existing 5 can be raised to 10.

By adding this, there are some side effects, such as adding a payload of profiling to sync-process cases.
How about adding a threshold `profile.accept_max_sub_parallel_num`? This should only control the number of sub-tasks.

mrproliu: Sure, but is this mean the total sub-thread count for all segments or one segment?

wu-sheng: I think per-endpoint-access(one main segment)

mrproliu: Sure, so this configuration should validate the profiling count of one endpoint(first segment), and the default value could be `1` I think. That's means we only keep one sub-thread profiling.

wu-sheng: It could be 5, as mostly it should not be in that case.

wu-sheng: After all releases(new agent and new OAP/UI), we could write a blog introducing this with undertow and real parallel cases. The the default as 5 would make the demo more easier.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -241,10 +241,15 @@ public static class Profile {
         public static boolean ACTIVE = true;
 
         /**
-         * Parallel monitor segment count
+         * Parallel monitor endpoint thread count
          */
         public static int MAX_PARALLEL = 5;
 
+        /**
+         * Max monitoring sub-tasks count of one endpoint
```
wu-sheng, line 249:
```suggestion
         * Max monitoring sub-tasks count of one single endpoint access
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionContext.java
mrproliu: Updated.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionContext.java
```diff
@@ -72,39 +72,57 @@ public void stopProfiling() {
...
             tracingContext, traceSegmentId, Thread.currentThread(), this);
         int slotLength = profilingSegmentSlots.length();
         for (int slot = 0; slot < slotLength; slot++) {
             if (profilingSegmentSlots.compareAndSet(slot, null, threadProfiler)) {
-                return threadProfiler.profilingStatus();
+                return threadProfiler;
             }
         }
-        return ProfileStatusReference.createWithNone();
+        return null;
```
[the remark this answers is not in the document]
mrproliu: Updated.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/profile/ProfileTaskExecutionContext.java
```diff
@@ -48,7 +48,7 @@ public class ProfileTaskExecutionContext {
 
     public ProfileTaskExecutionContext(ProfileTask task) {
         this.task = task;
-        profilingSegmentSlots = new AtomicReferenceArray<>(Config.Profile.MAX_PARALLEL);
+        profilingSegmentSlots = new AtomicReferenceArray<>(Config.Profile.MAX_PARALLEL * (Config.Profile.MAX_ACCEPT_SUB_PARALLEL + 1));
```
wu-sheng, line 51: I think this may not current. This would cause `if (profilingEndpointCount >= Config.Profile.MAX_PARALLEL) {` true even if the slots are still available.

mrproliu: The `profilingEndpointCount`  is only the count of the profiled endpoint, not including the sub-tasks.
Also, If we need the sub-process(`continue`) profiling, need to use `ProfileTaskExecutionContext#continueProfiling
` instance of `ProfileTaskExecutionContext#attemptProfiling` method.

wu-sheng (approved):
LGTM

PR #487, opened 2023-03-27
Fix a thread leak in `SamplingService` when updated sampling policy in the runtime.

hyhyf:


hyhyf:
[https://github.com/apache/skywalking/discussions/10549#discussioncomment-5437985](url)

wu-sheng:
Which JDK are you using? Have you checked the source codes?

wu-sheng:
I think the change makes sense. It is a leak. `scheduledFuture#cancel` doesn't close the whole pool actually.

wu-sheng:
@hyhyf Please add `* Fix a thread leak in `SamplingService` when updated sampling policy in the runtime.` as the change log into the `changes.md` in the root. We need the description for the changes.

wu-sheng (approved):

PR #489, opened 2023-03-28
support mysql when useServerPrepStmts

xu1009:
### <support mysql trace sql param when useServerPrepStmts>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<10589>.

### DIFF
--- a/test/plugin/scenarios/mysql-scenario/configuration.yml
+++ b/test/plugin/scenarios/mysql-scenario/configuration.yml
@@ -30,3 +30,4 @@ dependencies:
     environment:
       - MYSQL_ROOT_PASSWORD=root
       - MYSQL_DATABASE=test
+      - SW_JDBC_TRACE_SQL_PARAMETERS=true
--- a/test/plugin/scenarios/mysql-scenario/configuration.yml
+++ b/test/plugin/scenarios/mysql-scenario/configuration.yml
@@ -19,6 +19,7 @@ entryService: http://localhost:8080/mysql-scenario/case/mysql-scenario
 healthCheck: http://localhost:8080/mysql-scenario/case/healthCheck
 startScript: ./bin/startup.sh
 environment:
+  - SW_JDBC_TRACE_SQL_PARAMETERS=true
 depends_on:
   - mysql-server
 dependencies:
@@ -30,4 +31,4 @@ dependencies:
     environment:
       - MYSQL_ROOT_PASSWORD=root
       - MYSQL_DATABASE=test
-      - SW_JDBC_TRACE_SQL_PARAMETERS=true
+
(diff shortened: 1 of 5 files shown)


test/plugin/scenarios/mysql-scenario/src/main/resources/jdbc.properties
```diff
@@ -13,6 +13,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-mysql.url=jdbc:mysql://mysql-server:3306/test?serverTimezone=CST&enabledTLSProtocols=TLSv1.2&useSSL=false
+mysql.url=jdbc:mysql://mysql-server:3306/test?serverTimezone=CST&enabledTLSProtocols=TLSv1.2&useSSL=false&useServerPrepStmts=true
```
wu-sheng, line 16: Is this going to override the PreparedStatement cases?

wu-sheng: @xu1009 ☝️

wu-sheng: Generally, as this has been tested, I think we should be OK to revert the changes. To keep the duplicate cases for just this one thing seems a little overreactive.

test/plugin/scenarios/mysql-scenario/src/main/java/org/apache/skywalking/apm/testcase/mysql/SQLExecutor.java
```diff
@@ -53,28 +54,28 @@ public void insertData(String sql, String id, String value) throws SQLException
     public void dropTable(String sql) throws SQLException {
         executeStatement(sql);
     }
-    
+
```
wu-sheng, line 57: This seems unexpected change?

test/plugin/scenarios/mysql-scenario/configuration.yml
```diff
@@ -30,3 +30,4 @@ dependencies:
     environment:
       - MYSQL_ROOT_PASSWORD=root
       - MYSQL_DATABASE=test
+
```
wu-sheng, line 33: Please revert this too.

test/plugin/scenarios/mysql-scenario/src/main/java/org/apache/skywalking/apm/testcase/mysql/SQLExecutor.java
xu1009: it seems some Newline character different

CHANGES.md
```diff
@@ -23,6 +23,7 @@ Release Notes.
 * Support keep trace profiling when cross-thread.
 * Fix unexpected whitespace of the command catalogs in several Redis plugins.
 * Fix a thread leak in `SamplingService` when updated sampling policy in the runtime.
+* Support MySQL plugin trace sql parameters when useServerPrepStmts
```
wu-sheng, line 26:
```suggestion
* Support MySQL plugin trace SQL parameters when useServerPrepStmts 
```

wu-sheng (approved):

PR #515, opened 2023-04-25
clickhouse-jdbc-plugin support trace sql parameters when enable plugin.jdbc.trace_sql_parameters

xu1009:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number.  Close https://github.com/apache/skywalking/issues/10729.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
+++ b/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
@@ -38,9 +38,6 @@ public class StatementEnhanceInfos {
         this.statementName = statementName;
     }
 
-    public StatementEnhanceInfos() {
-    }
-
     public ConnectionInfo getConnectionInfo() {
         return connectionInfo;
     }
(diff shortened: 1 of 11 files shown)


wu-sheng:
Is this PR official? I see the title is `test add`.

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/define/StatementEnhanceInfos.java
```diff
@@ -38,6 +38,9 @@ public StatementEnhanceInfos(ConnectionInfo connectionInfo, String sql, String s
         this.statementName = statementName;
     }
 
+    public StatementEnhanceInfos() {
```
wu-sheng, line 41: Why do you build an empty one? You can't get the necessary information?

wu-sheng: Usually, if these information is available, it is better to build in the constructor.
Otherwise, one day, we may face uncertain NPE.

xu1009: i can get the information, but it is useless, i just use param set part,  ok, i will use the constructor and set  information to it

wu-sheng (approved):

PR #517, opened 2023-04-27
support jetty server work thread pool metric monitor

xu1009:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. https://github.com/apache/skywalking/issues/10736.

### DIFF
--- a/.github/workflows/plugins-jdk17-test.0.yaml
+++ b/.github/workflows/plugins-jdk17-test.0.yaml
@@ -58,6 +58,7 @@ jobs:
           - jdk17-with-gson-scenario
           - resttemplate-6.x-scenario
           - jetty-thread-pool-scenario
+          - jetty-11.x-thread-pool-scenario
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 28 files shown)


xu1009:
the plugin supports jetty 11，but jetty 11 example is different with lower version, should i write new scenario test for 11

xu1009:
ok，i  write another test for 11, have fun

.github/workflows/plugins-jdk17-test.0.yaml
```diff
@@ -57,6 +57,7 @@ jobs:
         case:
           - jdk17-with-gson-scenario
           - resttemplate-6.x-scenario
+          - jetty-thread-pool-scenario
```
wu-sheng, line 60:
```suggestion
          - jetty-thread-pool-scenario
          - jetty-11.x-thread-pool-scenario
```

wu-sheng: @xu1009 I believe the both cases should be added.

wu-sheng (approved):

PR #521, opened 2023-04-29
Improve bytebuddy class enhance

kylixs:
### Improve bytebuddy class enhance
- [ x ] If this is non-trivial feature, paste the links/URLs to the design doc.
  https://github.com/kylixs/kylixs.github.io/blob/master/skywalking/byteduddy-retransfrom-faliure-exploration.md

* Add bytebuddy patch to support retransform class multiple times compatible with other javaagent without class cache.
* Support for hotswap classes that are enhanced by skywalking plugins, like Spring MVC Controller, etc.


wu-sheng:
I will be back on this at May 3rd. 
Thanks for your detailed explanation.
I will take a close look and provide feedback.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
```diff
@@ -65,8 +65,8 @@ public abstract class ClassEnhancePluginDefine extends AbstractClassEnhancePlugi
      */
     @Override
     protected DynamicType.Builder<?> enhanceInstance(TypeDescription typeDescription,
-        DynamicType.Builder<?> newClassBuilder, ClassLoader classLoader,
-        EnhanceContext context) throws PluginException {
+                                                     DynamicType.Builder<?> newClassBuilder, ClassLoader classLoader,
+                                                     EnhanceContext context) throws PluginException {
```
wu-sheng, line 69: These seem some local code style differences. Could you revert these changes?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
```diff
@@ -113,15 +114,15 @@ protected DynamicType.Builder<?> enhanceInstance(TypeDescription typeDescription
...
                 if (isBootstrapInstrumentation()) {
                     newClassBuilder = newClassBuilder.constructor(constructorInterceptPoint.getConstructorMatcher())
-                                                     .intercept(SuperMethodCall.INSTANCE.andThen(MethodDelegation.withDefaultConfiguration()
-                                                                                                                 .to(BootstrapInstrumentBoost
-                                                                                                                     .forInternalDelegateClass(constructorInterceptPoint
-                                                                                                                         .getConstructorInterceptor()))));
+                            .intercept(SuperMethodCall.INSTANCE.andThen(MethodDelegation.withDefaultConfiguration()
+                                    .to(BootstrapInstrumentBoost
+                                            .forInternalDelegateClass(constructorInterceptPoint
+                                                    .getConstructorInterceptor()))));
```
wu-sheng, line 120: Please revert all unnecessary changes.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -161,9 +175,10 @@ public DynamicType.Builder<?> transform(final DynamicType.Builder<?> builder,
                                                 final JavaModule javaModule,
                                                 final ProtectionDomain protectionDomain) {
             LoadedLibraryCollector.registerURLClassLoader(classLoader);
+            DelegateNamingResolver.reset();
```
wu-sheng, line 178: I am a little confused about this `reset`. `#transform` should be called for every type, as a static method, we reset every time?

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/NativeMethodStrategySupport.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public static void inject(AgentBuilder agentBuilder) {
+        Class<? extends AgentBuilder> clazz = agentBuilder.getClass();
+        if (clazz != AgentBuilder.Default.class) {
+            throw new IllegalStateException("Only accept original AgentBuilder instance but not a wrapper instance: " + clazz.getName());
+        }
+        try {
+            Field nativeMethodStrategyField = clazz.getDeclaredField("nativeMethodStrategy");
+            nativeMethodStrategyField.setAccessible(true);
+            nativeMethodStrategyField.set(agentBuilder, new SWNativeMethodStrategy(PREFIX));
```
wu-sheng, line 41: These seem a hijacking and relying on upstream(bytebuddy) doesn't change the internal implementation.
Considering bytebuddy is always very friendly to our community, you are better to submit a proposal/request to bytebuddy, and discuss the official API to do so.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/implementation/SWAuxiliaryTypeNamingStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+    private String findDescription(AuxiliaryType auxiliaryType) {
+        try {
+            Class<? extends AuxiliaryType> auxiliaryTypeClass = auxiliaryType.getClass();
+            String auxiliaryTypeClassName = auxiliaryTypeClass.getName();
+            if (auxiliaryTypeClassName.endsWith("Morph$Binder$RedirectionProxy")
+                    || auxiliaryTypeClassName.endsWith("MethodCallProxy")) {
+                // get MethodDescription from field 'specialMethodInvocation.methodDescription'
+                Field specialMethodInvocationField = auxiliaryTypeClass.getDeclaredField("specialMethodInvocation");
+                specialMethodInvocationField.setAccessible(true);
+                SpecialMethodInvocation specialMethodInvocation = (SpecialMethodInvocation) specialMethodInvocationField.get(auxiliaryType);
```
wu-sheng, line 66: This is also another kind hijacking, please considering working with the upstream.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/implementation/SWAuxiliaryTypeNamingStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+                MethodDescription methodDescription = specialMethodInvocation.getMethodDescription();
+                return methodDescription.toString();
+
+            } else if (auxiliaryTypeClassName.endsWith("Pipe$Binder$RedirectionProxy")) {
+                // get MethodDescription from field 'sourceMethod'
+                Field sourceMethodField = auxiliaryTypeClass.getDeclaredField("sourceMethod");
```
wu-sheng, line 72: Same hijacking concern.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/implementation/SWAuxiliaryTypeNamingStrategy.java
```diff
@@ -0,0 +1,89 @@
...
+                sourceMethodField.setAccessible(true);
+                MethodDescription sourceMethod = (MethodDescription) sourceMethodField.get(auxiliaryType);
+                return sourceMethod.toString();
+
+            } else if (auxiliaryTypeClassName.endsWith("FieldProxy$Binder$AccessorProxy")) {
+                // get fieldDescription
+                Field fieldDescriptionField = auxiliaryTypeClass.getDeclaredField("fieldDescription");
```
wu-sheng, line 79: Same hijacking concern.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/implementation/SWImplementationContextFactory.java
```diff
@@ -0,0 +1,52 @@
...
+package net.bytebuddy.implementation;
+
+import net.bytebuddy.ClassFileVersion;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.dynamic.scaffold.TypeInitializer;
+import net.bytebuddy.implementation.auxiliary.AuxiliaryType;
+import net.bytebuddy.utility.RandomString;
+
+/**
+ * Design to generate fixed name for CacheValueField (cache delegating Method)
```
wu-sheng, line 28: This comment doesn't fit the codes. Please reword this, you are building a ContextFactory with target type name based suffix on the top of `Implementation.Context.Default`.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWAsmVisitorWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+import net.bytebuddy.jar.asm.Opcodes;
+import net.bytebuddy.pool.TypePool;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Remove duplicated fields of instrumentedType
+ */
+public class SWAsmVisitorWrapper implements AsmVisitorWrapper {
```
wu-sheng, line 38: I am not fully following this. Are you trying to cache all original methods before the instrumentation?
The name `RemoveDuplicatedFieldsClassVisitor` is not very clear to me.
Please reword the comments for `SWAsmVisitorWrapper`

wu-sheng (commented):
2 main things should be considered
1. Too many hijacking through reflection. This relies on byte-buddy internal codes/implementations. We don't want to make further upgrading hard. Please work with upstream and try to find a new way, or request byte-buddy to expose new APIs for this.
2. As you are introducing a new way of `retransfer`, removing the old mechanism for the class cache is better because it isn't always perfect.

wu-sheng:
Hi @kylixs Thanks for providing your way of debugging and resolving the issue. I believe locking the field and method name for auxiliary class and method is a way to resolve the transfer issue.
But meanwhile, I am not very certain, whether this is always perfect, as SkyWalking agent kernel has been used in many other cases, such as other APMs' agent are actually a fork of upstream. 
Is this way going to bring a new conflict between skywalking agent and skywalking agent fork(re-distribution)? Our answer doesn't have to be that, we support this kind of hybrid deployment, but I want to include this as a consideration.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
wu-sheng: If byte-buddy could expose the whole method signature rather than the name, I believe this would not be an issue.
How about asking for upstream first?

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/NativeMethodStrategySupport.java
kylixs: OK, That's what I am thinking.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWAsmVisitorWrapper.java
kylixs: I'll check it agian.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/implementation/SWAuxiliaryTypeNamingStrategy.java
kylixs: ok

kylixs:
> 

It's a bit complicated, and we haven't tested enough scenarios to know if there are other compatibility issues. I came up with a few ideas:
* Provide a fallback transform mode. If there is a compatibility problem, users can fall back to the previous mode. We can keep improving it.
* Create a retransform javaagent for automated testing every plugin for compatibility, and may need to modify testing tools to support it.

wu-sheng:
> Provide a fallback transform mode. If there is a compatibility problem, users can fall back to the previous mode. We can keep improving it.

Falling back usually is too complex to maintain. 

___

About the re-transfer, once we have a clear path about how skywalking agent works in that agent, I think we are fine.
The question I left is only about whether we are going to conflict with our own fork, due to names are always same in two agents.

kylixs:
> About the re-transfer, once we have a clear path about how skywalking agent works in that agent, I think we are fine.
> The question I left is only about whether we are going to conflict with our own fork, due to names are always same in two agents.

The key issue is the class and method that the plug-in intercepts. As you can see from the documentation of the process, the processing path is fairly clear, but it may not include all of the paths. Currently, all skywalking plug-ins are supported, but we don't know if any of the custom plug-ins have compatibility issues. At present, we have a better understanding of ByteBuddy processing process and debugging tools, which can solve the problem quickly.

kylixs:
> The question I left is only about whether we are going to conflict with our own fork, due to names are always same in two agents.

I don't quite understand this. Could you explain it in detail?

wu-sheng:
> > The question I left is only about whether we are going to conflict with our own fork, due to names are always same in two agents.
> 
> 
> 
> I don't quite understand this. Could you explain it in detail?

The name of fields and methods are predictable and will be same.
There are other solutions of other vendors using our agent to build their own distributions, which may be APM agents or even others.
Once the names are locked, would they conflict with each others?

kylixs:
ByteBuddy uses random names, which is essential in dynamic proxy/class creation, but in APM scenarios, dynamic names are bad.
We use `sw_xxx$` as a specific flag for auxiliary class names, field names, and method names that do not conflict with other vendor names.
Vendors can extend with the Skywalking agent interceptor mechanism, or register their own class transformer, which is fine. The locking name only affects the Skywalking agent itself, and vendors using their own ByteBuddy instances are unchanged.

wu-sheng:
I know they can, I mean many of them are not doing this.
I think you still are not following what is redistribution of open source. Thinking about Ubuntu relationship with upstream Linux. Take SkyWalking as upstream Linux, vendor agents are a kind of Ubuntu.

kylixs:
If they haven't modified the Skywalking agent's plugin interceptor flow, they should be compatible. If they change the interceptor mechanism or the ByteBuddy option, they need to be compatible with the new mode. More information? Are there any known redistribution projects of Skywalking agent?

wu-sheng:
I can't provide this information as I don't have their owner permission, sorry.
I just want to mention this as a case. Such as if they have a modified plugin/interceptor but targeting the same class and same method.

wu-sheng:
> Custom plugins: For example, two plugins enhance the same method of the same class. Not tested yet, don't know if this is currently supported. If we want to support it, we should define it as a feature and add test cases.

This is the case I want to discuss it. AFAIK, we have problems about this.

kylixs:
> > Custom plugins: For example, two plugins enhance the same method of the same class. Not tested yet, don't know if this is currently supported. If we want to support it, we should define it as a feature and add test cases.
> 
> This is the case I want to discuss it. AFAIK, we have problems about this.

Some discoveries:
* A ByteBuddy/AgentBuilder instance installs multiple interceptors on the same method of the same class, with only the last one taking effect
* Multiple ByteBuddy/AgentBuilder instances that install an interceptor on the same method of the same class form an interceptor call chain
* Locking names does not change the behavior above

However, we have only one instance of ByteBuddy, and it cannot directly generate an interceptor chain. If we want to solve the problem of multiple interceptions of the same method, we may need to modify the interceptor mechanism to maintain a chain of interceptors within the Skywalking agent.

wu-sheng:
> Multiple ByteBuddy/AgentBuilder instances that install an interceptor on the same method of the same class form an interceptor call chain

How about two skywalking agents for one target? I don't expect we would do two interceptors for one target inside skywalking. The other one is fork distribution. Will that work?

> Locking names does not change the behavior above

Generally, we are fine to the new solution.

kylixs:
> How about two skywalking agents for one target? I don't expect we would do two interceptors for one target inside skywalking. The other one is fork distribution. Will that work?

If both agents lock names and the special identifier in the name is the same, there will be a conflict. We can solve this problem by randomly generating an identity when the agent starts.
Others need to be tested in practice. I think there is no big problem to do so, feedback later.

wu-sheng:
> If both agents lock names and the special identifier in the name is the same, there will be a conflict. We can solve this problem by randomly generating an identity when the agent starts.

+1

kylixs:
> How about two skywalking agents for one target? I don't expect we would do two interceptors for one target inside skywalking. The other one is fork distribution. Will that work?

Unfortunately, there is a problem with intercepting constructors: Once a constructor is intercepted on one agent, other methods of the class that are intercepted on another agent will fail.
The second agent could not find the constructor's dependent auxiliary class (created in the first agent) when parsing the target class. The auxiliary class actually exists in the classloader, but couldn't get the bytecode, even using `ClassFileLocator. ForInstrumentation`:
```agentBuilder.with(ClassFileLocator.ForInstrumentation.fromInstalledAgent(...) )```

wu-sheng:
Got it. Please ask for help from byte-buddy. I think we at least need two issues on upstream for making this ready.

kylixs:
OK, I will submit the problem and demo code to ByteBuddy.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
kylixs: What do you mean? Tell me more about it.

kylixs:
> How about two skywalking agents for one target? I don't expect we would do two interceptors for one target inside skywalking. The other one is fork distribution. Will that work?

Intercept same classes with two AgentBuilders are ok. Do you have any ideas on how to add specific test cases?

wu-sheng:
> > How about two skywalking agents for one target? I don't expect we would do two interceptors for one target inside skywalking. The other one is fork distribution. Will that work?
> 
> 
> 
> Intercept same classes with two AgentBuilders are ok. Do you have any ideas on how to add specific test cases?

What do you mean OK? We only have one, right? You changed it to multiple? Or one per plugin or something?

kylixs:
> What do you mean OK? We only have one, right? You changed it to multiple? Or one per plugin or something?

Currently, it is supported to install an interceptor on the same method in two AgentBuilders respectively, resulting in a chain of proxy methods.

I think there are a few scenarios where conflicts need to be resolved:
1. The interception method of user-defined plug-ins may overlap with the existing plug-ins of Skywalking agent.
2. Two Skywalking agent plug-ins may handle the same method.

Interception methods can overlap in many ways, such as two plug-ins matching different annotations but having both annotations in the same class or method.

I think it is better to solve the above problems in one agent, because putting it into two agents will bring maintenance workload and confusion. I think users would prefer to maintain one custom plug-in rather than two agents.

One idea is to identify plug-ins that overlap and use different AgentBuilders to handle them. I'll try to see if this works.

wu-sheng:
Generally, the ides works for me.

kylixs:
Took some time to try two solutions:
1. Each plugin uses an isolated `AgentBuilder`, and it turns out that each AgentBuilder instance do a lot of repetitive processing, which makes Java application start too slowly.

2. Each plugin uses an isolated `Transformer` and found that ByteBuddy is not supported.
I made some rough changes to the ByteBuddy AgentBuilder code, and it is now running, loading two plugins at the same time to enhance the classes of Spring Controller, and they both work.

I will submit it to the ByteBuddy community and need more discussion and design.

ByteBuddy code: https://github.com/kylixs/byte-buddy/tree/separate-transformer
Custom spring mvc plugin: https://github.com/kylixs/spring-mvc-plugin-demo

wu-sheng:
I remembered that @raphw mentioned, the `advice` annotation rather than delegator could support multiple interceptors. But I didn't try.

kylixs:
> I remembered that @raphw mentioned, the `advice` annotation rather than delegator could support multiple interceptors. But I didn't try.

We can intercept the target class method using Advice, and then filter the matching plugins in the Advice static proxy method and invoke the intercepting classes of plugins.
It seems that this method can adapt to the original plugin, but need to change the plugin mechanism, I try it.

wu-sheng:
It is good to tesr various possibilities. Eventually, we should keep changes as less as possible, to keep stability.

raphw:
Sorry, but this does not go into a direction that I see being merged into Byte Buddy. But you are very welcome to implement this on top of existing APIs and to offer it as an extension.

wu-sheng:
@kylixs I am back on this as the 9.5.0 had been released. AFAIK, there are several things scoped in this PR, which makes it very hard to review and move forward from tech and stable perspectives.
So, I want to separate the issue into the following parts
1. Support predictable field names for interceptors
2. Support retransfer with other agents. I think it is easy if we landed <1>
3. Support multiple interceptors for one single method. This may be the hard part, but to be honest, this is not a very common use case inside SkyWalking.

So, how about we are going to do <1> only first? I think from your existing changes, it should be easy to support that without hijacking?

wu-sheng:
> I agree with your idea, but I think <1> and <2> are the same things, and have been reached.
I'm not sure to support that without hijacking, maybe make it a little more elegant.

I don't think we will avoid all hijackings, but we could limit the scope, currently, they are too wide. And we could add tests to verify them when we need to bump up the version of Byte-buddy.

kylixs:
move to: https://github.com/apache/skywalking-java/pull/561

PR #522, opened 2023-05-02
support MutableLogEvent in plugin log4j2

Chenfx-git:
If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<10617>.


wu-sheng:
Please update the docs according, and add plugin test to verify logs.

Chenfx-git:
> Please update the docs according, and add plugin test to verify logs.

I saw test scenarios was verity segment. How to verity traceid in log? And I didn't see that the log4j2 plugin had written test cases before.

wu-sheng:
You don't need to test that, you just need to verify logs having the trace ID.
There isn't for other log plugins because
1. We don't have log verification mechanism. We added later.
2. The previous plugins were added by the PMC member or committer. They would take the responsibility to respond bug reporter. But we don't add and can't add this to you.

wu-sheng:
Closing as no update

PR #524, opened 2023-05-04
Fix ClassCastException when mssql inserting data (#10741)

sky-dawn:
Fix SQLServer write data intercept failure
The issue link: [https://github.com/apache/skywalking/issues/10741](url)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@ Release Notes.
 * Support clickhouse-jdbc-plugin trace sql parameters
 * Support monitor jetty server work thread pool metric
 * Support Jersey REST framework
-* Fix ClassCastException when SQLServer inserting data (#10741)
+* Fix ClassCastException when SQLServer inserts data 
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Release Notes.
 * Support clickhouse-jdbc-plugin trace sql parameters
 * Support monitor jetty server work thread pool metric
 * Support Jersey REST framework
+* Fix ClassCastException when SQLServer inserting data (#10741)
 
 #### Documentation
 
--- a/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/JDBCPreparedStatementSetterInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/JDBCPreparedStatementSetterInterceptor.java
@@ -24,13 +24,14 @@ import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInt
 import org.apache.skywalking.apm.plugin.jdbc.define.StatementEnhanceInfos;
 
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 
 public class JDBCPreparedStatementSetterInterceptor implements InstanceMethodsAroundInterceptor {
     @Override
     public final void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
         Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
         final StatementEnhanceInfos statementEnhanceInfos = (StatementEnhanceInfos) objInst.getSkyWalkingDynamicField();
-        if (statementEnhanceInfos != null) {
+        if (statementEnhanceInfos != null && Modifier.isPublic(method.getModifiers())) {
           final int index = (Integer) allArguments[0];
           final Object parameter = allArguments[1];
           statementEnhanceInfos.setParameter(index, parameter);
--- a/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/JDBCPreparedStatementSetterInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/JDBCPreparedStatementSetterInterceptor.java
@@ -24,14 +24,13 @@ import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInt
 import org.apache.skywalking.apm.plugin.jdbc.define.StatementEnhanceInfos;
 
 import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
 
 public class JDBCPreparedStatementSetterInterceptor implements InstanceMethodsAroundInterceptor {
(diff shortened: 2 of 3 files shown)


apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/JDBCPreparedStatementSetterInterceptor.java
```diff
@@ -24,13 +24,14 @@
...
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 
 public class JDBCPreparedStatementSetterInterceptor implements InstanceMethodsAroundInterceptor {
     @Override
     public final void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments,
         Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
         final StatementEnhanceInfos statementEnhanceInfos = (StatementEnhanceInfos) objInst.getSkyWalkingDynamicField();
-        if (statementEnhanceInfos != null) {
+        if (statementEnhanceInfos != null && Modifier.isPublic(method.getModifiers())) {
```
wu-sheng, line 34: If you need to change the intercepting points, please change the codes of matcher in the instrument class.

sky-dawn: Do you mean don't change the interceptor code? 

In general, non-public method is the underlying implementation, I think it need not to be intercept.
So if I can change the PSSetterDefinitionOfJDBCInstrumentation code,like this:
```
@Override
        ElementMatcher.Junction<MethodDescription> matcher = none();

        // remove TRACE_SQL_PARAMETERS judgement for dynamic config
        final Set<String> setters = ignorable ? PS_IGNORABLE_SETTERS : PS_SETTERS;
        for (String setter : setters) {
            matcher = matcher.or(named(setter));
        }

        return matcher;
    }
```
change `matcher = matcher.or(named(setter))` to `matcher = matcher.or(named(setter).and(isPublic()))`

sky-dawn: or just change the mssql plugin's Instrumentation.  I only find this problem in the mssql's interface.

wu-sheng: According to the error, it seems mssql JDBC driver has some special public method, which isn't expected to trace.
As a result, it ends in ClassCastException.
Is this the case?

Changing mssql plugins only work for me, and adding `isPublic` to filter unnecessary intercepting points make sense too if there is no side-effect for other drivers.

sky-dawn: Yes, it's ClassCastException. For detail:
`java.lang.ClassCastException: com.microsoft.sqlserver.jdbc.Parameter cannot be cast to java.lang.Integer`

It happens in this piece of code:
```
final void setObject(Parameter param, Object obj, JavaType javaType, JDBCType jdbcType, Integer scale,
            Integer precision, boolean forceEncrypt, int parameterIndex, String tvpName) throws SQLServerException 
```
a non-public method

Normally, java.sql.PreparedStatement, The first param type is int of this interface's setXX method. But mssql's underlying implementation is non-standard.

And JDBCPreparedStatementSetterInterceptor intercepts methods like:
"setDouble", "setFloat", "setInt", "setLong", "setNString", "setObject", "setAsciiStream", "setBinaryStream", "setBlob", "setBytes" ...

This methods just set params for preparedstatement. SW just intercept public methods，not non-public methods.

wu-sheng: You could try to set it for public only, I think. As we should only care about those methods.

wu-sheng:
```
[INFO] Starting audit...
Error:  /home/runner/work/skywalking-java/skywalking-java/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/PSSetterDefinitionOfJDBCInstrumentation.java:27:52: Using the '.*' form of import should be avoided - net.bytebuddy.matcher.ElementMatchers.*. [AvoidStarImport]
Audit done.
[INFO] There is 1 error reported by Checkstyle 9.3 with /home/runner/work/skywalking-java/skywalking-java/apm-checkstyle/checkStyle.xml ruleset.
Error:  src/main/java/org/apache/skywalking/apm/plugin/jdbc/PSSetterDefinitionOfJDBCInstrumentation.java:[27,52] (imports) AvoidStarImport: Using the '.*' form of import should be avoided - net.bytebuddy.matcher.ElementMatchers.*.
```

It seems you don't check the code style locally.
And you should update the changelog according to the PR template.

sky-dawn:
> ```
> [INFO] Starting audit...
> Error:  /home/runner/work/skywalking-java/skywalking-java/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/PSSetterDefinitionOfJDBCInstrumentation.java:27:52: Using the '.*' form of import should be avoided - net.bytebuddy.matcher.ElementMatchers.*. [AvoidStarImport]
> Audit done.
> [INFO] There is 1 error reported by Checkstyle 9.3 with /home/runner/work/skywalking-java/skywalking-java/apm-checkstyle/checkStyle.xml ruleset.
> Error:  src/main/java/org/apache/skywalking/apm/plugin/jdbc/PSSetterDefinitionOfJDBCInstrumentation.java:[27,52] (imports) AvoidStarImport: Using the '.*' form of import should be avoided - net.bytebuddy.matcher.ElementMatchers.*.
> ```
> 
> It seems you don't check the code style locally. And you should update the changelog according to the PR template.

Sorry! Let me check it.

CHANGES.md
```diff
@@ -10,6 +10,7 @@ Release Notes.
 * Support clickhouse-jdbc-plugin trace sql parameters
 * Support monitor jetty server work thread pool metric
 * Support Jersey REST framework
+* Fix ClassCastException when SQLServer inserting data (#10741)
```
```suggestion
* Fix ClassCastException when SQLServer inserting data 
```

wu-sheng (approved):

PR #525, opened 2023-05-08
Chore: fix exclusion rules in maven-shade-plugin

lujiajing1126:
This PR fixes breaking exclusion rules for `org.checkerframework:checker-qual` and `com.google.j2objc:j2objc-annotations` which should have done excluded in the generated jar.

Besides, all proto files are excluded.

Current folder structure:

image


wu-sheng (approved):
LGTM

wu-sheng (approved):
LGTM

PR #526, opened 2023-05-08
fix jedis can not get host info when use jedis 3.3.x+ sdk

xu1009:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. https://github.com/apache/skywalking/issues/10768


apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-2.x-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v3/define/JedisInstrumentation.java
```diff
@@ -48,63 +38,17 @@ public ClassMatch enhanceClass() {
...
-                    }
-                }
-        };
+        return new ConstructorInterceptPoint[0];
     }
 
     @Override
     public InstanceMethodsInterceptPoint[] getInstanceMethodsInterceptPoints() {
-        return new InstanceMethodsInterceptPoint[] {
+        return new InstanceMethodsInterceptPoint[]{
```
wu-sheng, line 46: You seem not checking the style?

xu1009: maven build can success, i have fixed it

wu-sheng:
I think you need to add 3.3 to the versions of test scenarios.

apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-2.x-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v3/define/JedisInstrumentation.java
wu-sheng: I think this only fails in verify stage.

xu1009:
ok , i will add test from 3.3.0 use new jedis init way

test/plugin/scenarios/jedis-3.3.x-plus-scenario/src/main/java/org/apache/skywalking/apm/testcase/jedis/controller/JedisXreadparamAdaptor.java
```diff
@@ -0,0 +1,48 @@
...
+package org.apache.skywalking.apm.testcase.jedis.controller;
+
+import redis.clients.jedis.StreamEntryID;
+import java.lang.reflect.Method;
+import java.util.Map;
+
+public class JedisXreadparamAdaptor {
+
+    //  The stream apis have changed from 3.6.0
+    public static Object callXReadMethod(Object target, int count, Map<String, StreamEntryID> map) {
```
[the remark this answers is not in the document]
wu-sheng: Just due to keep test case runnable in [3.3.0, 3.6.0)?

apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-2.x-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v3/JedisMethodInterceptor.java
```diff
@@ -37,7 +39,13 @@ public class JedisMethodInterceptor implements InstanceMethodsAroundInterceptor
...
     public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
                              MethodInterceptResult result) throws Throwable {
-        String peer = String.valueOf(objInst.getSkyWalkingDynamicField());
+        String peer;
+        if (objInst instanceof Jedis) {
+            Client client = ((Jedis) objInst).getClient();
+            peer = client.getHost() + ":" + client.getPort();
+        } else {
+            peer = String.valueOf(objInst.getSkyWalkingDynamicField());
+        }
```
wu-sheng, line 48: Do you have to do through this way? This way is much slower in the runtime.
This interceptor was only executing codes relative to this redis client, but right now, it has to build peer again and again.
Considering accessing redis is super fast, this could be sensitive.
What do you think?

xu1009: get it，the peer can build in consturctor just when init，and set it to SkyWalkingDynamicField

wu-sheng (approved):

PR #527, opened 2023-05-08
Support agent running in the JBoss relative with its ModuleClassloader.

rechardguo:
### Fix https://github.com/apache/skywalking/issues/10770
  


lujiajing1126:
According to the https://github.com/apache/skywalking/issues/10770#issuecomment-1538276679

You should provide an E2E case to verify this behavior. For example, use JBoss as the servlet container.

wu-sheng:
> Not understand your point of test jboss in the CI.
Do you mean integrate tomcat and test, Could you give me some code or any guide document.

I don't have a JBoss env to test this. Generally, we don't rely on local testing. Most codes should be verified in the CI processes. Such as all plugins in this repository.
So, we need either JBoss to run in the plugin test or e2e test.

lujiajing1126:
Seems JBoss (Wildfly) is using LGPL v2.1 (https://www.apache.org/legal/resolved.html#category-x). Can we use this for testing purpose?

wu-sheng:
Yes, once it is open source, it is fine.

rechardguo:
Should be path seperator not set properly(use File.separator instead of hardcode /). I will change and verify in linux.

wu-sheng:
> Should be path seperator not set properly(use File.separator instead of hardcode /). I will change and verify in linux.

I don't think so. Our agent works on Window AFAIK.

apm-sniffer/apm-agent-test/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.apache.skywalking</groupId>
+        <artifactId>java-agent-sniffer</artifactId>
+        <version>8.16.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apm-agent-test</artifactId>
```
wu-sheng, line 28: What is this new module?

apm-sniffer/apm-agent-test/src/test/java/org/apache/skywalking/apm/agent/core/boot/AgentPackagePathTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Properties;
+
+/**
+ * this test case is to verify the fix of bug
+ * https://github.com/apache/skywalking/issues/10770
+ */
+public class AgentPackagePathTest {
+
+    @Test
+    public void testGetPathNotNull() throws ClassNotFoundException, IllegalAccessException, NoSuchMethodException, InvocationTargetException, IOException {
```
wu-sheng, line 39: I am confused about this UT. How could a UT verify class loader issue?

wu-sheng:
> This seperator is different between linux and window

They are definitely different. But I don't think it does matter. Otherwise, agent wouldn't work on Windows.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/boot/AgentPackagePath.java
```diff
@@ -50,6 +50,9 @@ private static File findPath() throws AgentPackageNotFoundException {
         String classResourcePath = AgentPackagePath.class.getName().replaceAll("\\.", "/") + ".class";
 
         URL resource = ClassLoader.getSystemClassLoader().getResource(classResourcePath);
+        if (resource == null) {
+            resource = AgentPackagePath.class.getClassLoader().getResource(classResourcePath);
+        }
```
wu-sheng, line 55:
```suggestion
        URL resource = AgentPackagePath.class.getClassLoader().getResource(classResourcePath);
```

I am wondering, is this enough? Why do we need a fail-back? I think in most cases, `AgentPackagePath` should be loaded by the expected classloader(appclassloader or moduleclassloader)

wu-sheng: I think it should be compatible. Please change and remove all your test codes. I could rerun the CI to verify this change.

rechardguo:
@wu-sheng 
How to update change logs?
Is it add 8.17.0 comment similar to 8.16.0
image

wu-sheng:
Add what you changed and fixed as a sentence.

wu-sheng:
Currently, it is 8.16.0. You just need to add a line of change description here. 
Once 8.16.0 is released, this file would be updated to 8.17.0 by the release manager. You don't need to bother about this.

CHANGES.md
```diff
@@ -14,6 +14,7 @@ Release Notes.
 * [Chore] Exclude org.checkerframework:checker-qual and com.google.j2objc:j2objc-annotations
 * [Chore] Exclude proto files in the generated jar
 * Fix Jedis-2.x plugin can not get host info in jedis 3.3.x+
+* Change AgentPackagePath#findPath() to get AgentPackagePath url by AgentPackagePath.class.getClassLoader()
```
wu-sheng, line 17:
```suggestion
* Change the classloader to locate the agent path in AgentPackagePath, from `SystemClassLoader` to AgentPackagePath's loader.
```

rechardguo:
@wu-sheng Thanks for your patient. I remove test case and update change log(not sure if properly)

wu-sheng (approved):
LGTM

PR #528, opened 2023-05-12
Support grizzly web server trace

xu1009:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. https://github.com/apache/skywalking/issues/10767


xu1009:
plan to support the grizzly thread pool metric plugin in free time

apm-sniffer/apm-sdk-plugin/grizzly-2.3.x-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/grizzly/v2/HttpServiceInterceptor.java
```diff
@@ -0,0 +1,61 @@
...
+        AbstractSpan span = ContextManager.createLocalSpan("GrizzlyRunService");
+        span.setComponent(ComponentsDefine.GRIZZLY);
+        ContextManager.continued(contextSnapshot);
+
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        Object[] context = (Object[]) objInst.getSkyWalkingDynamicField();
+        AbstractSpan abstractSpan = (AbstractSpan) context[0];
```
wu-sheng, line 46: Is the target class created for every request? Otherwise, you will face concurrency issue, as the dynamic field is shared.
The MethodInvocationContext is the only safe place to share between before and after for every method.

xu1009: get it, i will put it to request scope

wu-sheng (approved):
LGTM

wu-sheng (approved):
LGTM

PR #531, opened 2023-05-17
Fix possible IllegalStateException when using Micrometer

wapkch:
### Fix <possible IllegalStateException when using Micrometer>
- [x] Micrometer SpanHelper#tryToGetPeer may return null in some cases(e.g. OpenFeign), which produces a IllegalStateException when try to create span. See [2040](https://github.com/OpenFeign/feign/issues/2040), [20](https://github.com/jdbc-observations/datasource-micrometer/issues/20). So i try to add remoteName and URI path to that logic. @marcingrzejszczak Please help me review it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Release Notes.
 * Fix Jedis-2.x plugin can not get host info in jedis 3.3.x+
 * Change the classloader to locate the agent path in AgentPackagePath, from `SystemClassLoader` to AgentPackagePath's loader.
 * Support Grizzly Trace
+* Fix possible IllegalStateException when using Micrometer [#531](https://github.com/apache/skywalking-java/pull/531)
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,7 +16,7 @@ Release Notes.
 * Fix Jedis-2.x plugin can not get host info in jedis 3.3.x+
 * Change the classloader to locate the agent path in AgentPackagePath, from `SystemClassLoader` to AgentPackagePath's loader.
 * Support Grizzly Trace
-* Fix possible IllegalStateException when using Micrometer [#531](https://github.com/apache/skywalking-java/pull/531)
+* Fix possible IllegalStateException when using Micrometer.
 
 #### Documentation
 
--- a/apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/MicrometerReceiverTracingHandlerInterceptor.java
+++ b/apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/MicrometerReceiverTracingHandlerInterceptor.java
@@ -60,7 +60,8 @@ public class MicrometerReceiverTracingHandlerInterceptor implements InstanceMeth
             SpanLayer spanLayer = TaggingHelper.toLayer(context.getAllKeyValues());
             AbstractSpan abstractSpan = ContextManager.activeSpan();
             abstractSpan
-                .setPeer(SpanHelper.tryToGetPeer(context.getRemoteServiceAddress(), context.getAllKeyValues()))
+                .setPeer(SpanHelper.tryToGetPeer(context.getRemoteServiceAddress(), context.getRemoteServiceName(),
+                    context.getAllKeyValues()))
                 .setOperationName(getOperationName(context));
             context.getAllKeyValues()
                    .forEach(keyValue -> abstractSpan.tag(Tags.ofKey(keyValue.getKey()), keyValue.getValue()));
--- a/apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/MicrometerReceiverTracingHandlerInterceptor.java
+++ b/apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/MicrometerReceiverTracingHandlerInterceptor.java
@@ -60,6 +60,8 @@ public class MicrometerReceiverTracingHandlerInterceptor implements InstanceMeth
             SpanLayer spanLayer = TaggingHelper.toLayer(context.getAllKeyValues());
             AbstractSpan abstractSpan = ContextManager.activeSpan();
             abstractSpan
+                .setPeer(SpanHelper.tryToGetPeer(context.getRemoteServiceAddress(), context.getRemoteServiceName(),
+                    context.getAllKeyValues()))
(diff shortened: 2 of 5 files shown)


wu-sheng:
Please add changes.md in the root about this PR.

wu-sheng:
And your change seems to break the    Test / resttemplate-6.x case.
Please recheck.

CHANGES.md
```diff
@@ -16,6 +16,7 @@ Release Notes.
 * Fix Jedis-2.x plugin can not get host info in jedis 3.3.x+
 * Change the classloader to locate the agent path in AgentPackagePath, from `SystemClassLoader` to AgentPackagePath's loader.
 * Support Grizzly Trace
+* Fix possible IllegalStateException when using Micrometer [#531](https://github.com/apache/skywalking-java/pull/531)
```
wu-sheng, line 19:
```suggestion
* Fix possible IllegalStateException when using Micrometer.
```

apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/MicrometerReceiverTracingHandlerInterceptor.java
```diff
@@ -60,7 +60,6 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
             SpanLayer spanLayer = TaggingHelper.toLayer(context.getAllKeyValues());
             AbstractSpan abstractSpan = ContextManager.activeSpan();
             abstractSpan
-                .setPeer(SpanHelper.tryToGetPeer(context.getRemoteServiceAddress(), context.getAllKeyValues()))
```
marcingrzejszczak, line 63: We don't need peer on the receiver side?

apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/SpanHelper.java
```diff
@@ -39,10 +42,16 @@ static String tryToGetPeer(String remoteAddress, KeyValues allKeyValues) {
                                .orElse("unknown");
         try {
             URI uri = URI.create(result);
-            if (uri.getHost() == null) {
-                return null;
+            if (uri.getHost() != null) {
+                return uri.getHost() + ":" + uri.getPort();
+            }
+            if (!StringUtil.isEmpty(remoteName)) {
```
marcingrzejszczak, line 48: Now I'm wondering if we shouldn't do the same remoteName check if there was an exception. Cause if there was an exception when doing `URI.create()` then maybe we can fallback to remote name?

apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/MicrometerReceiverTracingHandlerInterceptor.java
wu-sheng: It is optional in most cases, and required in MQ consumer and producer cases.

wu-sheng: @wapkch I think this should be kept. Why do you remove this?

wapkch: Because it breaks the Test / resttemplate-6.x case on entry span, which expects an empty peer, but the peer from `tryToGetPeer` here will be `/resttemplate-6.x-scenario/resttemplate/syncback` now, it seems meaningless in this case. And now i realize that the peer is required in MQ consumer and producer, but i'm not sure if i should keep this and modify the peer of entry span in Test / resttemplate-6.x case to `/resttemplate-6.x-scenario/resttemplate/syncback`?

apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/SpanHelper.java
wapkch: It seems reasonable.

apm-sniffer/apm-toolkit-activation/apm-toolkit-micrometer-activation/src/main/java/org/apache/skywalking/apm/toolkit/activation/micrometer/MicrometerReceiverTracingHandlerInterceptor.java
wu-sheng: Why the peer is a URI? It should be IP and port only. Peer should be a network socket address only.

wapkch: Because IP and port can not retrieved from Micrometer ObservationContext in some cases, but a URI can be easily found from the keyValues in it. So i think we may should distinguish receiver side to MQ receiver and other receivers because MQ receiver need a peer and others do not. cc @marcingrzejszczak

wu-sheng: If you use URI, the IP:PORT mapping data in the backend would easily burn out and crash the OAP due to OOM.
[STAM paper](https://skywalking.apache.org/docs/main/next/en/papers/stam/) introduced why this is required to build topology in pure streaming rather than a time window oriented.

wapkch: I see. I will try to find another way.

wapkch: I add back peer on the receiver side and remove the uri path logic in tryToGetPee method.

wu-sheng (approved):

PR #536, opened 2023-05-26
add RocketMQ 5.x plugin

weihubeats:
- [x] https://github.com/apache/skywalking/issues/10858


wu-sheng:
The plugin-list.md and supported-list.md should be updated as well.

wu-sheng:
As this is a new plugin, please update the existing RocketMQ scenarios or add a new test scenario.

wu-sheng:
Ref to the existing test case. https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/rocketmq-scenario

If you don't know what is the plugin test, read this, https://skywalking.apache.org/docs/skywalking-java/next/en/setup/service-agent/java-agent/plugin-test/

wu-sheng:
Please make sure the tests are added and run through CI.

wu-sheng:
First of all, are you just copying the old cases? What is the difference besides versions listed?

weihubeats:
Modified the version of rocketmq client

wu-sheng:
Just changed the `support-version.list` in the existing RocketMQ case. If no code changes, this means from client side perspective, this 5.x is not a breaking change. Then `support-version.list` would use all listed versions to compile, from 4.x to 5.x
Have you run this test locally?

weihubeats:
The code of rocketmq consumer and producer didn't change much, but some packages of incremental classes changed their paths, I tested the correctness of the plugin locally, but I encountered a rather strange problem with the scenario test, running locally the main part of the rocketmq scenario test also reported an error
My Execution Script：

wu-sheng:
This error usually means, yout test codes are not working as server is not up.

You need to add new scenario(if it is different) to CI control files, we would see the results of it.

CHANGES.md
```diff
@@ -19,6 +19,7 @@ Release Notes.
 * Fix possible IllegalStateException when using Micrometer.
 * Support Grizzly Work ThreadPool Metric Monitor
 * Fix the gson dependency in the kafka-reporter-plugin.
+* add add Rocketm 5.x plugin
```
Superskyyy, line 22:
```suggestion
* Add Rocketmq 5.x plugin
```

wu-sheng:
Your test is falling, that is what I was saying.

weihubeats:
I understand you mean I don't need to add a new test module, I just need to add support for rocketmq5.x to the support-version.list of the original `rocketmq-scenario` test module, is that right?Because the test code remains unchanged.Because the test code remains unchanged (the test code is just an upgrade of the `rocketmq-client` version)

wu-sheng:
Yes, if all codes are unchanged, just need a new dependency version to recompile, you should not write everything again.
But more importantly, the test should pass. From what I see, it doesn't. This is irrelevant about new test scenario or existing test.

wu-sheng:
The tests with the new version added, are failing, as it was expected same as your new added separate tests failed. Please fix them

weihubeats:
Okay, let me see, the test error log is a bit strange
image

wu-sheng:
That is not a thing I can resolve.

wu-sheng:
One guess, check the server version. Is the new client failing to connect old server?

weihubeats:
I have tested locally using `client4.x` related client and `5.x` related client, both can connect to 5.x nameserve, using also the test project `rocketmq-scenario`
image
image

wu-sheng:
But the test case was using 4.x server, have you changed that?

weihubeats:
I thought the `${test.framework.version}` parameter and `${CASE_SERVER_IMAGE_VERSION}` were getting the version inside the `support-version.list` file by default, so they wouldn't have a problem, so `${CASE_SERVER_IMAGE_ VERSION} `where does the version of the docker image come from? How do I need to change it?Sorry I'm not very familiar with the integration testing piece

wu-sheng:
> Ref to the existing test case. https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/rocketmq-scenario
> 
> If you don't know what is the plugin test, read this, https://skywalking.apache.org/docs/skywalking-java/next/en/setup/service-agent/java-agent/plugin-test/

That is why I ref the doc to you at the beginning. Don't get what it is. Read the doc and run them locally.

wu-sheng:
Why do you only remove 5.1.1? Is that anything special?

weihubeats:
image
I found that the official docker image of rocketmq only supports ppc64le, which is the reason for my ci failure, is there any good way to handle it here? Or can I just go to rocketmq official issues and ask them to provide amd images?

wu-sheng:
Why don't they release x86/amd64 image?

wu-sheng:
Especially, they have latest release for amd64.

weihubeats:
image

weihubeats:
> Why don't they release x86/amd64 image?

I don't know.

wu-sheng:
You could open a ticket to ask. I think something missed.

weihubeats:
Okay, I'll go to rocketmq and raise an issues
https://github.com/apache/rocketmq/issues/6847

wu-sheng (approved):
@weihubeats As the test issue has been fixed by RocketMQ upstream, it works now.

Anything to add?

weihubeats:
no

PR #538, opened 2023-05-26
Fix the gson dependency in the kafka-reporter-plugin. (#10848)

gzlicanyi:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/10848.

### DIFF
--- a/apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
+++ b/apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
@@ -28,11 +28,20 @@
     <artifactId>kafka-reporter-plugin</artifactId>
     <packaging>jar</packaging>
 
+    <properties>
+        <shade.com.google.source>com.google</shade.com.google.source>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
             <version>${kafka-clients.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>${gson.version}</version>
         </dependency>
     </dependencies>
 
@@ -75,6 +84,10 @@
                                 <relocation>
                                     <pattern>org.slf4j</pattern>
                                     <shadedPattern>${shade.package}/org.slf4j</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>${shade.com.google.source}</pattern>
+                                    <shadedPattern>${shade.package}.${shade.com.google.source}</shadedPattern>
                                 </relocation>
                             </relocations>
                             <transformers>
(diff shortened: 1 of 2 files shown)


apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
```diff
@@ -28,12 +28,21 @@
...
+        <shade.com.google.source>com.google</shade.com.google.source>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
             <version>${kafka-clients.version}</version>
         </dependency>
+        <dependency>
```
lujiajing1126, line 41: I don't think we need to explicitly add this dependency since we've already include agent-core

lujiajing1126 (changes requested):

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
gzlicanyi: This is old code. Similar to Gson, the kafka-reporter-plugin does not use any dependencies from apm-agent-core.

lujiajing1126: > This is old code. Similar to Gson, the kafka-reporter-plugin does not use any dependencies from apm-agent-core.

IIRC, the shaded Gson library is already located in the agent-core jar. So just shading this library at compile-time is sufficient.

wu-sheng: > This is old code.

This is something you added, this is not old code.

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
```diff
@@ -28,12 +28,21 @@
     <artifactId>kafka-reporter-plugin</artifactId>
     <packaging>jar</packaging>
 
+    <properties>
+        <shade.com.google.source>com.google</shade.com.google.source>
```
wu-sheng, line 32: We don't need this. The property name is even longer than the original package name, there is no point.

wu-sheng: And, I don't think we are expecting shading all `com.google`. You are just going to shade GSON relative things. Please be clear about that.

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
gzlicanyi: > > This is old code. Similar to Gson, the kafka-reporter-plugin does not use any dependencies from apm-agent-core.
> 
> IIRC, the shaded Gson library is already located in the agent-core jar. So just shading this library at compile-time is sufficient.
> 
> Just like what we've done for slf4j.

You are right, I will give it a try.

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
gzlicanyi: > And, I don't think we are expecting shading all `com.google`. You are just going to shade GSON relative things. Please be clear about that.

Let me optimize this.

wu-sheng (approved):
This should be good. @lujiajing1126 It is better you could check locally.

lujiajing1126 (approved):
LGTM. I've compiled locally and everything is fine.

PR #540, opened 2023-05-27
[Feature] Support to configure custom decode method for kafka configurations

GuiSong01:
### Support to configure custom decryption methods for kafka configurations

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #[10815](https://github.com/apache/skywalking/issues/10815).


wu-sheng:
I remember there was a discussion about this, please link to the issue.

GuiSong01:
> I remember there was a discussion about this, please link to the issue.

[discussion ](https://github.com/apache/skywalking/discussions/10847) is a bug `kafka-reporter-plugin-8.15.0 throw exception "java.lang.NoClassDefFoundError: com/google/gson/Gson"`,it is fixed by [Fix the gson dependency in the kafka-reporter-plugin](https://github.com/apache/skywalking-java/pull/538).The PR is only about the feature [skywalking-java support to decrypt password of kafka with SASL_SSL enabled](https://github.com/apache/skywalking/issues/10815)

apm-sniffer/config/agent.config
```diff
@@ -263,6 +263,10 @@ plugin.kafka.topic_management=${SW_PLUGIN_KAFKA_TOPIC_MANAGEMENT:skywalking-mana
 plugin.kafka.topic_logging=${SW_PLUGIN_KAFKA_TOPIC_LOGGING:skywalking-logs}
 #  isolate multi OAP server when using same Kafka cluster (final topic name will append namespace before Kafka topics with `-` ).
 plugin.kafka.namespace=${SW_KAFKA_NAMESPACE:}
+# Specify which class to decrypt encrypted configuration of kafka.You can set encrypted information in `plugin.kafka.producer_config_json` or `plugin.kafka.producer_config` if you need.
+plugin.kafka.decrypt_class=${SW_KAFKA_DECRYPT_CLASS:}
+# Specify which mathed to decrypt encrypted configuration of kafka.
```
wu-sheng, line 268:
```suggestion
# Specify which method to decrypt encrypted configuration of kafka.
```

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/src/main/java/org/apache/skywalking/apm/agent/core/kafka/KafkaProducerManager.java
```diff
@@ -153,6 +156,18 @@ private void notifyListeners(KafkaConnectionStatus status) {
         }
     }
 
+    private Map<String, String> decrypt(Map<String, String> config) {
+        try {
+            Class<?> decryptClazz = Class.forName(Kafka.DECRYPT_CLASS);
+            Method decryptMethod = decryptClazz.getMethod(Kafka.DECRYPT_METHOD, Map.class);
+            return (Map<String, String>) decryptMethod.invoke(decryptClazz.newInstance(), config);
```
wu-sheng, line 163: I don't think reflection for the whole process is a good idea. You should add an interface in the reporter project,

```java
public interface KafkaConfigExtension {
    Map<String, String> decode(Map<String, String> config);
}
```

Then this interface would be released, and user could simply depend on and package. Reflection should be something that has to be that way.

Notice, your saying `decrypt` actually is just `decode`. Because decrypting usually means you have in-built algorithms.

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/src/main/java/org/apache/skywalking/apm/agent/core/kafka/KafkaProducerManager.java
```diff
@@ -153,6 +156,18 @@ private void notifyListeners(KafkaConnectionStatus status) {
...
+        } catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e) {
+            // ignore
+            LOGGER.warn("The decrypt class {} is not exist, exception:{}.", Kafka.DECRYPT_CLASS, e);
```
wu-sheng, line 166: And, you are reflecting the class without checking whether this config existence, which mean , this warning will be there as always. AFAIK.

Although I have suggested removing most of the reflection, still, you will still need reflections to create interface implementation. You should check the class existence before invoking.

wu-sheng (changes requested):

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/src/main/java/org/apache/skywalking/apm/agent/core/kafka/KafkaProducerManager.java
```diff
@@ -153,6 +155,22 @@ private void notifyListeners(KafkaConnectionStatus status) {
...
+            return config;
+        }
+        try {
+            Object decodeTool = Class.forName(Kafka.DECODE_CLASS).getDeclaredConstructor().newInstance();
+            if (decodeTool instanceof KafkaConfigExtension) {
+                return ((KafkaConfigExtension) decodeTool).decode(config);
+            }
+        } catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e) {
+            // ignore
+            LOGGER.warn("The decode class {} is not exist, exception:{}.", Kafka.DECODE_CLASS, e);
```
wu-sheng, line 169:
```suggestion
            LOGGER.warn("The decode class {} does not exist, exception:{}.", Kafka.DECODE_CLASS, e);
```

wu-sheng (approved):
@GuiSong01 I rewrote the docs for you. Others are good for me.

Sorry for letting you wait for so long, take a long time to recovery from jetleg

PR #541, opened 2023-05-30
Support jdk17 zgc

xu1009:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. https://github.com/apache/skywalking/issues/10870


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/GCModule.java
```diff
@@ -88,4 +98,12 @@ public List<GC> getGCList() {
...
 
     protected abstract String getNormalGCName();
+
+    protected String getCycleGcBeanName() {
+        return null;
+    }
+
+    protected String getPauseGcBeanName() {
+        return null;
+    }
```
wu-sheng, line 108: I don't feel this is a good way. Couldn't you use an implementation to override `getNormalGCName`?

xu1009: the zgc gc count needs be collected from ZGC Cycles bean
the zgc gc pause time needs be collected from  ZGC Pauses  bean

so the two bean name should be in one gc module and then get  gc count and gc pause time from two bean  when collect

wu-sheng: I mean we should have zgc version detection. Right now, keeping adding new methods seems not a good code practice.

xu1009: add a new zgcmodule for jdk 17 and keep ZGC Cycles and  ZGC Pauses bean name in this module, when it is jdk17, set this module ？

wu-sheng: I don't have implementation preference as I don't know details enough.
Generally, don't add new methods for GCModule.

xu1009: gc time and count is calculated in GCModule.  now the GCModule method getOldGCName, getNewGCName, getNormalGCName is for get bean name and collect metirc from the target bean.

Generally zgc just need one bean(bean name from getNormalGCName method) to collect the gc time and gc count

but jdk17 zgc need two beans to calculate，i can not use other three exists method get the bean.

now i have two ways when  don't add new methods for GCModule

1. in GCModule, use string constant bean name not use method get it，i think it is strange

2. zgcmodule overwrite GCModule caluldate method and implementation itself. because zgc is  different with other gc. the GCModule  caluldate method is just universal for non zgc.

wu-sheng: My point is, can't you build another two new `ZGCModule` implementations in `GCProvider`? And those two return new values of `getNormalGCName`

xu1009: it can not, if you add two new ZGCModule，in current project, just one can be active，so in GCModule, getNormalGCName just get one bean ZGC Cycles or ZGC Pauses, but gc count is from ZGC Cycles and gc pause time is from ZGC Pauses, i think it is not working

wu-sheng: Then please override `public List<GC> getGCList` from `GCMetricAccessor` to create a new `ZGCModule`, then to build a new `List<GC>`. `lastNormalGCCount` and `lastNormalGCTime` could be simply removed as well.
That was why it was built in two layers of the abstract concept.

CHANGES.md
```diff
@@ -5,7 +5,7 @@ Release Notes.
 8.17.0
 ------------------
 
-
+* Support Jdk17 ZGC metric collect
```
wu-sheng, line 8:
```suggestion
* Support Jdk17 ZGC metric collect

```

wu-sheng (approved):

wu-sheng:
@xu1009 As you have contributed dozens of features in Java agent, I think you are getting familiar with things here.
If you have interests to join committer team, and takes some responsibilities on PR review and release, ping me on slack.

wu-sheng:
And of course, you could keep it as a contributor, who only contributes codes when you want and have time. 
Don't feel pressured to say NO.

xu1009:
@wu-sheng i am very glad to join committer team, skywalking is very useful in our team,i am alson continue to contribute,

wu-sheng:
> @wu-sheng i am very glad to join committer team, skywalking is very useful in our team,i am alson continue to contribute,
> 
> 

Please join the slack, I need some information to start the process. Thanks.

xu1009:
i have joined slack, i am new to it, what shoul i do next

wu-sheng:
Could you ping me on the `#skywalking` channel? We are there.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/gc/ZGCModule.java
```diff
@@ -18,27 +18,52 @@
...
-        return null;
-    }
+                long time = bean.getCollectionTime();
+                gcTime = time - lastNormalGCTime;
+                lastNormalGCTime = time;
+            } else if (name.equals("ZGC Cycles")) {
+                long collectionCount = bean.getCollectionCount();
+                gcCount = collectionCount - lastNormalGCCount;
+                lastNormalGCCount = collectionCount;
+            } else if (name.equals(" ZGC Pauses")) {
```
kezhenxu94, line 57: Is this space ` ZGC Pauses` intentional?

wu-sheng: @xu1009 Is this a bug?

xu1009: yes, i will fix it

PR #542, opened 2023-05-31
Fix deserialization of kafka producer json config in the kafka-reporter-plugin.

gzlicanyi:
When the value of a JSON config in Kafka is a non-string type, such as {"batch.size":32768}, an error will be thrown when setting properties with the generated map.
```
ERROR 2023-05-31 14:56:57.847 SkywalkingAgent-13-org.apache.skywalking.apm.dependencies.kafkaProducerInitThread-0 KafkaProducerManager : unexpected exception. 
java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.String
        at java.util.Map.forEach(Map.java:630)
        at org.apache.skywalking.apm.agent.core.kafka.KafkaProducerManager.run(KafkaProducerManager.java:111)
```
Because this code does not deserialize the value of JSON into a String type. However, the value required by Properties is of String type.
```java
Map<String, String> config = (Map<String, String>) gson.fromJson(Kafka.PRODUCER_CONFIG_JSON, Map.class);
```

### DIFF
--- a/apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
+++ b/apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
@@ -80,6 +80,10 @@
                                     <pattern>com.google.gson</pattern>
                                     <shadedPattern>${shade.package}.com.google.gson</shadedPattern>
                                 </relocation>
+                                <relocation>
+                                    <pattern>com.google.gson</pattern>
+                                    <shadedPattern>${shade.package}.com.google.common</shadedPattern>
+                                </relocation>
                             </relocations>
                             <transformers>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
--- a/apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
+++ b/apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
@@ -28,11 +28,20 @@
     <artifactId>kafka-reporter-plugin</artifactId>
     <packaging>jar</packaging>
 
+    <properties>
+        <shade.com.google.source>com.google</shade.com.google.source>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
             <version>${kafka-clients.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>${gson.version}</version>
         </dependency>
     </dependencies>
 
@@ -75,6 +84,10 @@
                                 <relocation>
                                     <pattern>org.slf4j</pattern>
                                     <shadedPattern>${shade.package}/org.slf4j</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>${shade.com.google.source}</pattern>
+                                    <shadedPattern>${shade.package}.${shade.com.google.source}</shadedPattern>
                                 </relocation>
                             </relocations>
                             <transformers>
--- a/apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
+++ b/apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
@@ -28,20 +28,11 @@
     <artifactId>kafka-reporter-plugin</artifactId>
     <packaging>jar</packaging>
 
-    <properties>
-        <shade.com.google.source>com.google</shade.com.google.source>
-    </properties>
-
     <dependencies>
         <dependency>
             <groupId>org.apache.kafka</groupId>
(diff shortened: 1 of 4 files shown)


wu-sheng:
I think this is duplicated with #540?

gzlicanyi:
> 

#540 hasn't fixed this bug.

wu-sheng:
Didn't you have fixed? https://github.com/apache/skywalking-java/pull/538 I am confused.

gzlicanyi:
> Didn't you have fixed? #538 I am confused.

It's not the same issue. #538 fixed the Gson dependency issue, while #542 fixed the Gson deserialization problem, but both addressed the Kafka JSON config issue.

gzlicanyi:
> Didn't you have fixed? #538 I am confused.

When the value of a JSON config in Kafka is a non-string type, such as {"batch.size":32768}, an error will be thrown when setting properties with the generated map.

But such as {"batch.size":"32768"}, it is ok.

wu-sheng:
Could you provide more details? I am not familiar with this kind of Kafka configuration.

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/src/test/java/org/apache/skywalking/apm/agent/core/kafka/KafkaProducerManagerTest.java
```diff
@@ -54,6 +55,17 @@ public void testFormatTopicNameThenRegister() {
...
+    public void testSetPropertiesFromJsonConfig() {
+        KafkaProducerManager kafkaProducerManager = new KafkaProducerManager();
+        Properties properties = new Properties();
+
+        KafkaReporterPluginConfig.Plugin.Kafka.PRODUCER_CONFIG_JSON = "{\"batch.size\":32768}";
+        kafkaProducerManager.setPropertiesFromJsonConfig(properties);
+
+        assertEquals(properties.get("batch.size"), "32768");
+    }
+
```
wu-sheng, line 68: What does this UT prove? The codes are not shaded ever in the UT scope, right?

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
```diff
@@ -80,6 +80,10 @@
                                     <pattern>com.google.gson</pattern>
                                     <shadedPattern>${shade.package}.com.google.gson</shadedPattern>
                                 </relocation>
+                                <relocation>
+                                    <pattern>com.google.gson</pattern>
+                                    <shadedPattern>${shade.package}.com.google.common</shadedPattern>
+                                </relocation>
```
wu-sheng, line 86: `com.google.gson` -> `com.google.common`?

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/src/test/java/org/apache/skywalking/apm/agent/core/kafka/KafkaProducerManagerTest.java
gzlicanyi: Show an example of how a JSON configuration can be correctly set to properties.

gzlicanyi: > What does this UT prove? The codes are not shaded ever in the UT scope, right?

right.

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/pom.xml
gzlicanyi: fixed

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/src/test/java/org/apache/skywalking/apm/agent/core/kafka/KafkaProducerManagerTest.java
wu-sheng: So, the point of this PR is `TypeToken`, which cast the type accordingly.

But still `import com.google.common.reflect.TypeToken;` seems not from gson package. Why?

wu-sheng:
Please fix the conflicts as another PR gets merged.

apm-sniffer/optional-reporter-plugins/kafka-reporter-plugin/src/test/java/org/apache/skywalking/apm/agent/core/kafka/KafkaProducerManagerTest.java
gzlicanyi: > So, the point of this PR is `TypeToken`, which cast the type accordingly.
> 
> But still `import com.google.common.reflect.TypeToken;` seems not from gson package. Why?

I have used the wrong TypeToken out of two.

wu-sheng (approved):

PR #546, opened 2023-06-06
Feature/fix spring data hadoop

gzlicanyi:

#### problem
When using HbaseTemplate, the following error occurs:
```
ERROR 2023-06-05 17:06:37.757 http-nio-0.0.0.0-80-exec-3 InstMethodsInter : class[class org.apache.hadoop.hbase.client.HTable] before method[getScanner] intercept failure 
java.lang.IllegalStateException: Exit span doesn't include meaningful peer information.
        at org.apache.skywalking.apm.agent.core.context.TracingContext.inject(TracingContext.java:169)
        at org.apache.skywalking.apm.agent.core.context.TracingContext.inject(TracingContext.java:149)
        at org.apache.skywalking.apm.agent.core.context.ContextManager.createExitSpan(ContextManager.java:134)
        at org.apache.skywalking.apm.plugin.hbase.HTableInterceptor.beforeMethod(HTableInterceptor.java:72)
        at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:76)
        at org.apache.hadoop.hbase.client.HTable.getScanner(HTable.java)
        at org.springframework.data.hadoop.hbase.HbaseTemplate$1.doInTable(HbaseTemplate.java:132)
        at org.springframework.data.hadoop.hbase.HbaseTemplate.execute(HbaseTemplate.java:61)
        at org.springframework.data.hadoop.hbase.HbaseTemplate.find(HbaseTemplate.java:129)
        at org.springframework.data.hadoop.hbase.HbaseTemplate.find(HbaseTemplate.java:158)
```

#### Steps to reproduce

1. Using the following dependencies:
```xml
<dependency>
            <groupId>org.apache.hbase</groupId>
            <artifactId>hbase-client</artifactId>
            <version>1.6.0</version>
        </dependency>
        <dependency>
            <groupId>org.springframework.data</groupId>
            <artifactId>spring-data-hadoop</artifactId>
            <version>2.5.0.RELEASE</version>
        </dependency>
```

2. spring bean config
```xml
<hdp:configuration id="hadoopConfiguration"></hdp:configuration>
    <hdp:hbase-configuration delete-connection="true" stop-proxy="false">
        hbase.zookeeper.quorum=${hbase.zookeeper.quorum}
        hbase.zookeeper.property.clientPort=${hbase.zookeeper.port}
    </hdp:hbase-configuration>

    <bean id="hbaseTemplate" class="org.springframework.data.hadoop.hbase.HbaseTemplate">
        <property name="configuration" ref="hbaseConfiguration" />
    </bean>
```

3. using HbaseTemplate
```java
hbaseTemplate.find(String tableName, final Scan scan, final RowMapper<T> action);
```

#### Causes
HbaseTemplate uses HbaseUtils to create [HTable](https://github.com/apache/hbase/blob/rel/1.6.0/hbase-client/src/main/java/org/apache/hadoop/hbase/client/HTable.java), and the constructor used by HTable is `public HTable(Configuration conf, final TableName tableName)`, which is not byte code enhanced.

`HbaseUtils.getHTable` source code:
```java
public static HTableInterface getHTable(String tableName, Configuration configuration, Charset charset, HTableInterfaceFactory tableFactory) {
		if (HbaseSynchronizationManager.hasResource(tableName)) {
			return (HTable) HbaseSynchronizationManager.getResource(tableName);
		}

		HTableInterface t = null;
		try {
			if (tableFactory != null) {
				t = tableFactory.createHTableInterface(configuration, tableName.getBytes(charset));
			}
			else {
				t = new HTable(configuration, tableName.getBytes(charset));
			}

			return t;

		} catch (Exception ex) {
			throw convertHbaseException(ex);
		}
	}
```

#### how to fix it
Add byte code enhancement for `public HTable(Configuration conf, final TableName tableName)`.


wu-sheng (approved):
LGTM

PR #555, opened 2023-06-16
Fix the thread safety bug of finishing operation for the span named "SpringCloudGateway/sendRequest"

gzlicanyi:

### problem
```
java.lang.RuntimeException: Can not do async finish for the span repeatedly.
        at org.apache.skywalking.apm.agent.core.context.trace.AbstractTracingSpan.asyncFinish(AbstractTracingSpan.java:332)
        at org.apache.skywalking.apm.plugin.spring.cloud.gateway.v20x.HttpClientRequestInterceptor$2.accept(HttpClientRequestInterceptor.java:106)
        at org.apache.skywalking.apm.plugin.spring.cloud.gateway.v20x.HttpClientRequestInterceptor$2.accept(HttpClientRequestInterceptor.java:94)
```
In addition, there is another question：
A NullPointerException is thrown when executing enhanceCacheObject.getSendSpan() in the afterMethod() method of the HttpClientRequestInterceptor class.

### Steps to reproduce
Using the following dependencies:
```xml
        <dependency>
            <groupId>org.springframework.cloud</groupId>
            <artifactId>spring-cloud-starter-gateway</artifactId>
            <version>2.0.4.RELEASE</version>
        </dependency>
```

Using the WebClient in custom GlobalFilter.

### Causes

1. The HttpClient instance is a singleton, and the setting and retrieval of EnhanceCacheObject are not thread-safe. Therefore, the EnhanceCacheObject obtained in the afterMethod may not necessarily be set by the same thread in the beforeMethod.

2. HttpClient is used in WebClient scenarios, rather than in NettyRoutingFilter scenarios, where ContextManager is inactive. So, enhanceCacheObject may be null.

3. `responseMono.doAfterSuccessOrError` method may be executed multiple times.

4. If an exception is thrown inside the accept method of responseMono.doAfterSuccessOrError, it may disrupt the ongoing business process.

### how to fix it
1. Store the EnhanceCacheObject in a ThreadLocal.
2. To enhance code robustness


apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.0.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v20x/HttpClientRequestInterceptor.java
```diff
@@ -89,27 +97,45 @@ public Object afterMethod(final EnhancedInstance objInst,
...
+            Tags.HTTP_RESPONSE_STATUS_CODE.set(abstractSpan, httpClientResponse.status().code());
+
+            abstractSpan.asyncFinish();
+            enhanceCacheObject.getFilterSpan().asyncFinish();
+
+            enhanceCacheObject.setSpanFinish(true);
+        } catch (Throwable e) {
+            //Catch unknown exceptions to avoid interrupting business processes.
+            LOGGER.error("Failed to finish the Span of reactor.ipc.netty.http.client.HttpClient.", e);
+        }
```
wu-sheng, line 138: I don't think the interceptor doesn't need to print a log. What is special here?

gzlicanyi: Used to discover if there are any other unknown exceptions.

gzlicanyi: Of course, I can remove it as well.

apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.0.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v20x/HttpClientRequestInterceptor.java
```diff
@@ -80,7 +84,7 @@ public Publisher<Void> apply(final HttpClientRequest httpClientRequest) {
             }
         };
 
-        objInst.setSkyWalkingDynamicField(new EnhanceCacheObject(span, abstractSpan));
+        ENHANCE_CONTEXT.set(new EnhanceCacheObject(span, abstractSpan));
```
wuwen5, line 87: > The HttpClient instance is a singleton, and the setting and retrieval of EnhanceCacheObject are not thread-safe. Therefore, the EnhanceCacheObject obtained in the afterMethod may not necessarily be set by the same thread in the beforeMethod.

In fact, this description is not correct, no need for threadlocal here

gzlicanyi: In the beforeMethod method, the name of the thread is passed to EnhanceCacheObject, but in the afterMethod method, when it is retrieved and compared to the current thread, they are not consistent, and the URL is also not consistent.

gzlicanyi: step:
1. thread1 -> beforeMethod -> set EnhanceCacheObject1
2. thread2 -> beforeMethod -> set EnhanceCacheObject2
3. thread1 -> afterMethod -> get EnhanceCacheObject2
4. thread2 -> afterMethod -> get EnhanceCacheObject2

gzlicanyi: Reproduction steps：
In a high-concurrency scenario, the allArguments[1] is stored in EnhanceCacheObject in the beforeMethod method, but in the afterMethod method, the allArguments[1] of the method parameter is different from the allArguments[1] in EnhanceCacheObject.
Similarly, the thread name can also be reproduced.

apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.0.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v20x/HttpClientRequestInterceptor.java
wu-sheng: The error would be printed at core level(InstInter) by default.

apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.0.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v20x/HttpClientRequestInterceptor.java
wu-sheng: dynamicField is instance level, not thread safe.
If you just need context between before and after, please use v2 API, which has MethodInvokeContext for this.

wu-sheng: https://skywalking.apache.org/docs/skywalking-java/next/en/setup/service-agent/java-agent/java-plugin-development-guide/#v2-apis-1

gzlicanyi: > dynamicField is instance level, not thread safe. If you just need context between before and after, please use v2 API, which has MethodInvokeContext for this.

I will make the changes later.

wuwen5: Okay, I misunderstood before，I misunderstood it as an enhanced instance of passing in parameters.

wu-sheng: That was why we introduced v2.
I believe it would make this kind of scenarios easier.

gzlicanyi: > That was why we introduced v2. I believe it would make this kind of scenarios easier.

I have a question. I just found out in the document that there are both v1 and v2 versions of ClassEnhancePluginDefine. Do I have to use ClassEnhancePluginDefineV2 to define InstanceMethodsAroundInterceptorV2?

gzlicanyi: > > That was why we introduced v2. I believe it would make this kind of scenarios easier.
> 
> I have a question. I just found out in the document that there are both v1 and v2 versions of ClassEnhancePluginDefine. Do I have to use ClassEnhancePluginDefineV2 to define InstanceMethodsAroundInterceptorV2?

```java
Caused by: java.lang.ClassCastException: org.apache.skywalking.apm.plugin.spring.cloud.gateway.v20x.HttpClientRequestInterceptor cannot be cast to org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor
```

From the exception message, it seems that you need to modify the definition.

wu-sheng: Yes. They are using different kernel codes. You need to align them.

gzlicanyi: > Yes. They are using different kernel codes. You need to align them.

I have already made the modification, and it's running fine on my project.

wu-sheng (approved):
LGTM. @wuwen5 Please confirm.

wuwen5 (approved):
LGTM

PR #556, opened 2023-06-16
Fix the apm-jdk-threadpool-plugin does not work in certain scenarios

xiaqi1210:
- [ ] Closes #10925.

**why the bug exists:**
apm-jdk-threadpool-plugin does not work when instrumented target had been loaded and used in the FileWriter(kernel) before the instrumentation.

**how to fix it**
The solution is to remove the dependency on ThreadPoolExecutor in the FileWriter and use a simple Thread and a Loop to make this work.


wu-sheng:
Please write the title in a proper way.

wu-sheng (approved):
LGTM

wu-sheng:
Oops @xiaqi1210 This PR doesn't update the changes.md.

wu-sheng:
Could you open another PR to fix this?

xiaqi1210:
> Could you open another PR to fix this?

ok

PR #559, opened 2023-06-20
fix npe in guava eventbus plugin

xzyJavaX:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/10967

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Release Notes.
 * Add RocketMQ 5.x plugin
 * Fix the conflict between the logging kernel and the JDK threadpool plugin.
 * Fix the thread safety bug of finishing operation for the span named "SpringCloudGateway/sendRequest"
+* Fix NPE in guava-eventbus-plugin.
 
 #### Documentation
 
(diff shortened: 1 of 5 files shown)


wu-sheng:
I think we need to explain why the context could be either active or not?

wu-sheng:
@xzyJavaX Any update?

xzyJavaX:
When start a new thread to call `asyncEventBus.post(testEvent)`, the context will be inactive, this situation has no context.

wu-sheng:
So, in this case, the posting message is not traced? Isn't this a bug rather than fix NPE?

lujiajing1126:
> So, in this case, the posting message is not traced? Isn't this a bug rather than fix NPE?

It is traced but orphan. When calling `Context.continue` to correlate two Spans, it throws NPE since the TracingContext does not exist in a new thread.

I think it is users' responsibility to propagate tracing context in this case, e.g. use `@TraceCrossThread`.

This PR can only avoid NPE to be thrown.

wu-sheng:
Make sense. Please fix the UT failing.

wu-sheng (approved):

PR #561, opened 2023-06-23
Improve bytebuddy class enhance for retransform classes

kylixs:
### Improve bytebuddy class enhance for retransform classes

* Add bytebuddy patch to support retransform class multiple times compatible with other javaagent without class cache.
* Support for hotswap classes that are enhanced by skywalking plugins, like Spring MVC Controller, etc.


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/v2/ClassEnhancePluginDefineV2.java
```diff
@@ -56,15 +57,22 @@
  */
 public abstract class ClassEnhancePluginDefineV2 extends AbstractClassEnhancePluginDefine {
 
+    private ConstructorInterceptPoint[] constructorInterceptPoints;
+    private InstanceMethodsInterceptV2Point[] instanceMethodsInterceptV2Points;
+    private StaticMethodsInterceptV2Point[] staticMethodsInterceptV2Points;
```
wu-sheng, line 62: I think these should stay inside method scope? The classes were stateless before

kylixs: Cache the `XxxInterceptPoint`  instances which used by  `DelegateNamingResolver#resolve(Object interceptPoint)`  to generate fixed field name on re-transform class

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/DelegateNamingResolver.java
```diff
@@ -0,0 +1,55 @@
...
+    private final String fieldNamePrefix;
+
+    public static void setNameTrait(String nameTrait) {
+        DelegateNamingResolver.NAME_TRAIT = nameTrait;
+    }
+
+    public DelegateNamingResolver(String className, int identifier) {
+        this.className = className;
+        this.identifier = identifier;
+        this.fieldNamePrefix = NAME_TRAIT + PREFIX + RandomString.hashOf(className.hashCode()) + "$" + RandomString.hashOf(identifier) + "$";
```
wu-sheng, line 43: Could you add comments about the fields created with this prefix?

apm-sniffer/bytebuddy-patch/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+        <artifactId>java-agent-sniffer</artifactId>
+        <version>8.17.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>bytebuddy-patch</artifactId>
+
+    <properties>
+        <maven.compiler.source>8</maven.compiler.source>
+        <maven.compiler.target>8</maven.compiler.target>
```
wu-sheng, line 34: I think these are defined globally in the parent?

apm-sniffer/bytebuddy-patch/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+                    <includes>
+                        <include>**/*Test*.java</include>
+                    </includes>
+                    <excludes>
+                        <exclude>**/Abstract*.java</exclude>
+                    </excludes>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
```
wu-sheng, line 89: What are these for?

apm-sniffer/bytebuddy-patch/src/main/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Resolve auxiliary type
+ */
+public class SWClassFileLocator implements ClassFileLocator {
+    private static ILog LOGGER = LogManager.getLogger(SWClassFileLocator.class);
+
+    private final ForInstrumentation.ClassLoadingDelegate classLoadingDelegate;
+    private Instrumentation instrumentation;
+    private ClassLoader classLoader;
+    private String[] typeNameTraits = {"auxiliary$", "ByteBuddy$", "sw$"};
```
wu-sheng, line 41: When are these default value used? I can see you override this on constructor.

apm-sniffer/bytebuddy-patch/src/main/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
```diff
@@ -0,0 +1,178 @@
...
+import java.lang.instrument.Instrumentation;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Resolve auxiliary type
+ */
+public class SWClassFileLocator implements ClassFileLocator {
```
wu-sheng, line 35: Is this class copied and changed from byte-buddy upstream? If so, we should declare it.
Because there is a lot of class resolution logic, we seem never used. It looks like bytebuddy default behavior.

apm-sniffer/bytebuddy-patch/src/main/java/org/apache/skywalking/apm/agent/bytebuddy/SWExtractionClassFileTransformer.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * An indicator that an attempted class file transformation did not alter the handed class file.
+     */
+    private static final byte[] DO_NOT_TRANSFORM = null;
+
+//    /**
+//     * The class loader that is expected to have loaded the looked-up a class.
+//     */
+//    private final ClassLoader classLoader;
```
wu-sheng, line 34: Are these forgotten to remove?

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/case1/AbstractInterceptTest.java
```diff
@@ -0,0 +1,247 @@
...
+                        System.err.println(String.format("Transform Error: interceptorClassName: %s, typeName: %s, classLoader: %s, module: %s, loaded: %s", interceptorClassName, typeName, classLoader, module, loaded));
+                        throwable.printStackTrace();
+                    }
+                })
+                .installOn(ByteBuddyAgent.install());
+    }
+
+    protected void installConstructorInterceptor(String className, int round) {
+        installConstructorInterceptor1(className, round);
+//        installConstructorInterceptor2(className, round);
```
wu-sheng, line 153: Forget to remove? What is intercetopro1 and interceptor2? Could you give a more proper name?

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
```diff
@@ -17,7 +17,5 @@
 type: tomcat
 entryService: http://localhost:8080/retransform-class-tomcat-scenario/case/retransform-class
 healthCheck: http://localhost:8080/retransform-class-tomcat-scenario/case/healthCheck
-environment:
-  - CATALINA_OPTS="-Dskywalking.agent.is_cache_enhanced_class=true -Dskywalking.agent.class_cache_mode=FILE"
 dependencies:
-withPlugins: apm-spring-annotation-plugin-*.jar
+withPlugins: apm-*.jar
```
wu-sheng, line 21: With all plugins? Are these necessary? I prefer the explicit names of plugins.

wu-sheng:
Could you write pseudocode about the new generated class? Including new field names and types for constructor, instance methods and static methods.

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
kylixs: retransform class test scenario should test with more plugins, especially with the jdk plugins, test scenario is not enough currently, may add more cases later.

wu-sheng: We should separate them, otherwise, it would be confusing when it fails.

apm-sniffer/bytebuddy-patch/src/main/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
kylixs: `SWClassFileLocator`  is used to resovle auxiliary type cross `AgentBuilders`, it is unused currently.

wu-sheng: Could you be more specific about the workflow? Such as why new thread, and when do we need this?

kylixs:
> Could you write pseudocode about the new generated class? Including new field names and types for constructor, instance methods and static methods.

* SWAuxiliaryTypeNamingStrategy
Auxiliary type name pattern: <origin_class_name>$<name_trait>$auxiliary$<auxiliary_type_instance_hash>

* DelegateNamingResolver
Interceptor delegate field name pattern: <name_trait>$delegate$<class_name_hash>$<plugin_define_hash>$<intercept_point_hash>

* SWMethodNameTransformer
Origin method rename pattern: <name_trait>$original$<method_name>$<method_description_hash>

* SWImplementationContextFactory
Method cache value field pattern: cachedValue$<name_trait>$<origin_class_name_hash>$<field_value_hash>
Accessor method name pattern:  <origin_method_rename>$accessor$<name_trait>$<origin_class_name_hash>

```java
 import sample.mybatis.controller.HotelController$sw$auxiliary$19cja42;
 import sample.mybatis.controller.HotelController$sw$auxiliary$p257su0;
 import sample.mybatis.domain.Hotel;
 import sample.mybatis.service.HotelService;

 @RequestMapping(value={"/hotel"})
 @RestController
 public class HotelController
 implements EnhancedInstance {
     @Autowired
     @Lazy
     private HotelService hotelService;
     private volatile Object _$EnhancedClassField_ws;

     // Interceptor delegate fields
     public static volatile /* synthetic */ InstMethodsInter sw$delegate$td03673$ain2do0$8im5jm1;
     public static volatile /* synthetic */ InstMethodsInter sw$delegate$td03673$ain2do0$edkmf61;
     public static volatile /* synthetic */ ConstructorInter sw$delegate$td03673$ain2do0$qs9unv1;
     public static volatile /* synthetic */ InstMethodsInter sw$delegate$td03673$fl4lnk1$m3ia3a2;
     public static volatile /* synthetic */ InstMethodsInter sw$delegate$td03673$fl4lnk1$sufrvp1;
     public static volatile /* synthetic */ ConstructorInter sw$delegate$td03673$fl4lnk1$cteu7s1;

     // Origin method cache value field
     private static final /* synthetic */ Method cachedValue$sw$td03673$g5sobj1;

     public HotelController() {
         this(null);
         sw$delegate$td03673$ain2do0$qs9unv1.intercept(this, new Object[0]);
     }

     private /* synthetic */ HotelController(sw.auxiliary.p257su0 p257su02) {
     }

     @GetMapping(value={"city/{cityId}"})
     public Hotel selectByCityId(@PathVariable(value="cityId") int n) {
         // call interceptor with auxiliary type and parameters and origin method object
         return (Hotel)sw$delegate$td03673$ain2do0$8im5jm1.intercept(this, new Object[]{n}, new HotelController$sw$auxiliary$19cja42(this, n), cachedValue$sw$td03673$g5sobj1);
     }

     // Renamed origin method
     private /* synthetic */ Hotel sw$origin$selectByCityId$a8458p3(int cityId) {
/*22*/         return this.hotelService.selectByCityId(cityId);
     }

     // Accessor of renamed origin method, calling from auxiliary type
     final /* synthetic */ Hotel sw$origin$selectByCityId$a8458p3$accessor$sw$td03673(int n) {
         // Calling renamed origin method
         return this.sw$origin$selectByCityId$a8458p3(n);
     }

     @Override
     public Object getSkyWalkingDynamicField() {
         return this._$EnhancedClassField_ws;
     }

     @Override
     public void setSkyWalkingDynamicField(Object object) {
         this._$EnhancedClassField_ws = object;
     }

     static {
         ClassLoader.getSystemClassLoader().loadClass("org.apache.skywalking.apm.dependencies.net.bytebuddy.dynamic.Nexus").getMethod("initialize", Class.class, Integer.TYPE).invoke(null, HotelController.class, -1072476370);
         // Method object
         cachedValue$sw$td03673$g5sobj1 = HotelController.class.getMethod("selectByCityId", Integer.TYPE);
     }
 }
```

Auxiliary type of Constructor : 
```java
class HotelController$sw$auxiliary$p257su0 {
}
```

Auxiliary type of  `selectByCityId` method: 
```java
class HotelController$sw$auxiliary$19cja42
implements Runnable,
Callable {
    private HotelController argument0;
    private int argument1;

    public Object call() throws Exception {
        return this.argument0.sw$origin$selectByCityId$a8458p3$accessor$sw$td03673(this.argument1);
    }

    @Override
    public void run() {
        this.argument0.sw$origin$selectByCityId$a8458p3$accessor$sw$td03673(this.argument1);
    }

    HotelController$sw$auxiliary$19cja42(HotelController hotelController, int n) {
        this.argument0 = hotelController;
        this.argument1 = n;
    }
}
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/DelegateNamingResolver.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.v2.StaticMethodsInterceptV2Point;
+
+import java.util.Objects;
+
+/**
+ * Generate fixed delegate field name for MethodDelegation
+ */
+public class DelegateNamingResolver {
+    private static final String PREFIX = "delegate$";
+    private static String NAME_TRAIT = "sw$";
```
wu-sheng, line 36: Another duplicate case, I think. This should be either from constructor or final field.

kylixs: The name trait is defined globally in `SkyWalkingAgent` , and passed into `DelegateNamingResolver `  with static method `setNameTrait( )`, how to  use final field?

apm-sniffer/bytebuddy-patch/src/main/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
kylixs: Get class bytecode from separate thread to bypass jdk limitation or bug:  https://github.com/raphw/byte-buddy/issues/1434.  Since we haven't used it yet, I'll move it to the test code for now.

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
kylixs: I think it is ok to load all plugins in this test scenario. In order to ensure the retransform function, plugins that affect retransform need to be identified in time. Retransform class is not a standalone feature and in principle all plugins need to support it.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/DelegateNamingResolver.java
wu-sheng: Yes, I mean that. Then why do we need a default value?

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -91,9 +95,11 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
             return;
         }
 
-        final ByteBuddy byteBuddy = new ByteBuddy().with(TypeValidation.of(Config.Agent.IS_OPEN_DEBUGGING_CLASS));
+        LOGGER.info("Skywalking agent begin to install transformer ...");
+        String nameTrait = getNameTrait();
+        DelegateNamingResolver.setNameTrait(nameTrait);
```
wu-sheng, line 100: These two should move inside `newAgentBuilder`. And `"sw$"` should be moved into `Constants`. And `SWNativeMethodStrategy` doesn't need to forward this, `DelegateNamingResolver#NAME_TRAIT` should be removed too.
Let's reduce those unnecessary codes.

apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
```diff
@@ -147,6 +146,23 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
                .addShutdownHook(new Thread(ServiceManager.INSTANCE::shutdown, "skywalking service shutdown thread"));
     }
 
+    private static AgentBuilder newAgentBuilder(Instrumentation instrumentation, String nameTrait) {
```
wu-sheng, line 149:
```suggestion
        /**
     * Create a new agent builder through customized {@link ByteBuddy} powered by 
     * {@link SWAuxiliaryTypeNamingStrategy} {@link DelegateNamingResolver} {@link SWMethodNameTransformer} and {@link SWImplementationContextFactory}
     */
    private static AgentBuilder newAgentBuilder() {
```

You actually don't need all these parameters and recommend adding comments like this to indicate the changes.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/DelegateNamingResolver.java
wu-sheng: Further, I don't think you need a static `setNameTrait` method as well. Read https://github.com/apache/skywalking-java/pull/561#discussion_r1239792018. 
Keeping class as instance level only would be helpful to understand codes and avoid bugs.

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
```diff
@@ -0,0 +1,174 @@
...
+            return new Resolution.Illegal(name);
+        }
+        if (closed) {
+            throw new IOException("resolve class failure: closed");
+        }
+        // get class binary representation in a clean thread, avoiding nest calling transformer!
+        ResolutionFutureTask futureTask = new ResolutionFutureTask(name);
+        queue.offer(futureTask);
+        try {
+            return futureTask.getFuture().get(timeoutSeconds, TimeUnit.SECONDS);
```
wu-sheng, line 89: offer in 2 sec, but `queue.poll` in 5 sec, what is the logic behind this?

kylixs: Default resolve timeout is 2 sec,  poll timeout has nothing to do with it, just to avoid being blocked for a long time.

wu-sheng: OK, so, you just want the `poll` would be not blocked if there is nothing in the queue, and still be able to exit due to `closed`. Am I right?

apm-sniffer/bytebuddy-patch/src/main/java/org/apache/skywalking/apm/agent/bytebuddy/SWMethodNameTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class SWMethodNameTransformer implements MethodNameTransformer {
+
+    private static final String DEFAULT_PREFIX = "original$";
+
+    private String prefix;
+
+    public SWMethodNameTransformer() {
+        this(DEFAULT_PREFIX);
+    }
```
wu-sheng, line 36: Codes with no usage. Please remove.

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/ConstructorAdvice.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Arrays;
+
+public class ConstructorAdvice {
+
+    @Advice.OnMethodExit(inline = false)
+    public static void onExit(@Advice.Origin Constructor constructor,
+                              @Advice.AllArguments Object[] args) throws Exception {
+        Log.info(String.format("ConstructorAdvice: constructor: %s, args: %s", constructor, Arrays.asList(args)));
+    }
+}
```
wu-sheng, line 33: This advice doesn't only log out. It is only used in `AbstractInterceptTest`. How is this verified? This looks like a local debug rather than a UT.
If this is in a UT, you should verify this log is logged out. But test like `Intercept2Test`, nothing like this happens.
Could you confirm the case?

kylixs: Advice enhancements are a bit difficult to control, so leave them out for now.

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
kylixs: Yes, you are right.

wu-sheng:
Please clean the test codes
1. `EnhanceHelper` has unusage methods
2. `org.apache.skywalking.apm.agent.bytebuddy.case1` to `org.apache.skywalking.apm.agent.bytebuddy.cases`
3. `AbstractRetransformTest` typo `Retransform`
4. `UnmodifiableClassException` of UTs are never thrown
5. `catch` and `e.printStackTrace();` are unexpected
6. About `Log` in test, what is the purpose of that? Logs should not be in the console unless necessary.

kylixs:
> 6\. About `Log` in test, what is the purpose of that? Logs should not be in the console unless necessary.

I want to check output of test directly if any error.

wu-sheng:
> > 6. About `Log` in test, what is the purpose of that? Logs should not be in the console unless necessary.
> 
> I want to check output of test directly if any error.

But it will burn out the CI resources as too many logs. How about use a dummy longer in default, but have to change codes to get logs in the console?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/DelegateNamingResolver.java
```diff
@@ -0,0 +1,79 @@
...
+            StaticMethodsInterceptPoint point = (StaticMethodsInterceptPoint) interceptPoint;
+            return Objects.hash(interceptPointClassName, point.getMethodsMatcher().toString(), point.getMethodsInterceptor(), point.isOverrideArgs());
+        } else if (interceptPoint instanceof StaticMethodsInterceptV2Point) {
+            StaticMethodsInterceptV2Point point = (StaticMethodsInterceptV2Point) interceptPoint;
+            return Objects.hash(interceptPointClassName, point.getMethodsMatcher().toString(), point.getMethodsInterceptorV2(), point.isOverrideArgs());
+        }
+        return interceptPoint.hashCode();
+    }
+
+    public static DelegateNamingResolver get(String className, AbstractClassEnhancePluginDefine pluginDefine) {
```
wu-sheng, line 75: How about using a constructor directly? There is no cache mechanism here, and you always create a new one.

kylixs:
Should we move `hashCode `method into `XXXInterceptPoint` interface from `DelegateNamingResolver#computeHashCode()`?
Something like: 
```java
public interface ConstructorInterceptPoint {

    ElementMatcher<MethodDescription> getConstructorMatcher();

    String getConstructorInterceptor();

    static int hashCode(ConstructorInterceptPoint point) {
        return Objects.hash(point.getClass().getName(), point.getConstructorMatcher().toString(), point.getConstructorInterceptor());
    }
}
```

wu-sheng:
> ```java
> hashCode
> ```

It is better so, and you could do `default` implementation in interfaces rather than static methods
```java
    default int hashcode() {
        return ...;
    }
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/DelegateNamingResolver.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final String PREFIX = "delegate$";
+    private final String fieldNamePrefix;
+
+    public DelegateNamingResolver(String className, AbstractClassEnhancePluginDefine pluginDefine) {
+        // Interceptor delegate field name pattern: <name_trait>$delegate$<class_name_hash>$<plugin_define_hash>$<intercept_point_hash>
+        // something like: InstMethodsInter sw$delegate$td03673$sib0lj0$5n874b1;
+        this.fieldNamePrefix = Constants.NAME_TRAIT + PREFIX + RandomString.hashOf(className.hashCode()) + "$" + RandomString.hashOf(pluginDefine.hashCode()) + "$";
+    }
+
+    public String resolve(Object interceptPoint) {
```
wu-sheng, line 45: To keep Java codes style, please use multiple methods overload rather than a general `Object` type to make the codes more clear.

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
wu-sheng: The point is not about these two plugin tests, it is about the stability of the testing. The plugins are going to be added/removed from time to time, if you copy all, actually, you don't know which plugins exist, which may cause these tests to fail by other contributions.

kylixs: I got it. It's better to test retransform in test scenario of each plugin if we think of retransform as a basic feature. 
Another point, there may be a conflict when some plugins put together, may be the class/method/field name generated by  class enhancement , how to find this out?

wu-sheng: I don't think general plugins matters. After all, the worst case, some plugins would work in retransfer case, right?

We could pick some for tests.
1. General Spring with another agents.
2. Tomcat(another class loader mode) with another agent
3. SkyWalking before or after another agents.
4. SkyWalking before or after bytebuddy based agent(a test agent maybe?)

kylixs: The worst case is not that some plugins don't support retransform, but that the plugins don't work together if there really is a class/method/field name conflict. So before I want to test with all the plugins, on you decide.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/DelegateNamingResolver.java
kylixs: done

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
wu-sheng: I think we don't cover multiple plugins for one method case, how could that happens?

wu-sheng: And even yoi copied all, it doesn't approve this, because many plugins are not activated(not used) in your scenarios.

wu-sheng:
hashcode is JVM included, why you need a new one as parent interface?

kylixs:
> hashcode is JVM included, why you need a new one as parent interface?

Java 8 default methods can not override `equals`, `hashCode` and `toString`.
https://www.java2novice.com/java_interview_questions/java-8-forbidden-to-default-method/
> Can Java 8 default methods override equals, hashCode and toString?
> 
> Answer:
> The methods declared in java.lang.Object class can not be override in Java 8 default methods. It is forbidden to define default methods in interfaces for methods in java.lang.Object.
> 
> Default interface methods can be overwritten in classes implementing the interface and the class implementation of the method has a higher precedence than the interface implementation, even if the method is implemented in a superclass. Since all classes inherit from java.lang.Object, the methods in java.lang.Object would have precedence over the default method in the interface and be invoked instead.

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
kylixs: > I think we don't cover multiple plugins for one method case, how could that happens?

We now support multiple plugins enhancing the same class, although enhancing different methods, but if the same method/field name is generated, there will be an error.

> And even yoi copied all, it doesn't verify this, because many plugins are not activated(not used) in your scenarios.

Yes, and we may need more testcases.

wu-sheng:
Then we need a good name, interceptpoint interface has a method of hashcode, that doesn't make sense.

wu-sheng:
You could override without parent class, and in each using place, you have specific type, you could call hashcode directly.

kylixs:
> You could override without parent class, and in each using place, you have specific type, you could call hashcode directly.

Implementing hashCode by subclasses is prone to omissions, low reliability, and can produce a lot of errors in the future.
1. Change the method name (for example, computeHashCode) to distinguish it from the default JDK method.
2. Extract computeHashCode to the tool class
Which way do you prefer?

wu-sheng:
Please remove intercep point.

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
wu-sheng: > We now support multiple plugins enhancing the same class, although enhancing different methods, but if the same method/field name is generated, there will be an error.

Yes, but that could be error when the plugin tests run too. What your new cases cover is only re-transfer. They are verifying for different cases.

wu-sheng:
> > You could override without parent class, and in each using place, you have specific type, you could call hashcode directly.
> 
> Implementing hashCode by subclasses is prone to omissions, low reliability, and can produce a lot of errors in the future.
> 
> 1. Change the method name (for example, computeHashCode) to distinguish it from the default JDK method.
> 2. Extract computeHashCode to the tool class
>    Which way do you prefer?

I asked for `DelegateNamingResolver` having multiple `resolve` overloads for all interceptor point type, rather than a new `InterceptPoint` interface. The key point is, `InterceptPoint` doesn't include anything really about `intercepting`.

`computeHashCode` is fine to be kept as what it is.

wu-sheng:
I think you misunderstand my point, and did it in the exactly opposite direction of my asking. 
Could you recheck https://github.com/apache/skywalking-java/pull/561#issuecomment-1605276928?

kylixs:
> The key point is, `InterceptPoint` doesn't include anything really about `intercepting`.

Sometimes an interface is used to identify a certain type of objects, such as `java.io.Serializable` , `java.lang.Cloneable`, `java.util.EventListener`

kylixs:
> I asked for `DelegateNamingResolver` having multiple `resolve` overloads for all interceptor point type, rather than a new `InterceptPoint` interface.

I don't quite understand this. Can you give me an example? How do you override a method without an interface and a parent class?

wu-sheng:
1. Remove `InterceptPoint`
2. DelegateNamingResolver has methods for `resolve#ConstructorInterceptV2Point`, `resolve#ConstructorInterceptPoint`, `resolve#InstanceMethodsInterceptV2Point`, etc.(6 methods)
3. `computeHashCode` stays as default method in `ConstructorInterceptV2Point`, `ConstructorInterceptPoint`, etc.

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
```diff
@@ -0,0 +1,174 @@
...
+        thread.interrupt();
+    }
+
+    public int getTimeoutSeconds() {
+        return timeoutSeconds;
+    }
+
+    public void setTimeoutSeconds(int timeoutSeconds) {
+        this.timeoutSeconds = timeoutSeconds;
+    }
```
wu-sheng, line 155: Unused methods.

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
```diff
@@ -0,0 +1,174 @@
...
+            Class aClass = allLoadedClasses[i];
+            if (className.equals(aClass.getName())) {
+                return aClass;
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public void close() throws IOException {
```
wu-sheng, line 143:
```suggestion
    public void close() {
```
No I/O Exception.

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
```diff
@@ -0,0 +1,174 @@
...
+
+    private Resolution getResolution(String name) throws Exception {
+        SWExtractionClassFileTransformer classFileTransformer = new SWExtractionClassFileTransformer(name);
+        try {
+            instrumentation.addTransformer(classFileTransformer, true);
+            Class aClass = locateClass(name);
+            if (aClass == null) {
+                return new Resolution.Illegal(name);
+            }
+            instrumentation.retransformClasses(new Class[]{aClass});
```
wu-sheng, line 114: What does this `retransformClasses` mean?

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
kylixs: It's used to config resolve timeout, not every public method need a usage, might need it somewhere later.

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
kylixs: Trigger re-transforming the target class, and receive bytecode in SWExtractionClassFileTransformer.
This is a way to get bytecode from a class, especially a dynamically generated or enhanced class. Only original bytecode can be read from the classpath, and without dynamically generated classes.

apm-sniffer/bytebuddy-patch/src/test/java/org/apache/skywalking/apm/agent/bytebuddy/SWClassFileLocator.java
wu-sheng: This is only a test code, I think usually manual hard code is good enough?

wu-sheng:
Generally, this PR should be good enough. Update change logs about the new naming policy of code generation

After this PR is merged, I think you should set another PR to update the version to `9.0.0-SNAPSHOT`, we are going to bump up the version number as new kernel relative codes.

Update main compatibility doc, https://skywalking.apache.org/docs/main/next/en/setup/service-agent/agent-compatibility/ `> = 8.0 & > = 9.0`

kylixs:
To ensure that the hashCode for recreating the InterceptPoint instance is the same as the previous instance, each ElementMatcher implementation class needs to implement toString(). Where should I put this hint? There seems to be no way to guarantee it.

wu-sheng:
Do you want a consistent and certain hashcode for every method?

kylixs:
> Do you want a consistent and certain hashcode for every method?

Yes,  should return a consistent and certain hashcode from `DelegateNamingResolver#resolve(InterceptPoint)`  for same plugin define on re-transforming.

wu-sheng:
Isn't your current algorithm guaranteed? hashcode is not changed once the values of involved fields are not changed. 
Could you explain your concern? 
Are you concern about code changed after retransfer or conflicts among methods in a same class?

kylixs:
The current code is fine, and both ByteBuddy and our custom ElementMatcher implement the toString() method. I worry that in the future, new ElementMatcher subclasses might not implement the toString() method, with no mandatory constraint guarantees. If the hashcode is inconsistent, the retransform will fail.

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
kylixs: It's not that simple. Different plugin enhancers of the same class may affect each other during re-transforming. 
Leave this aside for now and add test cases later if incompatibilities arise.

wu-sheng:
That should be able to check through UT.
Are you ware of OAP class scan? You could check whether the scaned matchers implement/override(declared) the expected methods.

kylixs:
Any doc/example about  OAP class scan?

wu-sheng:
https://github.com/apache/skywalking/blob/master/oap-server/server-core/src/main/java/org/apache/skywalking/oap/server/core/annotation/AnnotationScan.java

Ref this, this is an annotation scan. You could easily do an inheritance scan(matcher implementations, and/or other class type implementations).

wu-sheng (approved):
LGTM. Thanks for making this ready fast. Please update the version in another PR.

wu-sheng:
@kylixs As you have been involved into core mechanism of the agent, and did significant contributions, I am willing to invite you join the committer team. If you want to, please ping me on the slack

kylixs:
That's good. I'm interested in that.

test/plugin/scenarios/retransform-class-tomcat-scenario/configuration.yml
wspray: > > I think we don't cover multiple plugins for one method case, how could that happens?
> 
> We now support multiple plugins enhancing the same class, although enhancing different methods, but if the same method/field name is generated, there will be an error.
> 
> > And even yoi copied all, it doesn't verify this, because many plugins are not activated(not used) in your scenarios.
> 
> Yes, and we may need more testcases.

@wu-sheng @kylixs  Hello，I wonder if multiple plugins enhancing the same class has been supported?

wu-sheng: No, it isn't. Also, because we added a new field to the target class, I don't recommend to supporting that.

PR #570, opened 2023-07-05
fix CVE-2023-2976 and CVE-2023-34462

alanlvle:

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/guava-eventbus-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/guava-eventbus-plugin/pom.xml
@@ -27,7 +27,7 @@
     <artifactId>apm-guava-eventbus-plugin</artifactId>
 
     <properties>
-        <guava.version>32.0.1-jre</guava.version>
+        <guava.version>31.0.1-jre</guava.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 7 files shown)


wu-sheng:
You need to update change logs and licenses to adopt these dependencies lib changes.
And tests codes don't matter.

alanlvle:
update

wu-sheng:
I can't see the guava version update in the license file.

apm-sniffer/apm-sdk-plugin/guava-eventbus-plugin/pom.xml
```diff
@@ -27,7 +27,7 @@
     <artifactId>apm-guava-eventbus-plugin</artifactId>
 
     <properties>
-        <guava.version>31.0.1-jre</guava.version>
+        <guava.version>32.0.1-jre</guava.version>
```
lujiajing1126, line 30: This is not necessary since it should be a provided dependency

lujiajing1126 (changes requested):

wu-sheng:
All versions in test are not necessary, actually.

alanlvle:
> All versions in test are not necessary, actually.

I need revert it?

lujiajing1126:
> > All versions in test are not necessary, actually.
> 
> I need revert it?

Sure. Pls only make necessary changes.

wu-sheng:
It's fine

wu-sheng (dismissed):
I am good in either way. Rely on @wallezhang

lujiajing1126 (approved):
lgtm

PR #590, opened 2023-08-03
Support asynchronous invocation in jetty client 9.0 and 9.x plugin

yangyulely:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/11169


wu-sheng:
@xu1009 Could you review this as you added Jetty plugins before.

wu-sheng:
@yangyulely You need to fix CI and all tests first.

yangyulely:
> @yangyulely You need to fix CI and all tests first.

The reason is the test case code style checks failed, i hava modify and push. In fact, I have already pass the case in my local env by run `test/plugin/run.sh jetty-scenario`. By the way, why it is not same result between workflow ? How could i check test cases local ?

test/plugin/scenarios/jetty-scenario/jettyclient-scenario/src/main/java/org/apache/skywalking/apm/testcase/jettyclient/controller/CaseController.java
```diff
@@ -45,6 +46,8 @@ public void init() throws Exception {
     @ResponseBody
     public String jettyClientScenario() throws Exception {
         client.newRequest("http://" + jettyServerHost + ":18080/jettyserver-case/case/receiveContext-0").send();
+        Response.CompleteListener listener = result -> { };
```
xu1009, line 49: if you do something in listener such as query mysql, the trace context can not been got because your plugin just put span in request attribute

xu1009:
> 

you can test in local with checkstyle maven plugin

test/plugin/scenarios/jetty-scenario/jettyclient-scenario/src/main/java/org/apache/skywalking/apm/testcase/jettyclient/controller/CaseController.java
wu-sheng: @yangyulely I think the suggestion is, you should have improvements for the listener callback, such as a wrapper and a local span created for further processes in the listener.

yangyulely: I've try to a wrapper for listener, and reset the argument to the new wrapper, but it looks like the argument was not that setted. could you share some experience or suggestion for it?

wu-sheng: You need to declare argument override in the instrument, please read the doc.

xu1009, line 49: @yangyulely can you add something(such as http request) in listener callback to test it

yangyulely: ok, i will update later

test/plugin/scenarios/jetty-scenario/config/expectedData.yaml
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 segmentItems:
 - serviceName: jettyserver-scenario
-  segmentSize: 2
+  segmentSize: 3
```
wu-sheng, line 18: Why is this a specific value? I think it should be `gt 3` or something.
There are re-try mechanisms there, this determined value could cause unnecessary tests failing.

yangyulely: has update, and I‘ve test ok local

xu1009 (approved):
LGTM

PR #593, opened 2023-08-15
Add nacos-client 2.x plugin

peachisai:
- [x] Add a component id in [the component-libraries.yml]


apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -241,4 +241,5 @@ public class ComponentsDefine {
 
     public static final OfficialComponent AEROSPIKE = new OfficialComponent(149, "Aerospike");
 
+    public static final OfficialComponent NACOS = new OfficialComponent(150, "Nacos");
```
wu-sheng, line 244: Please register the ID in the main repo. 
If you want to show the logo on the topo, you need to submit the logo to the UI repo as well.

test/plugin/scenarios/nacos-client-2.x-scenario/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+    <pluginRepositories>
+        <pluginRepository>
+            <id>spring-snapshots</id>
+            <url>https://repo.spring.io/snapshot</url>
+        </pluginRepository>
+        <pluginRepository>
+            <id>spring-milestones</id>
+            <url>https://repo.spring.io/milestone</url>
+        </pluginRepository>
+    </pluginRepositories>
```
wu-sheng, line 106: Why do you need this?

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
peachisai: ok,i will summit the ID in the main repo

test/plugin/scenarios/nacos-client-2.x-scenario/pom.xml
peachisai: i will remove them

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -69,6 +69,7 @@ metrics based on the tracing data.
   * [Thrift](https://github.com/apache/thrift/tree/master/lib/java) 0.10.0 -> 0.12.0
   * [Apache CXF](https://github.com/apache/cxf) 3.x
   * [JSONRPC4J](https://github.com/briandilley/jsonrpc4j) 1.2.0 -> 1.6
+  * [Nacos-Client](https://github.com/alibaba/nacos) 2.x
```
wu-sheng, line 72: Refer to the Zookeeper plugin(Optional-plugins.md), it should be optional. 

> Plugin of Zookeeper 3.4.x in optional plugin folder. The reason of being optional plugin is, many business irrelevant traces are generated, which cause extra payload to agents and backends. At the same time, those traces may be just heartbeat(s).

docs/en/setup/service-agent/java-agent/Optional-plugins.md
```diff
@@ -23,3 +23,4 @@ Now, we have the following known optional plugins.
 * Plugin of Apache ShenYu(incubating) Gateway 2.4.x in optional plugin folder. Please only activate this plugin when you install agent in Apache ShenYu Gateway.
 * Plugin of trace sampler CPU policy in the optional plugin folder. Please only activate this plugin when you need to disable trace collecting when the agent process CPU usage is too high(over threshold).
 * Plugin for Spring 6.x and RestTemplate 6.x are in the optional plugin folder. Spring 6 requires Java 17 but SkyWalking is still compatible with Java 8. So, we put it in the optional plugin folder.
+* Plugin of nacos-client 2.x lib in optional plugin folder.
```
wu-sheng, line 26: Please add reasons of being optional.

docs/en/setup/service-agent/java-agent/Supported-list.md
wu-sheng: Please add optional suffix.

docs/en/setup/service-agent/java-agent/Optional-plugins.md
peachisai: OK, I'm done

wu-sheng:
I think you should submit Nacos logo to the webui, as your generated spans include both exit and entry, which would another conjecture node on the topo graph to represent the NacosServer.

peachisai:
> I think you should submit Nacos logo to the webui, as your generated spans include both exit and entry, which would another conjecture node on the topo graph to represent the NacosServer.



> I think you should submit Nacos logo to the webui, as your generated spans include both exit and entry, which would another conjecture node on the topo graph to represent the NacosServer.

OK,i will try it

wu-sheng (approved):
LGTM

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
wu-sheng: According to https://github.com/apache/skywalking-booster-ui/pull/311#issuecomment-1680548512, this should be renamed as `nacos-client`.

wu-sheng (changes requested):

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
peachisai: > According to [apache/skywalking-booster-ui#311 (comment)](https://github.com/apache/skywalking-booster-ui/pull/311#issuecomment-1680548512), this should be renamed as `nacos-client`.

shall i add nacos server component ?

wu-sheng:
I am going to wait for your new test result.

wu-sheng:
@peachisai Do you find out where is the issue?

peachisai:
> @peachisai Do you find out where is the issue?

 i found something, i will post it on the night

wu-sheng (approved):
According to https://github.com/apache/skywalking-booster-ui/pull/311#issuecomment-1682510564, the server-side logo would not show up as the RPC type span.
So, adding a component mapping and/or logo is not required.

LGTM

PR #594, opened 2023-08-15
Add Spring Runner plugin

lsq27:
### Add an agent plugin to support [Spring runner](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.spring-application.command-line-runner)

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/spring-plugins/runner-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/runner/RunnerMethodInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/spring-plugins/runner-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/runner/RunnerMethodInterceptor.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.spring.runner;
+
+import java.lang.reflect.Method;
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.context.tag.Tags;
+import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
+
+/**
+ * Record the Spring runner task local span.
+ */
+public class RunnerMethodInterceptor implements InstanceMethodsAroundInterceptor {
+
+    @Override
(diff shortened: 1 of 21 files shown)


kezhenxu94:
I don't really see what's the point to trace the runner, Spring runners are basically one-off tasks. Can you share why we want to trace them?

apm-sniffer/apm-sdk-plugin/spring-plugins/runner-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/runner/RunnerMethodInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+ * Record the Spring runner task local span.
+ */
+public class RunnerMethodInterceptor implements InstanceMethodsAroundInterceptor {
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
+                             MethodInterceptResult result) throws Throwable {
+        String operationName = ComponentsDefine.SPRING_RUNNER.getName() + "/" + allArguments[0].getClass().getName();
+        AbstractSpan span = ContextManager.createLocalSpan(operationName);
+        Tags.LOGIC_ENDPOINT.set(span, Tags.VAL_LOCAL_SPAN_AS_LOGIC_ENDPOINT);
```
wu-sheng, line 40: Create a logic span in plugin? Why do we need to measure a local run? 

I have doubts about whether this is a general use case. If you want to measure this, the toolkit annotation and APIs could help you, rather than a public plugin.

lsq27:
> I don't really see what's the point to trace the runner, Spring runners are usually one-off tasks. Can you share why we want to trace them?

The feature was inspired by the [spring-scheduled-annotation-plugin](https://github.com/apache/skywalking-java/tree/main/apm-sniffer/apm-sdk-plugin/spring-plugins/scheduled-annotation-plugin). Spring runner can be considered a special scheduled task triggered only once after the SpringApplication has started.

Asuming that an application needs access several services via HTTP or RPC in Spring runners, spans can be integrated in one segment with this plugin, otherwise each span will be in different segments. This plugin can show which class send the requests like `spring-scheduled-annotation-plugin` does.

The feature can be used to improve the tracing of Spring Boot. Since this feature does only a samll thing at the beginning of application, whether the feature should be made a stand alone plugin or a part of so-called spring-boot-plugin in which the code about Spring Boot can be put?

before

image

after

image

wu-sheng:
How to trigger a runner? Is it always an one time task or repeatable triggered?

apm-sniffer/apm-sdk-plugin/spring-plugins/runner-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/runner/RunnerMethodInterceptor.java
lsq27: I have a SpringApplication with complex interactions with other services in CommandLineRunner. Skywalking show many segments at this situation and I try to integrate them into one segment according to code in `spring-scheduled-annotation-plugin`.

It's my fault to add Logic Endpoint here, the runner only run once.

lsq27:
> How to trigger a runner? Is it always an one time task or repeatable triggered?

According to [Spring](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.spring-application.command-line-runner)

> If you need to run some specific code once the SpringApplication has started, you can implement the ApplicationRunner or CommandLineRunner interfaces. Both interfaces work in the same way and offer a single run method, which is called just before SpringApplication.run(…​) completes.

Runners only run once in order after the application starts accepting traffic.

wu-sheng:
If this is a one time running, especially on booting stage, I am highly doubt whether agent always monitors this.
The agent would not work until it is connected with backend, but the runner could have finished by then.

SkyWalking is monitoring server style application running status, rather than booting details.

wu-sheng:
@lsq27 Any feedback about the use case concern.

lsq27:
> If this is a one time running, especially on booting stage, I am highly doubt whether agent always monitors this. The agent would not work until it is connected with backend, but the runner could have finished by then.
> 
> SkyWalking is monitoring server style application running status, rather than booting details.

Spring runners only run when Spring has fully started, just like a user is accessing a normal endpoint and then some code in controller run.

So it is possible that a application is ready to accept user traffic but the agent can not trace the user access until it is connected with backend? If so, this plugin may not work sometimes.

wu-sheng:
> So it is possible that a application is ready to accept user traffic but the agent can not trace the user access until it is connected with backend? If so, this plugin can not work sometimes.

There is no guarantee. The agent only works when connected. So, who would come first is not a certain thing. 
We are not `can't`, we `choose don't` to protect the user app memory.

PR #596, opened 2023-08-21
Staticize the tags for preventing synchronization in JDK 8

peachisai:


wu-sheng:
It seems rocketMQ plugins have other issues. Please fix.

PR #598, opened 2023-08-25
add rocketmq-client-java-5.x plugin

xyyz150:
- [ ] Add a test case for the new plugin, refer to [the doc](https://github.com/apache/skywalking-java/blob/main/docs/en/setup/service-agent/java-agent/Plugin-test.md)

### DIFF
--- a/.github/workflows/plugins-test.2.yaml
+++ b/.github/workflows/plugins-test.2.yaml
@@ -81,6 +81,7 @@ jobs:
           - jersey-2.26.x-2.39.x-scenario
           - websphere-liberty-23.x-scenario
           - nacos-client-2.x-scenario
+          - rocketmq-client-java-scenario
     steps:
       - uses: actions/checkout@v2
         with:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -152,7 +152,8 @@ Callable {
 * Merge two instrumentation classes to avoid duplicate enhancements in MySQL plugins.
 * Support asynchronous invocation in jetty client 9.0 and 9.x plugin
 * Add nacos-client 2.x plugin
-* Staticize the tags for preventing synchronization in JDK 8 
+* Staticize the tags for preventing synchronization in JDK 8
+* Add RocketMQ-Client-Java 5.x plugin
 
 #### Documentation
 
--- a/apm-sniffer/apm-sdk-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/pom.xml
@@ -134,6 +134,7 @@
         <module>rocketMQ-5.x-plugin</module>
         <module>websphere-liberty-23.x-plugin</module>
         <module>aerospike-plugin</module>
+        <module>rocketMQ-client-java-5.x-plugin</module>
     </modules>
     <packaging>pom</packaging>
 
--- a/apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/pom.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
+  ~ contributor license agreements.  See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright ownership.
+  ~ The ASF licenses this file to You under the Apache License, Version 2.0
+  ~ (the "License"); you may not use this file except in compliance with
+  ~ the License.  You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  ~
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>apm-sdk-plugin</artifactId>
+        <groupId>org.apache.skywalking</groupId>
+        <version>9.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>apm-rocketmq-client-java-5.x-plugin</artifactId>
+    <name>rocketMQ-client-java-5.x-plugin</name>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <rocketmq-client-java.version>5.0.5</rocketmq-client-java.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.rocketmq</groupId>
+            <artifactId>rocketmq-client-java</artifactId>
+            <version>${rocketmq-client-java.version}</version>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-deploy-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageListenerInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageListenerInterceptor.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.rocketMQ.client.java.v5;
+
+import org.apache.rocketmq.client.apis.consumer.ConsumeResult;
+import org.apache.rocketmq.client.apis.message.MessageView;
+import org.apache.skywalking.apm.agent.core.context.CarrierItem;
+import org.apache.skywalking.apm.agent.core.context.ContextCarrier;
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.context.tag.Tags;
+import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
+import org.apache.skywalking.apm.agent.core.context.trace.SpanLayer;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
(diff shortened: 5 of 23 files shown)


wu-sheng:
`rocketmq-scenario` is there, and have tested for 5.1.0. What is new of this one?

xyyz150:
This is another official client of Rocketmq, which is the gRPC protocol. Better to use than Rocketmq's client

wu-sheng:
Please be clear about, it seems also from ASF RocketMQ. What do you mean better than RocketMQ client.

xyyz150:
> 



> Please be clear about, it seems also from ASF RocketMQ. What do you mean better than RocketMQ client.

The gRPC protocol was newly introduced in version 5.0, aimed at evolving a more lightweight, standardized, and easily extensible client-server communication protocol with mainstream cloud native technologies. The gRPC protocol SDK evolves as an independent repository [RocketMQ Clients ](https://github.com/apache/rocketmq-clients), supporting languages such as Java/C++/.NET/Go/Rust. There is a relatively decoupled relationship between the client and the server, following the RocketMQ API protocol interface agreement. [the doc](https://rocketmq.apache.org/docs/sdk/01overview/)

I think `rocketmq-client-java` is easier to use.

wu-sheng:
Thanks for the clearance.

wu-sheng:
Then please rename your scenario to rocketmq-5-grpc-scenario

xyyz150:
> Then please rename your scenario to rocketmq-5-grpc-scenario

modify completed，check please

wu-sheng:
You need to fix tests and make sure all could pass locally, before you asked for our review.

xyyz150:
> You need to fix tests and make sure all could pass locally, before you asked for our review.

I'm sorry，but i test locally `bash ./test/plugin/run.sh -f rocketmq-5-grpc-scenario` ，no code style exceptions occurred

wu-sheng:
Remove the commented codes, that is unnecessary.
Code style will be checked when it runs on CI.

xyyz150:
> Remove the commented codes, that is unnecessary. Code style will be checked when it runs on CI.

i'm sorry, i check style pass locally

> [INFO] Scanning for projects...
[INFO] 
[INFO] ----< org.apache.skywalking.apm.testcase:rocketmq-5-grpc-scenario >-----
[INFO] Building skywalking-rocketmq-5-grpc-scenario 1.0.0
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-checkstyle-plugin:3.1.1:check (validate) @ rocketmq-5-grpc-scenario ---
[INFO] Starting audit...
Audit done.
[INFO] You have 0 Checkstyle violations.

xyyz150:
testing encountered an issue. `rocketmq-client` not support autoCreateTopicEnable , [look here](https://github.com/apache/rocketmq-clients/issues/430)


so Plugin Test occurred exception: `CODE: 17  DESC: No topic route info in name server for the topic: TopicTest`. can you give  some advice @wu-sheng 

but i'm not create topic by `rocketmq-client` .  when i test pass locally , access the rocketmq docker I started myself

image

wu-sheng:
What kind of ways do they support to create a topic?  A CLI or using legacy client lib?

This [test/plugin/scenarios/rocketmq-5-grpc-scenario/bin/startup.sh](https://github.com/apache/skywalking-java/pull/598/files#diff-19967ea15aa4b699d826353dbe654369b0ac7fdb6ee23108f32824c1fb855532) script would be called.

xyyz150:
> What kind of ways do they support to create a topic? A CLI or using legacy client lib?
> 
> This [test/plugin/scenarios/rocketmq-5-grpc-scenario/bin/startup.sh](https://github.com/apache/skywalking-java/pull/598/files#diff-19967ea15aa4b699d826353dbe654369b0ac7fdb6ee23108f32824c1fb855532) script would be called.

can use [admin tool](https://rocketmq.apache.org/docs/deploymentOperations/02admintool), but I don't know how to use it appropriately in a plugin test case.

wu-sheng:
This seems a shell, you could download it and run the command in the startup script.
Anyway, this is mostly RocketMQ thing. I have limited knowledge for that.

xyyz150:
> This seems a shell, you could download it and run the command in the startup script. Anyway, this is mostly RocketMQ thing. I have limited knowledge for that.

ok, i will try it

wu-sheng:
Why are you adding so many binaries and scripts for only setting a topic? The binaries are not allowed.

xyyz150:
i test pass locally

image

wu-sheng:
No binary is accepted by us, as they are not source codes. I believe this would not be so hard to just create a topic for RocketMQ server. If you don't know how to do this, I would suggest you talk with their community, to find out `What is the quickest way to create a topic through CLI?`.

wu-sheng:
> > 
> 
> > Why are you adding so many binaries and scripts for only setting a topic? The binaries are not allowed.
> 
> tool dependent on this binaries ,I'm not sure which ones can not be used, so I uploaded all of them

None of them is acceptable. People would not like to have huge binaries in the source codes tar. Also, this is not allowed by the ASF.

This tool should be downloaded through shell, rather than like this. 

Like I said, this should not be this hard. Please reach RocketMQ community about the question I raised.

xyyz150:
> > > 
> > 
> > 
> > > Why are you adding so many binaries and scripts for only setting a topic? The binaries are not allowed.
> > 
> > 
> > tool dependent on this binaries ,I'm not sure which ones can not be used, so I uploaded all of them
> 
> None of them is acceptable. People would not like to have huge binaries in the source codes tar. Also, this is not allowed by the ASF.
> 
> This tool should be downloaded through shell, rather than like this.
> 
> Like I said, this should not be this hard. Please reach RocketMQ community about the question I raised.



> > > 
> > 
> > 
> > > Why are you adding so many binaries and scripts for only setting a topic? The binaries are not allowed.
> > 
> > 
> > tool dependent on this binaries ,I'm not sure which ones can not be used, so I uploaded all of them
> 
> None of them is acceptable. People would not like to have huge binaries in the source codes tar. Also, this is not allowed by the ASF.
> 
> This tool should be downloaded through shell, rather than like this.
> 
> Like I said, this should not be this hard. Please reach RocketMQ community about the question I raised.

like download from `RocketMQ Releases URL` ，after execute shell ？

wu-sheng:
Maybe. My point is
1. Binaries are not allowed
2. If you need any binary, download it when the tests run
3. Ask the upstream for their suggestions. How do they set up test env? Do they always need so huge admin rather than a simple one-file CLI? I don't know, I am just guessing.

wu-sheng:
Another question is, why can't you use your demo codes to create the topic? I think `/health` request is the entrance point for your env preparation.

xyyz150:
> Another question is, why can't you use your demo codes to create the topic? I think `/health` request is the entrance point for your env preparation.

`rocketmq-client-java` not provided create topic and group Apis, can i use other client to create topic and group

wu-sheng:
> > Another question is, why can't you use your demo codes to create the topic? I think `/health` request is the entrance point for your env preparation.
> 
> `rocketmq-client-java` not provided create topic and group Apis, can i use other client to create topic and group

That is my point. You should have a simpler way to do so.

test/plugin/scenarios/rocketmq-5-grpc-scenario/src/main/java/test/apache/skywalking/apm/testcase/rocketmq/client/java/controller/CaseController.java
```diff
@@ -0,0 +1,151 @@
...
+                    .setTag(TAG)
+                    // Key(s) of the message, another way to mark message besides message id.
+                    .setKeys("KeyA")
+                    .setBody("This is a normal message for Apache RocketMQ".getBytes(StandardCharsets.UTF_8))
+                    .build();
+            SendReceipt sendReceipt = producer.send(message);
+            System.out.printf("%s send msg successfully, message: %s%n", new Date(), sendReceipt);
+
+            // start consumer
+            Thread thread = new Thread(new Runnable() {
```
wu-sheng, line 90: Is the provider and comsumer created duplicate for every request? This endpoint is called repeatedly when retry.

test/plugin/scenarios/rocketmq-5-grpc-scenario/src/main/java/test/apache/skywalking/apm/testcase/rocketmq/client/java/controller/CaseController.java
```diff
@@ -0,0 +1,151 @@
...
+        subArgs = new String[]{
+                "updateSubGroup",
+                "-n",
+                nameServer,
+                "-c",
+                "DefaultCluster",
+                "-g",
+                "group1"};
+        MQAdminStartup.main(subArgs);
+        System.out.printf("HealthCheck Provider Started.%n");
```
wu-sheng, line 137: No system out in codes is allowed.
Health check should include provider and consumer creation, right?
And the topic, provider and consumer should be created once.

test/plugin/scenarios/rocketmq-5-grpc-scenario/src/main/java/test/apache/skywalking/apm/testcase/rocketmq/client/java/controller/CaseController.java
xyyz150: I am a reference `rocketmq-scenario` demo

image

test/plugin/scenarios/rocketmq-5-grpc-scenario/src/main/java/test/apache/skywalking/apm/testcase/rocketmq/client/java/controller/CaseController.java
xyyz150: i find system out in many code. If the rule is like this, I will remove it

i will add create provider  in Health check

test/plugin/scenarios/rocketmq-5-grpc-scenario/src/main/java/test/apache/skywalking/apm/testcase/rocketmq/client/java/controller/CaseController.java
wu-sheng: Ref can't prove it is correct or wrong. Sometimes, wrong things got missed from last review.

test/plugin/scenarios/rocketmq-5-grpc-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,85 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+segmentItems:
+  - serviceName: rocketmq-5-grpc-scenario
+    segmentSize: ge 2
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: HEAD:/case/healthCheck
```
wu-sheng, line 22: `/healthCheck` should not be included in the expected data.

test/plugin/scenarios/rocketmq-5-grpc-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,85 @@
...
+            spanType: Entry
+            peer: ''
+            tags:
+              - {key: url, value: 'http://localhost:8080/rocketmq-5-grpc-scenario/case/rocketmq-5-grpc-scenario'}
+              - {key: http.method, value: GET}
+              - {key: http.status_code, value: '200'}
+            skipAnalysis: 'false'
+      - segmentId: not null
+        spans:
+          - operationName: RocketMQ/TopicTest/Consumer
```
wu-sheng, line 72: For the consumer, where is the `ref`? This entry span should have a ref pointing to the provider's span. Otherwise, you actually don't verify the tracing is continuing across processes.

wu-sheng:
`Supported-list.md` should be updated as well.

test/plugin/scenarios/rocketmq-5-grpc-scenario/config/expectedData.yaml
xyyz150: i fix it

wu-sheng:
> `Supported-list.md` should be updated as well.

This is still missing. @xyyz150 And why do you add 5.1.0 and 5.1.1, are they different?

xyyz150:
> > `Supported-list.md` should be updated as well.
> 
> This is still missing. @xyyz150 And why do you add 5.1.0 and 5.1.1, are they different?

i not sure the difference about 5.1.0 and 5.1.1, i just add test pass version

should i add `Supported-list.md` like this : 

> * [RocketMQ](http://github.com/apache/rocketmq-clients) 5.x

wu-sheng:
> > > `Supported-list.md` should be updated as well.
> 
> > 
> 
> > This is still missing. @xyyz150 And why do you add 5.1.0 and 5.1.1, are they different?
> 
> 
> 
> i not sure the difference about 5.1.0 and 5.1.1, i just add test pass version
> 

Once tests are passed. Please remove 5.1.0. We don't want to cost too many CI resources.

> 
> 
> should i add `Supported-list.md` like this : 
> 
> 
> 
> > * [RocketMQ](http://github.com/apache/rocketmq-clients) 5.x

You should indicate 5.x gRPC client, and update previous RocketMQ to include 5.x as well. I think we have supported the legacy client?

wu-sheng (approved):
LGTM

PR #601, opened 2023-08-31
Support maven revision

zhaoyunxing92:
### <Feature description>
- [x] [Using of CI friendly versions ](https://maven.apache.org/docs/3.5.0/release-notes.html#overview-about-the-changes) eg: `mvnd clean package -Dmaven.test.skip=true -Drevision=8.16.12 -Pall`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
 
     <groupId>org.apache.skywalking</groupId>
     <artifactId>java-agent</artifactId>
-    <version>9.1.0-SNAPSHOT</version>
+    <version>${revision}</version>
 
     <parent>
         <groupId>org.apache</groupId>
@@ -79,6 +79,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <compiler.version>8</compiler.version>
+        <revision>9.1.0-SNAPSHOT</revision>
         <checkstyle.version>6.18</checkstyle.version>
         <junit.version>4.12</junit.version>
         <mockito-core.version>5.0.0</mockito-core.version>
@@ -122,7 +123,8 @@
         <jmh.version>1.33</jmh.version>
         <gmaven-plugin.version>1.5</gmaven-plugin.version>
         <checkstyle.fails.on.error>true</checkstyle.fails.on.error>
-
+        <flatten-maven-plugin.version>1.5.0</flatten-maven-plugin.version>
+        <maven-surefire-plugin.version>3.0.0-M8</maven-surefire-plugin.version>
     </properties>
 
     <profiles>
@@ -273,22 +275,27 @@
                     <version>${takari-maven-plugin.version}</version>
                 </plugin>
                 <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-antrun-plugin</artifactId>
                     <version>${maven-antrun-plugin.version}</version>
                 </plugin>
                 <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-deploy-plugin</artifactId>
                     <version>${maven-deploy-plugin.version}</version>
                 </plugin>
                 <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-assembly-plugin</artifactId>
                     <version>${maven-assembly-plugin.version}</version>
                 </plugin>
                 <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-failsafe-plugin</artifactId>
                     <version>${maven-failsafe-plugin.version}</version>
                 </plugin>
                 <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-jar-plugin</artifactId>
                     <version>${maven-jar-plugin.version}</version>
                     <configuration>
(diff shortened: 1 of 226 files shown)


wu-sheng:
This is never discussed. Why do you add this?

zhaoyunxing92:
> This is never discussed. Why do you add this?

Unify the version management in the parent class, which is convenient for publishing and packaging without modifying a version file every time you publish

lujiajing1126:
> > This is never discussed. Why do you add this?
> 
> Unify the version management in the parent class, which is convenient for publishing and packaging without modifying a version file every time you publish

Is this compatible with the maven-publish plugin as well as the Apache release procedure?

zhaoyunxing92:
> > > This is never discussed. Why do you add this?
> > 
> > 
> > Unify the version management in the parent class, which is convenient for publishing and packaging without modifying a version file every time you publish
>
> Is this compatible with the maven-publish plugin as well as the Apache release procedure?

Compatible. In the past, `mvn versions:set -DnewVersion=9.1.0` may be used to modify the version number in batches. Now it is only used to modify one place in the pom file. If you use the command, you can use `-Drevision=9.1.0` to modify, eg: `mvnd clean package -Dmaven.test.skip=true -Drevision=8.16.12 -Pall`

wu-sheng:
I can't see the point of using this. If you need to package for specific version, why don't you use the give git tag?

kezhenxu94:
> > > > This is never discussed. Why do you add this?
> > > 
> > > 
> > > Unify the version management in the parent class, which is convenient for publishing and packaging without modifying a version file every time you publish
> > 
> > 
> > Is this compatible with the maven-publish plugin as well as the Apache release procedure?
> 
> Compatible. In the past, `mvn versions:set -DnewVersion=9.1.0` may be used to modify the version number in batches. Now it is only used to modify one place in the pom file. If you use the command, you can use `-Drevision=9.1.0` to modify, eg: `mvnd clean package -Dmaven.test.skip=true -Drevision=8.16.12 -Pall`

Check the full steps about how to do the release, and verify the changes in this PR is OK in those process

https://github.com/apache/skywalking/blob/1256f7059df3affd73c27d1b17a3cf8d7ea6efff/docs/en/guides/How-to-release.md?plain=1#L48

If so I'm glad to approve the changes so that we don't have to change hundreds of lines in every release.

zhaoyunxing92:
> > > > > This is never discussed. Why do you add this?
> > > > 
> > > > 
> > > > Unify the version management in the parent class, which is convenient for publishing and packaging without modifying a version file every time you publish
> > > 
> > > 
> > > Is this compatible with the maven-publish plugin as well as the Apache release procedure?
> > 
> > 
> > Compatible. In the past, `mvn versions:set -DnewVersion=9.1.0` may be used to modify the version number in batches. Now it is only used to modify one place in the pom file. If you use the command, you can use `-Drevision=9.1.0` to modify, eg: `mvnd clean package -Dmaven.test.skip=true -Drevision=8.16.12 -Pall`
> 
> That's not how we do a release, check the full steps about how to do the release, and verify the changes in this PR is OK in those process
> 
> https://github.com/apache/skywalking/blob/1256f7059df3affd73c27d1b17a3cf8d7ea6efff/docs/en/guides/How-to-release.md?plain=1#L48
> 
> If so I'm glad to approve the changes so that we don't have to change hundreds of lines in every release.

Yes, you only need to modify one file for future releases

wu-sheng:
We don't need to change any files manually. We just run one command only. That is my confusion.

zhaoyunxing92:
But the modified files are different, now only one file needs to be modified, even without modifying the command

zhaoyunxing92:
I saw that a bunch of files have to be modified every time I publish. Now I don’t need to modify the files, or I only need to modify one, and this is compatible with your previous ones. This is supported in maven3.5.0

zhaoyunxing92:
> You are not reading clearly about Zhenxu posted previously. [This command](https://github.com/apache/skywalking/blob/1256f7059df3affd73c27d1b17a3cf8d7ea6efff/docs/en/guides/How-to-release.md?plain=1#L48) would do all the changes and even `git tag`/`git commit` from v1-SNAPSHOT to v1 and then v2-SNAPSHOT automatically.

I have read this, but I feel that I can try a new release method. I modified too many files in each previous release, and now I don’t need to modify the files.

wu-sheng:
That is your practice, not an upstream case. 
For us, you are actually adding a new manual file change, rather than an all automatic process.

zhaoyunxing92:
> That is your practice, not an upstream case. For us, you are actually adding a new manual file change, rather than an all automatic process.

He is compatible with the previous release method

pom.xml
```diff
@@ -79,6 +79,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <compiler.version>8</compiler.version>
+        <revision>9.1.0-SNAPSHOT</revision>
```
wu-sheng, line 82: This has to be manually changed to 9.1.0 for release and tag, and then 9.2.0-SNAPSHOT for the next dev stage. 
What do you mean as same as before? You are not an upstream release manager, I don't think you have gone through all Java agent release process.

kezhenxu94: > This has to be manually changed to 9.1.0 for release and tag, and then 9.2.0-SNAPSHOT for the next dev stage.

This doesn't actually need to be manually changed, when you run the release command, you'll be prompted like before about what the version is, what the next SNAPSHOT version is

PR #603, opened 2023-09-01
Fix the issue of createSpan failure caused by invalid request URL in HttpClient 4.x/5.x plugin

xiaqi1210:
### Fix the issue of createSpan failure caused by invalid request URL in HttpClient 4.x/5.x plugin
**bad case:**
uri=http://localhost:8080:8080
In HttpClient 4.x/5.x plugin, isUrl(uri) will return true, but new URL(uri) will throw Exception which will interrupt createSpan.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/httpClient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpClient/v4/HttpClientExecuteInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/httpClient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpClient/v4/HttpClientExecuteInterceptor.java
@@ -42,6 +42,7 @@ import org.apache.skywalking.apm.plugin.httpclient.HttpClientPluginConfig;
 import org.apache.skywalking.apm.util.StringUtil;
 
 public class HttpClientExecuteInterceptor implements InstanceMethodsAroundInterceptor {
+    private static final String ERROR_URI = "/_blank";
 
     @Override
     public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
@@ -60,7 +61,9 @@ public class HttpClientExecuteInterceptor implements InstanceMethodsAroundInterc
         String requestURI = getRequestURI(uri);
         String operationName = requestURI;
         AbstractSpan span = ContextManager.createExitSpan(operationName, contextCarrier, remotePeer);
-
+        if (ERROR_URI.equals(requestURI)) {
+            span.errorOccurred();
+        }
         span.setComponent(ComponentsDefine.HTTPCLIENT);
         Tags.URL.set(span, buildSpanValue(httpHost, uri));
         Tags.HTTP.METHOD.set(span, httpRequest.getRequestLine().getMethod());
@@ -118,7 +121,7 @@ public class HttpClientExecuteInterceptor implements InstanceMethodsAroundInterc
             try {
                 requestPath = new URL(uri).getPath();
             } catch (MalformedURLException e) {
-                requestPath = uri;
+                return ERROR_URI;
             }
             return requestPath != null && requestPath.length() > 0 ? requestPath : "/";
         } else {
--- a/apm-sniffer/apm-sdk-plugin/httpClient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpClient/v4/HttpClientExecuteInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/httpClient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpClient/v4/HttpClientExecuteInterceptor.java
@@ -112,9 +112,14 @@ public class HttpClientExecuteInterceptor implements InstanceMethodsAroundInterc
         activeSpan.log(t);
     }
 
-    private String getRequestURI(String uri) throws MalformedURLException {
+    private String getRequestURI(String uri) {
         if (isUrl(uri)) {
-            String requestPath = new URL(uri).getPath();
+            String requestPath;
+            try {
+                requestPath = new URL(uri).getPath();
+            } catch (MalformedURLException e) {
+                requestPath = uri;
+            }
             return requestPath != null && requestPath.length() > 0 ? requestPath : "/";
         } else {
             return uri;
--- a/apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpClientDoExecuteInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpClientDoExecuteInterceptor.java
@@ -101,9 +101,14 @@ public class HttpClientDoExecuteInterceptor implements InstanceMethodsAroundInte
         activeSpan.log(t);
     }
 
-    private String getRequestURI(String uri) throws MalformedURLException {
+    private String getRequestURI(String uri) {
         if (isUrl(uri)) {
-            String requestPath = new URL(uri).getPath();
+            String requestPath;
+            try {
+                requestPath = new URL(uri).getPath();
+            } catch (MalformedURLException e) {
+                requestPath = uri;
+            }
             return requestPath != null && requestPath.length() > 0 ? requestPath : "/";
         } else {
             return uri;
--- a/apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpClientDoExecuteInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpClientDoExecuteInterceptor.java
@@ -39,6 +39,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 
 public class HttpClientDoExecuteInterceptor implements InstanceMethodsAroundInterceptor {
+    private static final String ERROR_URI = "/_blank";
 
     private static final ILog LOGGER = LogManager.getLogger(HttpClientDoExecuteInterceptor.class);
 
@@ -59,7 +60,9 @@ public class HttpClientDoExecuteInterceptor implements InstanceMethodsAroundInte
         String requestURI = getRequestURI(uri);
         String operationName = requestURI;
         AbstractSpan span = ContextManager.createExitSpan(operationName, contextCarrier, remotePeer);
-
+        if (ERROR_URI.equals(requestURI)) {
+            span.errorOccurred();
+        }
         span.setComponent(ComponentsDefine.HTTPCLIENT);
         Tags.URL.set(span, buildURL(httpHost, uri));
         Tags.HTTP.METHOD.set(span, httpRequest.getMethod());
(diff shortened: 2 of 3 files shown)


wu-sheng:
Can't we use some methods to check this but have to use `try/catch`? This is not a good practice in Java agent.
Another thing is, shouldn't we use `/` instead of the whole URL? After all, span name should be URI rather then URL, which may not be friendly in statistics.

xiaqi1210:
> Can't we use some methods to check this but have to use `try/catch`? This is not a good practice in Java agent. Another thing is, shouldn't we use `/` instead of the whole URL? After all, span name should be URI rather then URL, which may not be friendly in statistics.

Is not using try/catch considered for performance? What I can think of is that I can use substring to get the content after the first "/" after "://" in the URL as the URI. If I can't get it, the default is "/". What do you think of this?

xiaqi1210:
I also did some testing and compared the performance between `try..catch` and `substring`, as shown below:

image
image

wu-sheng:
This isn't correct to test performance. Please run benchmark through jmh.

And reflection performance is determined by different JDK.

wu-sheng:
> > Can't we use some methods to check this but have to use `try/catch`? This is not a good practice in Java agent. Another thing is, shouldn't we use `/` instead of the whole URL? After all, span name should be URI rather then URL, which may not be friendly in statistics.
> 
> 
> 
> Is not using try/catch considered for performance? What I can think of is that I can use substring to get the content after the first "/" after "://" in the URL as the URI. If I can't get it, the default is "/". What do you think of this?

I think all path has http:// prefix, why they are good?

xiaqi1210:
> > > Can't we use some methods to check this but have to use `try/catch`? This is not a good practice in Java agent. Another thing is, shouldn't we use `/` instead of the whole URL? After all, span name should be URI rather then URL, which may not be friendly in statistics.
> > 
> > 
> > Is not using try/catch considered for performance? What I can think of is that I can use substring to get the content after the first "/" after "://" in the URL as the URI. If I can't get it, the default is "/". What do you think of this?
> 
> I think all path has http:// prefix, why they are good?

I'm not quite clear on your point. In the plugin, it only checks if the URL starts with http or https, but in my project I have encountered cases where the URL contains 2 port numbers, which causes errors in new URL().

wu-sheng:
Please provide HTTP standard that, two ports are legal.
I am highly doubting that.
If this is your private protocol implementation, we can't accept this.

xiaqi1210:
>

xiaqi1210:
> Please provide HTTP standard that, two ports are legal. I am highly doubting that. If this is your private protocol implementation, we can't accept this.

Actually in my project, that URL was from a configuration that had some issues. But eventually due to this, trace data for the whole chain was lost. I think we shouldn't allow an incorrect configuration to affect trace data for the entire chain.

xiaqi1210:
Also, I think new URL() is a very heavy operation, while substring is more lightweight, which should also improve performance.

wu-sheng:
If this is for illegal case, I think we should create a span as /_blank and log as an error.

wu-sheng:
> Also, I think new URL() is a very heavy operation, while substring is more lightweight, which should also improve performance.

You could propose a new solution, but substring is not an easy way to resolve URL always correctly.

xiaqi1210:
> > Also, I think new URL() is a very heavy operation, while substring is more lightweight, which should also improve performance.
> 
> You could propose a new solution, but substring is not an easy way to resolve URL always correctly.

First, the HTTP URL specification is mentioned in [RFC1738](https://www.rfc-editor.org/rfc/rfc1738.txt) (3.3. HTTP), with the format as follows:
```
http://<host>:<port>/<path>?<searchpart>
```
There won't be '/' in host and port, so finding the first '/' after "//" and taking the substring after should always work. 
The only special case is: http://<host>:<port>, where the first '/' is not found, then the default uri=/ can be used.
I can't think of any other special cases.

On the other hand, for invalid URL scenarios, I don't think it's important to validate if the URL is valid before createSpan, because an invalid URL would definitely throw an exception when making the HTTP request, and the error span would be logged in `handleMethodException`. Also, invalid HTTP URLs should be very rare cases, so doing it this way wouldn't have much impact on performance either.

wu-sheng:
Invalid URI as span name would not break the analysis, but it will cost on metrics analysis for entry span. That is the concern.

wu-sheng:
My point is, the URL should not be illegal in any product env. So, this performance impact is not our concern.
We could keep the span name as `_blank` and keep the original URL in the span tag, status as failing as well.

xiaqi1210:
> Invalid URI as span name would not break the analysis, but it will cost on metrics analysis for entry span. That is the concern.

I don't know how this is related to entry span. The HttpClient plugin only creates exit spans, and in case of invalid URLs, the request won't reach the remote end, so it does not involve entry span.

wu-sheng:
> > Invalid URI as span name would not break the analysis, but it will cost on metrics analysis for entry span. That is the concern.
> 
> I don't know how this is related to entry span. The HttpClient plugin only creates exit spans, and in case of invalid URLs, the request won't reach the remote end, so it does not involve entry span.

My fault. But still, generally, we still only recommend URI or URI style operation name when it involves with HTTP.

xiaqi1210:
So your point is:
If URL is invalid, set the span's operationName to `/_blank` and mark it as error.
If URL is valid, follow original logic.

Is my understanding correct?

wu-sheng:
Yes.

wu-sheng:
@xiaqi1210 Would you able to finish this patch?

apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpClientDoExecuteInterceptor.java
```diff
@@ -101,9 +101,14 @@ public void handleMethodException(EnhancedInstance objInst, Method method, Objec
         activeSpan.log(t);
     }
 
-    private String getRequestURI(String uri) throws MalformedURLException {
+    private String getRequestURI(String uri) {
```
wuwen5, line 104: I suggest referring to the implementation of `v3` version and directly using `URI#getPath`

e.g.
```java
private String getRequestURI(URI uri) {
        if (isUrl(uri.toString())) {
            String requestPath = uri.getPath();
            return requestPath != null && requestPath.length() > 0 ? requestPath : "/";
        } else {
            return uri.toString();
        }
    }
```

wu-sheng: Is this working for you? I thought you are using illegal path, such as two ports. 
This is just to separate URI and URL.

wuwen5: I agree with what was said above "span name should be URI rather then URL",
I don't think there is a need to instantiate `URL` here.

The use of urls should only be in this area `Tags.URL.set(span, buildURL(httpHost, uri.toString()));`

wu-sheng: That works. My point is reminding your original issue only.

xiaqi1210:
> @xiaqi1210 Would you able to finish this patch?

The difficulty here lies in how to determine if a URL is valid. The only accurate way I can think of is trying new URL(). Another way is using regular expressions, but that risks filtering out some originally valid URLs. So I'm still thinking of other ways to validate.

xiaqi1210:
Or I could focus on filtering just the specific case of two port numbers that I encountered, and improve to cover other cases in the future.

apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpClientDoExecuteInterceptor.java
xiaqi1210: > I suggest referring to the implementation of `v3` version and directly using `URI#getPath`
> 
> e.g.
> 
> ```java
> private String getRequestURI(URI uri) {
>         if (isUrl(uri.toString())) {
>             String requestPath = uri.getPath();
>             return requestPath != null && requestPath.length() > 0 ? requestPath : "/";
>         } else {
>             return uri.toString();
>         }
>     }
> ```

I didn't find a way to directly get the URI in HttpClient 4.x/5.x plugin.

wu-sheng:
I am confused, URI provided a way to catch the error. Why don't you want to use it?
Like we discussed, there is no performance concern as exception is only for illegal cases.

apm-sniffer/apm-sdk-plugin/httpclient-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpclient/v5/HttpClientDoExecuteInterceptor.java
wuwen5: line#58  `String uri = httpRequest.getUri().toString();`

This is originally the `URI`, and in this use case, the `URI#getPath` does not have any exceptions, 
such as:` http://ip:port:port/ ` will be returned is an empty string, why is it necessary to forcibly convert it to a `URL` to obtain the path here?

If it is necessary to handle the incorrect `URL`, it should be done in the `buildURL` method.

xiaqi1210:
> I am confused, URI provided a way to catch the error. Why don't you want to use it? Like we discussed, there is no performance concern as exception is only for illegal cases.

I'm afraid I didn't quite understand. Are you suggesting to change `new URL(uri)` to `new URI(uri)`?

apm-sniffer/apm-sdk-plugin/httpClient-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/httpClient/v4/HttpClientExecuteInterceptor.java
```diff
@@ -112,9 +112,14 @@ public void handleMethodException(EnhancedInstance objInst, Method method, Objec
...
 
-    private String getRequestURI(String uri) throws MalformedURLException {
+    private String getRequestURI(String uri) {
         if (isUrl(uri)) {
-            String requestPath = new URL(uri).getPath();
+            String requestPath;
+            try {
+                requestPath = new URL(uri).getPath();
+            } catch (MalformedURLException e) {
+                requestPath = uri;
```
wu-sheng, line 121: @xiaqi1210 Let's put in this way. First of all, are you only facing issues when the URL is illegal? Rather than other things. Let's focus on the issue first.

xiaqi1210: ok。So here you think try..catch is fine, just that in the catch block it should be changed to set `requestPath="_blank"`, right?

wu-sheng: Yes, I am.

xiaqi1210: ok, I will push a new version of code later.

wu-sheng:
Please resolve the conflicts.

wu-sheng (approved):
I am good with this, as this is just to avoid exceptions in the plugin codes. @wuwen5 Is this good for you?

PR #607, opened 2023-09-06
Optimize ElasticSearch 6.x 7.x plugin compatibility

shichaoyuan:
- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/11293

The parameters of the analyze method are different between 6.x and 7.x versions:
* 6.x org.elasticsearch.action.admin.indices.analyze.AnalyzeRequest
* 7.x org.elasticsearch.client.indices.AnalyzeRequest


shichaoyuan:
> Could you update the test scenarios covering this case?

add function test in `elasticsearch-7.x-scenario`

wu-sheng:
It seems the test fails on elasticsearch-7 case, please recheck.

https://github.com/apache/skywalking-java/actions/runs/6107504935/job/16576014494?pr=607

shichaoyuan:
> It seems the test fails on elasticsearch-7 case, please recheck.
> 
> https://github.com/apache/skywalking-java/actions/runs/6107504935/job/16576014494?pr=607

fixed

shichaoyuan:
> Still failing.

https://github.com/apache/skywalking-java/blob/bc7447a2c2de1c6f11ba8bd6f35dd64a2c474746/test/plugin/scenarios/elasticsearch-7.x-scenario/src/main/java/org/apache/skywalking/apm/testcase/elasticsearch/controller/CaseController.java#L43C5-L46

```
    public String healthcheck() throws Exception {
        restHighLevelClientCase.healthCheck();
        return "Success";
    }
```

It looks like `healthcheck` is mixed into the result check.

I think `restHighLevelClientCase.healthCheck()` can be removed

wu-sheng:
Healthcheck is for determining the app and dependencies are standing by. I think it is necessary. What is the cause of this failing?

shichaoyuan:
> Healthcheck is for determining the app and dependencies are standing by. I think it is necessary. What is the cause of this failing?

```
actual data: {
    "segmentItems": [
      {
        "serviceName": "elasticsearch-7.x-scenario",
        "segmentSize": "2",
        "segments": [
          {
            "segmentId": "1e84afbcaf3541af8df1be94b22cd050.47.16940869262530000",
            "spans": [
              {
                "operationName": "Elasticsearch/Health",
                "parentSpanId": "0",
                "spanId": "1",
                "spanLayer": "Database",
                "tags": [
                  {
                    "key": "db.type",
                    "value": "Elasticsearch"
                  }
                ],
                "startTime": "1694086926351",
                "endTime": "1694086926610",
                "componentId": "77",
                "isError": "false",
                "spanType": "Exit",
                "peer": "elasticsearch-server-7.x:9200",
                "skipAnalysis": "false"
              },
              {
                "operationName": "HEAD:/elasticsearch-case/case/healthCheck",
                "parentSpanId": "-1",
                "spanId": "0",
                "spanLayer": "Http",
                "tags": [
                  {
                    "key": "url",
                    "value": "http://localhost:8080/elasticsearch-case/case/healthCheck"
                  },
                  {
                    "key": "http.method",
                    "value": "HEAD"
                  },
                  {
                    "key": "http.status_code",
                    "value": "200"
                  }
                ],
                "startTime": "1694086926255",
                "endTime": "1694086926638",
                "componentId": "1",
                "isError": "false",
                "spanType": "Entry",
                "peer": "",
                "skipAnalysis": "false"
              }
            ]
          },
          {
            "segmentId": "1e84afbcaf3541af8df1be94b22cd050.34.16940869273440000",
            "spans": [
              {
                "operationName": "/76b8e333-ed41-42e7-bac8-fc44b3b53088/_refresh?ignore_throttled\u003dfalse\u0026ignore_unavailable\u003dfalse\u0026expand_wildcards\u003dopen\u0026allow_no_indices\u003dtrue",
                "parentSpanId": "0",
                "spanId": "1",
                "spanLayer": "Http",
                "tags": [
                  {
                    "key": "url",
                    "value": "/76b8e333-ed41-42e7-bac8-fc44b3b53088/_refresh?ignore_throttled\u003dfalse\u0026ignore_unavailable\u003dfalse\u0026expand_wildcards\u003dopen\u0026allow_no_indices\u003dtrue"
                  },
                  {
                    "key": "http.method",
                    "value": "POST"
                  },
                  {
                    "key": "http.status_code",
                    "value": "200"
                  }
                ],
                "startTime": "1694086927346",
                "endTime": "1694086927453",
                "componentId": "26",
                "isError": "true",
                "spanType": "Exit",
                "peer": "elasticsearch-server-7.x:9200",
                "skipAnalysis": "false"
              },
              {
                "operationName": "HttpAsyncClient/local",
                "parentSpanId": "-1",
                "spanId": "0",
                "spanLayer": "Http",
                "refs": [
                  {
                    "parentEndpoint": "GET:/elasticsearch-case/case/elasticsearch",
                    "networkAddress": "",
                    "refType": "CrossThread",
                    "parentSpanId": 0,
                    "parentTraceSegmentId": "1e84afbcaf3541af8df1be94b22cd050.48.16940869266830000",
                    "parentServiceInstance": "f07a08c04[799](https://github.com/apache/skywalking-java/actions/runs/6109031224/job/16579783820#step:4:805)42e3bad96604216178ad@172.18.0.3",
                    "parentService": "elasticsearch-7.x-scenario",
                    "traceId": "1e84afbcaf3541af8df1be94b22cd050.48.16940869266830001"
                  }
                ],
                "startTime": "1694086927344",
                "endTime": "1694086927453",
                "componentId": "26",
                "isError": "false",
                "spanType": "Local",
                "peer": "",
                "skipAnalysis": "false"
              }
            ]
          }
        ]
      }
    ],
```

It seems that there is an abnormal `"operationName": "HttpAsyncClient/local"`

wu-sheng:
This is something I can't directly tell the reason. If it exists, it is traced. So, maybe some parts of the codes are using or calling `HttpAsyncClient`? It indicates a cross-thread ref, so is there something(context snapshot) that is read and continues? 
I think you have to debug the codes.

shichaoyuan:
> 

image

Found the reason.

The `refresh` method is not enhanced in the ElasticSearch plugin.

Let me deal with the `refresh` method

shichaoyuan:
image
An exception is thrown here, I will fix it again

wu-sheng:
This could be an illegal string?

shichaoyuan:
> This could be an illegal string?

the `BytesReference` returned by `mappings()` method was changed.

7.3
```

public abstract class BytesReference implements Comparable<BytesReference>, ToXContentFragment 
```

7.17
```
package org.elasticsearch.common.bytes;
public interface BytesReference extends Comparable<BytesReference>, ToXContentFragment 
```

wu-sheng:
Are these new instrumentations or new versions you are adding to the tests? This seems to be 7.x relative issues, and the title of the PR indicates 6.x plugin bug. Could you update the title and change logs accordingly?

shichaoyuan:
> Are these new instrumentations or new versions you are adding to the tests? This seems to be 7.x relative issues, and the title of the PR indicates 6.x plugin bug. Could you update the title and change logs accordingly?

image

added more versions in `support-version.list`

wu-sheng:
Still have errors.

shichaoyuan:
> Still have errors.

```
  Pulling elasticsearch-server-7.x (elasticsearch:7.10.2)...
  manifest for elasticsearch:7.10.2 not found: manifest unknown: manifest unknown
  docker startup failure!
  Error response from daemon: No such image: elasticsearch:7.10.2
```

Let me see where the image was pulled from.

wu-sheng:
I remember Elasticsearch 7.10 changed the license, so maybe it needs to download from a new docker hub repo? If so, you have to separate the test cases

apm-sniffer/apm-sdk-plugin/elasticsearch-common/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+
+    <name>elasticsearch-common</name>
+    <url>http://maven.apache.org</url>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+    <dependencies>
+    </dependencies>
```
wu-sheng, line 38: I think this is not necessary.

apm-sniffer/apm-sdk-plugin/elasticsearch-common/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+
+    <artifactId>elasticsearch-common</artifactId>
+    <packaging>jar</packaging>
+
+    <name>elasticsearch-common</name>
+    <url>http://maven.apache.org</url>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
```
wu-sheng, line 35: Do we need this?

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/TransportActionNodeProxyExecuteMethodsInterceptor.java
```diff
@@ -76,7 +76,15 @@ public void handleMethodException(EnhancedInstance objInst, Method method, Objec
...
-        EnhancedInstance actions = (EnhancedInstance) allArguments[2];
+        EnhancedInstance actions;
+        if (allArguments.length > 2) {
+            // (Settings settings, GenericAction<Request, Response> action, TransportService transportService)
+            actions = (EnhancedInstance) allArguments[2];
+        } else {
+            // 7.11 +
+            // (ActionType<Response> action, TransportService transportService)
+            actions = (EnhancedInstance) allArguments[1];
+        }
```
wu-sheng, line 87: This is not a proper way to fix this, please use different interceptors with different instrumentations.

apm-sniffer/apm-sdk-plugin/elasticsearch-common/pom.xml
shichaoyuan: removed

apm-sniffer/apm-sdk-plugin/elasticsearch-6.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/elasticsearch/v6/interceptor/TransportActionNodeProxyExecuteMethodsInterceptor.java
shichaoyuan: fixed

shichaoyuan:
The `XContentBuilder` moved the package location in version 7.16, and I fixed the test cases again.

shichaoyuan:
image
image
This image exists, it may be due to an occasional pull failure.

wu-sheng:
I tried to re-run, if you pull too many versions of images, I have concerns whether the disk volume in the CI env is enough.

wu-sheng (approved):

PR #608, opened 2023-09-14
fix: Fix an issue with the httpasyncclient component where the isErro…

w2dp:
### Fix <Fix an issue with the httpasyncclient component where the isError state is incorrect.>
Normally, HTTP status codes below 400 should be marked as isError=false. However, due to a previous code migration, the isError state in the httpasyncclient component became inaccurate. It incorrectly marked isError=true even for HTTP status codes below 400.


wu-sheng:

w2dp:
> Please update the changes.md in the root.

Added as requested.

wu-sheng (approved):

PR #609, opened 2023-09-14
add max length configuration for service_name and instance_name

life-:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. close https://github.com/apache/skywalking/issues/11327.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,8 @@ Release Notes.
 * Fix hbase onConstruct NPE in the file configuration scenario
 * Fix the issue of createSpan failure caused by invalid request URL in HttpClient 4.x/5.x plugin
 * Optimize ElasticSearch 6.x 7.x plugin compatibility
-* Add max length configuration in agent.config file for service_name and instance_name
+* Support customization for the length limitation of string configurations
+* Add max length configurations in `agent.config` file for service_name and instance_name
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ Release Notes.
 * Fix hbase onConstruct NPE in the file configuration scenario
 * Fix the issue of createSpan failure caused by invalid request URL in HttpClient 4.x/5.x plugin
 * Optimize ElasticSearch 6.x 7.x plugin compatibility
+* Add max length configuration in agent.config file for service_name and instance_name
 
 #### Documentation
 
--- a/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
+++ b/apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
@@ -89,7 +89,7 @@ public class ConfigInitializer {
                             int lengthLimited = lengthDefine.value();
                             String lengthKey = String.format("%s#length", configKey);
                             if (properties.containsKey(lengthKey)) {
-                                lengthLimited = Math.max(lengthLimited, Integer.valueOf(properties.getProperty(lengthKey, "0")));
+                                lengthLimited = Integer.valueOf(properties.getProperty(lengthKey, "0"));
                             }
                             if (propertyValue.length() > lengthLimited) {
                                 StringUtil.cut(propertyValue, lengthLimited);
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -8,6 +8,7 @@ Release Notes.
 * Fix hbase onConstruct NPE in the file configuration scenario
 * Fix the issue of createSpan failure caused by invalid request URL in HttpClient 4.x/5.x plugin
 * Optimize ElasticSearch 6.x 7.x plugin compatibility
+* Add max length configuration in agent.config file for service_name and instance_name
```
wu-sheng, line 11:
```suggestion
* Support customization for the length limitation of string configurations
* Add max length configurations in `agent.config` file for service_name and instance_name
```

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
```diff
@@ -85,9 +85,16 @@ private static void initNextLevel(Properties properties, Class<?> recentConfigTy
...
                         final Length lengthDefine = field.getAnnotation(Length.class);
-                        if (lengthDefine != null && propertyValue.length() > lengthDefine.value()) {
-                            StringUtil.cut(propertyValue, lengthDefine.value());
-                            System.err.printf("The config value will be truncated , because the length max than %d : %s -> %s%n", lengthDefine.value(), configKey, propertyValue);
+                        if (lengthDefine != null) {
+                            int lengthLimited = lengthDefine.value();
+                            String lengthKey = String.format("%s#length", configKey);
+                            if (properties.containsKey(lengthKey)) {
+                                lengthLimited = Math.max(lengthLimited, Integer.valueOf(properties.getProperty(lengthKey, "0")));
+                            }
```
wu-sheng, line 93: I think `max` would be a little tricky. As we provided a configuration item, I would prefer to use the configuration value to override static codes as always. 
Could you explain why do you code as `max`?

life-: The default value is 50, which means it is the suggested value. If the configuration is smaller than it which means most of the service names or instance names will be cut especially in k8s environment.  

So we suggest the configuration for the length is better bigger than the default value.

wu-sheng: You are adding a feature to provide customization, I can't see why bigger is acceptable, but smaller isn't.
Your explanation seems doesn't make sense.

All length relative configurations are accepting value length in both ways. 

Please update the codes to provide an override capability, rather than within this condition.

life-: ok, I will change code to prefer user's configuration.

life-: @wu-sheng  done

apm-commons/apm-util/src/main/java/org/apache/skywalking/apm/util/ConfigInitializer.java
```diff
@@ -85,9 +85,16 @@ private static void initNextLevel(Properties properties, Class<?> recentConfigTy
...
                         // Convert the value into real type
                         final Length lengthDefine = field.getAnnotation(Length.class);
-                        if (lengthDefine != null && propertyValue.length() > lengthDefine.value()) {
-                            StringUtil.cut(propertyValue, lengthDefine.value());
-                            System.err.printf("The config value will be truncated , because the length max than %d : %s -> %s%n", lengthDefine.value(), configKey, propertyValue);
+                        if (lengthDefine != null) {
+                            int lengthLimited = lengthDefine.value();
+                            String lengthKey = String.format("%s#length", configKey);
+                            if (properties.containsKey(lengthKey)) {
+                                lengthLimited = Integer.valueOf(properties.getProperty(lengthKey, "0"));
```
wu-sheng, line 92: I think we are better to try catch the `Integer#valueOf` and `system.err` if type cast fails. WDYT?

wu-sheng (approved):
LGTM. Thanks for the patient

PR #615, opened 2023-09-24
Fix peer when configured as domain for rabbitmq plugin.

jmjoy:

Previous rabbitmq plugin get peer by `InetAddress.toString().replace("/", "")`, The format of `InetAddress.toString()` is `DOMAIN/IP`, so, when the rabbitmq address configured as IP, it's ok, but when the address configured as Domain, isn't ok, the peer will become `DOMAINIP:PORT`, such as `localhost127.0.0.1:5672`.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/rabbitmq-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/ChannelNConstructorInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/rabbitmq-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/ChannelNConstructorInterceptor.java
@@ -28,12 +28,12 @@ public class ChannelNConstructorInterceptor implements InstanceConstructorInterc
     public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
         Connection connection = (Connection) allArguments[0];
 
-        String[] segments = connection.getAddress().toString().split("/");
+        String[] parts = connection.getAddress().toString().split("/");
         String address;
-        if (StringUtil.isNotEmpty(segments[0])) {
-            address = segments[0];
-        } else if (segments.length >= 2 && StringUtil.isNotEmpty(segments[1])) {
-            address = segments[1];
+        if (StringUtil.isNotEmpty(parts[0])) {
+            address = parts[0];
+        } else if (parts.length >= 2 && StringUtil.isNotEmpty(parts[1])) {
+            address = parts[1];
         } else {
             address = "";
         }
--- a/apm-sniffer/apm-sdk-plugin/rabbitmq-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/ChannelNConstructorInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/rabbitmq-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/ChannelNConstructorInterceptor.java
@@ -21,12 +21,24 @@ package org.apache.skywalking.apm.plugin.rabbitmq;
 import com.rabbitmq.client.Connection;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+import org.apache.skywalking.apm.util.StringUtil;
 
 public class ChannelNConstructorInterceptor implements InstanceConstructorInterceptor {
     @Override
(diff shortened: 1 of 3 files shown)


apm-sniffer/apm-sdk-plugin/rabbitmq-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/ChannelNConstructorInterceptor.java
```diff
@@ -21,12 +21,24 @@
...
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+import org.apache.skywalking.apm.util.StringUtil;
 
 public class ChannelNConstructorInterceptor implements InstanceConstructorInterceptor {
     @Override
     public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
         Connection connection = (Connection) allArguments[0];
-        String url = connection.getAddress().toString().replace("/", "") + ":" + connection.getPort();
+
+        String[] segments = connection.getAddress().toString().split("/");
```
wu-sheng, line 31: Please use parts. Segments have a special meaning in the agent.

apm-sniffer/apm-sdk-plugin/rabbitmq-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/ChannelNConstructorInterceptor.java
```diff
@@ -21,12 +21,24 @@
...
-        String url = connection.getAddress().toString().replace("/", "") + ":" + connection.getPort();
+
+        String[] parts = connection.getAddress().toString().split("/");
+        String address;
+        if (StringUtil.isNotEmpty(parts[0])) {
+            address = parts[0];
+        } else if (parts.length >= 2 && StringUtil.isNotEmpty(parts[1])) {
+            address = parts[1];
+        } else {
+            address = "";
```
wu-sheng, line 38: If you can't resolve the address, this should not be an empty address. 
`connection.getAddress().toString()` could be used.

apm-sniffer/apm-sdk-plugin/rabbitmq-plugin/src/main/java/org/apache/skywalking/apm/plugin/rabbitmq/ChannelNConstructorInterceptor.java
```diff
@@ -21,12 +21,24 @@
...
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+import org.apache.skywalking.apm.util.StringUtil;
 
 public class ChannelNConstructorInterceptor implements InstanceConstructorInterceptor {
     @Override
     public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
         Connection connection = (Connection) allArguments[0];
-        String url = connection.getAddress().toString().replace("/", "") + ":" + connection.getPort();
+
+        String[] parts = connection.getAddress().toString().split("/");
```
wu-sheng, line 31: Could you explain what do you mean `split` through `/`, then read <0> or <1>? The UT seems can't represent the use cases.

jmjoy: For `InetAddress.getByName("localhost")`, the result of `InetAddress.toString()` is `localhost/127.0.0.1`.

For `InetAddress.getByName("127.0.0.1")`, the result of `InetAddress.toString()` is `/127.0.0.1`.

I also try `InetAddress.getHostName()` or `InetAddress.getHostAddress()`, but he former must return domain, while the latter must return IP.

So, I choose to split the result of `.toString()`.

wu-sheng: As a peer, I don't think we need to resolve anything, it doesn't matter. We just need an unique address(DNS, IP, alias name, domain name, anything) to represent a remote service.
That is all.

wu-sheng: For rabbitMQ case, it would be only used in `virtual MQ service` name on the UI. This should be the only place you could see the point of the data. At the kernel analysis level, it is used in network alias mapping. You could find this table/index in your storage.

PR #621, opened 2023-10-12
Fix oracle lower case tns url parse error. (#11399)

Builder34:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes [[Bug] oracle lower case tns url parse err: skywalking#11399](https://github.com/apache/skywalking/issues/11399)

- To compatible upper and lower case Oracle TNS url parse.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Release Notes.
 * Add max length configurations in `agent.config` file for service_name and instance_name
 * Optimize spring-cloud-gateway 2.1.x, 3.x witness class.
 * Support report MongoDB instance info in Mongodb 4.x plugin.
+* To compatible upper and lower case Oracle TNS url parse.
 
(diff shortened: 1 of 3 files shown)


wu-sheng:
You should update the changes file. You clicked the box in the PR, but didn't actually do.

Builder34:
https://www.caucho.com/resin-4.0/  dead Link check failed，do i need fix it ?

wu-sheng:
Is the website changed? Or it is just down?

Builder34:
I am not sure; it cannot be accessed, and no new accessible address was found.

wu-sheng (approved):

wu-sheng:
The whole https://www.caucho.com is down. I am not sure. Merged your PR. It should not be your concern.

Thanks for your contribution.

PR #622, opened 2023-10-13
Support collect ZGC memory pool metrics

Z-Beatles:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Release Notes.
 * Add max length configurations in `agent.config` file for service_name and instance_name
 * Optimize spring-cloud-gateway 2.1.x, 3.x witness class.
 * Support report MongoDB instance info in Mongodb 4.x plugin.
+* Support collect ZGC memory pool metrics.
 
 #### Documentation
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@ Release Notes.
 * Support report MongoDB instance info in Mongodb 4.x plugin.
 * To compatible upper and lower case Oracle TNS url parse.
 * Fix config length limitation.
-* Support collect ZGC memory pool metrics.
+* Support collecting ZGC memory pool metrics. Require OAP 9.7.0 to support these new metrics.
 
 
 #### Documentation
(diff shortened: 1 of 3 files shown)


wu-sheng:
So, this relies on the protocol update.
Java 9.1 can't release before OAP 9.7, right?

And do we expect UI setting change?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/memorypool/ZGCCollectorModule.java
```diff
@@ -0,0 +1,70 @@
...
+        for (MemoryPoolMXBean bean : beans) {
+            String name = bean.getName();
+            PoolType type;
+            if (name.equals("ZHeap")) {
+                type = PoolType.ZHEAP_USAGE;
+            } else if (name.equals("Metaspace")) {
+                type = PoolType.METASPACE_USAGE;
+            } else if (name.equals("Compressed Class Space")) {
+                type = PoolType.COMPRESSED_CLASS_SPACE_USAGE;
+            } else if (name.equals("CodeHeap 'non-nmethods'")) {
```
wu-sheng, line 49: Is the single quote from official definition? It is strange to read.

Z-Beatles:
> So, this relies on the protocol update. Java 9.1 can't release before OAP 9.7, right?
> 
> And do we expect UI setting change?



> So, this relies on the protocol update. Java 9.1 can't release before OAP 9.7, right?
> 
> And do we expect UI setting change?

yes, i will pull a new request on main repo later for UI setting change.

wu-sheng:
Thread pool data is meter oriented.

https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/meter-analyzer-config/threadpool.yaml

I can see JVM metrics don't include pool metric.

https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/oal/java-agent.oal

Could you share which part of OAP would be relative?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/memorypool/ZGCCollectorModule.java
Z-Beatles: Yes, as we can see i added a panel `JVM Memory Detail (MB)` on UI to show memory pool detail  after `JVM Memory (MB)`.
image

wu-sheng (dismissed):

CHANGES.md
```diff
@@ -15,6 +15,8 @@ Release Notes.
 * Support report MongoDB instance info in Mongodb 4.x plugin.
 * To compatible upper and lower case Oracle TNS url parse.
 * Fix config length limitation.
+* Support collect ZGC memory pool metrics.
```
wu-sheng, line 18:
```suggestion
* Support collecting ZGC memory pool metrics. Require OAP 9.7.0 to support these new metrics.
```

wu-sheng (approved):

PR #625, opened 2023-10-21
Add a netty-4.1.x plugin to trace http

peachisai:
### Add an agent plugin to support <framework name>
- [x] Add a logo in [the UI repo](https://github.com/apache/skywalking-rocketbot-ui/tree/master/src/views/components/topology/assets)


wu-sheng:
Before any review, I think we need a clearance from you, what is netty plugin.
Netty is a widely used TCP level framework. So in theory, there isn't a general solution to fit it, as TCP doesn't have a header to inject.

peachisai:
> Before any review, I think we need a clearance from you, what is netty plugin. Netty is a widely used TCP level framework. So in theory, there isn't a general solution to fit it, as TCP doesn't have a header to inject.

Yes, it is only tracing HTTP now. If I have rest of the time, I will provide other protocol tracings. So, is it acceptable?

wu-sheng:
Then you should be clear for all relative codes and docs.
This is only netty-http

peachisai:
> Then you should be clear for all relative codes and docs. This is only netty-http

ok i will change those

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -242,4 +242,6 @@ public class ComponentsDefine {
     public static final OfficialComponent AEROSPIKE = new OfficialComponent(149, "Aerospike");
 
     public static final OfficialComponent NACOS = new OfficialComponent(150, "Nacos");
+
+    public static final OfficialComponent NETTY = new OfficialComponent(151, "Netty");
```
wu-sheng, line 246: This should be updated as well.

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -25,6 +25,7 @@ metrics based on the tracing data.
   * [Jersey REST framework](https://github.com/eclipse-ee4j/jersey) 2.x -> 3.x
   * [Grizzly](https://github.com/eclipse-ee4j/grizzly) 2.3.x -> 4.x
   * [WebSphere Liberty](https://github.com/OpenLiberty/open-liberty) 23.x
+  * [Netty-http](https://github.com/netty/netty) 4.1.x
```
wu-sheng, line 28:
```suggestion
  * [Netty HTTP](https://github.com/netty/netty) 4.1.x
```

And only 4.1, no 4.0.x?

test/plugin/scenarios/netty-http-4.1.x-scenario/support-version.list
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+4.1.25.Final
+4.1.51.Final
+4.1.91.Final
```
wu-sheng, line 19: Could you share why these versions? These are randomly picked or for some reasons?

wu-sheng:
The plugin name is not consistent with the doc. CI fails

```
--- /home/runner/work/skywalking-java/skywalking-java/tools/plugin/md-javaagent-plugin-list.txt	2023-10-21 12:54:33.665221539 +0000
+++ /home/runner/work/skywalking-java/skywalking-java/tools/plugin/genernate-javaagent-plugin-list.txt	2023-10-21 12:54:33.65[7](https://github.com/apache/skywalking-java/actions/runs/6596772865/job/17924511365?pr=625#step:5:8)221426 +0000
@@ -95 +94,0 @@
-netty-http-4.1.x
```

test/plugin/scenarios/netty-http-4.1.x-scenario/support-version.list
peachisai: just randomly picked,should i reduce them?

wu-sheng: It is fine. I am just wondering whether there are some unsupported cases. Could you submit more to verify and exclude most in the final commits?

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
peachisai: > This should be updated as well.

done

test/plugin/scenarios/netty-http-4.1.x-scenario/support-version.list
peachisai: > It is fine. I am just wondering whether there are some unsupported cases. Could you submit more to verify and exclude most in the final commits?

no problem

docs/en/setup/service-agent/java-agent/Supported-list.md
peachisai: > And only 4.1, no 4.0.x?

When I was coding, I only considered the 4.1.x version. Just now I  included support for the 4.0.x version as well. There is something wrong with it, and I will check it.

wu-sheng:
I didn't check the failed test cases. I feel as many other frameworks are using netty, you newly added a plugin for it could change the expected tracing data(spans or something). 
You should check the fail reason, if the reason is as I mentioned above, we should consider to put netty HTTP plugin as optional only.

peachisai:
> I didn't check the failed test cases. I feel as many other frameworks are using netty, you newly added a plugin for it could change the expected tracing data(spans or something). You should check the fail reason, if the reason is as I mentioned above, we should consider to put netty HTTP plugin as optional only.

I am trying to solve these issues. It may take a little longer

wu-sheng:
Take your time. We are not in urgent.

peachisai:
> I didn't check the failed test cases. I feel as many other frameworks are using netty, you newly added a plugin for it could change the expected tracing data(spans or something). You should check the fail reason, if the reason is as I mentioned above, we should consider to put netty HTTP plugin as optional only.

I have fixed some issues, and the other are just segment size errors.I agree to put the plugin into optional plugins. Some frameworks, such as Vert.x and Spring Cloud Gateway, use Netty as an HTTP server and they also have their own plugins. This could lead to duplicate collections.

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -25,6 +25,7 @@ metrics based on the tracing data.
   * [Jersey REST framework](https://github.com/eclipse-ee4j/jersey) 2.x -> 3.x
   * [Grizzly](https://github.com/eclipse-ee4j/grizzly) 2.3.x -> 4.x
   * [WebSphere Liberty](https://github.com/OpenLiberty/open-liberty) 23.x
+  * [Netty Http](https://github.com/netty/netty) 4.1.x
```
wu-sheng, line 28:
```suggestion
  * [Netty HTTP](https://github.com/netty/netty) 4.1.x
```

apm-sniffer/optional-plugins/netty-http-4.1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/netty/http/define/AbstractNettyInstrumentation.java
```diff
@@ -0,0 +1,35 @@
...
+public abstract class AbstractNettyInstrumentation extends ClassEnhancePluginDefine {
+
+    @Override
+    public StaticMethodsInterceptPoint[] getStaticMethodsInterceptPoints() {
+        return new StaticMethodsInterceptPoint[0];
+    }
+
+    @Override
+    protected String[] witnessClasses() {
+        return new String[]{"io.netty.handler.codec.compression.SnappyFrameEncoder"};
```
wu-sheng, line 33: Is this witness class for locking 4.x netty?

peachisai: > Is this witness class for locking 4.x netty?

only 4.1.x

apm-sniffer/optional-plugins/netty-http-4.1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/netty/http/ChannelPipelineRemoveLastInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+public class ChannelPipelineRemoveLastInterceptor implements InstanceMethodsAroundInterceptor {
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        // if the removed handler is a enhanced handler by skywalking, we call the method again to remove and return the user handler
```
wu-sheng, line 39:
```suggestion
        // If the removed handler is enhanced by skywalking, we call the method again to remove and return the user handler.
```

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -25,6 +25,7 @@ metrics based on the tracing data.
   * [Jersey REST framework](https://github.com/eclipse-ee4j/jersey) 2.x -> 3.x
   * [Grizzly](https://github.com/eclipse-ee4j/grizzly) 2.3.x -> 4.x
   * [WebSphere Liberty](https://github.com/OpenLiberty/open-liberty) 23.x
+  * [Netty HTTP](https://github.com/netty/netty) 4.1.x
```
wu-sheng, line 28:
```suggestion
  * [Netty HTTP](https://github.com/netty/netty) 4.1.x (Optional¹)
```

I checked your test scenario, this plugin seems to work for both client and server of Netty? Am I right? If so, I feel it is better to put it inside `RPC` catalog.

wu-sheng: And `optional` suffix is required as it is not activated by default.

peachisai: > I checked your test scenario, this plugin seems to work for both client and server of Netty? Am I right? If so, I feel it is better to put it inside `RPC` catalog.
Yes ,it works on the both side.

peachisai: > I checked your test scenario, this plugin seems to work for both client and server of Netty? Am I right? If so, I feel it is better to put it inside `RPC` catalog.

Yes,it works on the both side

peachisai: > (Optional¹)

image
Is it ok?

wu-sheng: It is Optional 2. Optional 1 means having a license incompitable issue.

wu-sheng: The descriptions of Optional 1 and 2 are at the bottom of the doc file.

peachisai: > incompitable

I got it

peachisai:
@wu-sheng 
should i reduce the test version？

wu-sheng:
It seems to be fine. Let me check later.

CHANGES.md
```diff
@@ -16,7 +16,7 @@ Release Notes.
 * To compatible upper and lower case Oracle TNS url parse.
 * Fix config length limitation.
 * Support collecting ZGC memory pool metrics. Require OAP 9.7.0 to support these new metrics.
-
+* Add a netty-http 4.1.x plugin to trace http
```
wu-sheng, line 19:
```suggestion
* Add a netty-http 4.1.x plugin to trace HTTP requests
```

test/plugin/scenarios/netty-http-4.1.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,77 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+segmentItems:
+  - serviceName: netty-http-4.1.x-scenario
+    segmentSize: gt 0
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: HEAD:/case/healthCheck
```
wu-sheng, line 22: `/healtchCheck` is not needed in the expected data.

test/plugin/scenarios/netty-http-4.1.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,77 @@
...
+              - { key: url, value: https://127.0.0.1:8070/user }
+              - { key: http.body, value: "{\"name\":\"david\",\"age\":30}" }
+              - { key: http.status_code, value: '200' }
+            refs:
+              - { parentEndpoint: Netty-http:/user, networkAddress: 'localhost:8070',
+                  refType: CrossProcess, parentSpanId: 0, parentTraceSegmentId: not null, parentServiceInstance: not
+                                                                                            null, parentService: netty-http-4.1.x-scenario, traceId: not null }
+      - segmentId: not null
+        spans:
+          - operationName: Netty-http:/user
```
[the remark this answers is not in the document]
peachisai: > Is this a timer-driven?
> 
> Usually, plugin test should be driven by the test framework, URI=`/netty-http-4.1.x-case`. This URI<1> should lead to an entry span, then an exit span to call another URI<2> to test cross-process propagation. In the test URI <1> and <2> could be in one service, but let's make sure you are following the general workflow.

1.It is not a timer-driven
2.The workflow: GET:/netty-http-4.1.x-scenario/case/netty-http-4.1.x-case->Netty-http:/user exist span->Netty-http:/user entry span

Do I need to fix something?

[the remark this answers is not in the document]
wu-sheng: OK. Then it is fine for the workflow. Ignore my asking. Thanks.

test/plugin/scenarios/netty-http-4.1.x-scenario/support-version.list
peachisai: > It is fine. I am just wondering whether there are some unsupported cases. Could you submit more to verify and exclude most in the final commits?

Should I reduce these test versions?

wu-sheng: No, the current status is good

test/plugin/scenarios/netty-http-4.1.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,60 @@
...
+            parentSpanId: -1
+            spanId: 0
```
wu-sheng, line 46: I want to double check, the reason that this exit span doesn't have a parent entry span, is this RPC client called in async?

wu-sheng: According to https://github.com/apache/skywalking-java/pull/625#discussion_r1370236560, there is a parent entry span expected. If this is caused by async, you should consider connecting the parent entry and this exit span, through application toolkit lib or some mechanism you prefer.

wu-sheng:
@peachisai Any feedback?

peachisai:
> @peachisai Any feedback?

This week i am a little busy,i will provide feedback as soon as possible.

wu-sheng:
So, the previous codes miss the way connecting between the main thread and netty HTTP sender channel?

peachisai:
> @peachisai Any feedback?

done

peachisai:
> So, the previous codes miss the way connecting between the main thread and netty HTTP sender channel?

Yes

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -25,6 +25,7 @@ metrics based on the tracing data.
   * [Jersey REST framework](https://github.com/eclipse-ee4j/jersey) 2.x -> 3.x
   * [Grizzly](https://github.com/eclipse-ee4j/grizzly) 2.3.x -> 4.x
   * [WebSphere Liberty](https://github.com/OpenLiberty/open-liberty) 23.x
+  * [Netty HTTP](https://github.com/netty/netty) 4.1.x (Optional¹)
```
wu-sheng, line 28:
```suggestion
  * [Netty HTTP](https://github.com/netty/netty) 4.1.x (Optional²)
```

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -70,6 +71,7 @@ metrics based on the tracing data.
   * [Apache CXF](https://github.com/apache/cxf) 3.x
   * [JSONRPC4J](https://github.com/briandilley/jsonrpc4j) 1.2.0 -> 1.6
   * [Nacos-Client](https://github.com/alibaba/nacos) 2.x (Optional²)
+  * [Netty](https://github.com/netty/netty) 4.1.x (Optional²)
```
wu-sheng, line 74:
```suggestion
```

docs/en/setup/service-agent/java-agent/Optional-plugins.md
```diff
@@ -23,4 +23,5 @@ Now, we have the following known optional plugins.
...
 * Plugin of trace sampler CPU policy in the optional plugin folder. Please only activate this plugin when you need to disable trace collecting when the agent process CPU usage is too high(over threshold).
 * Plugin for Spring 6.x and RestTemplate 6.x are in the optional plugin folder. Spring 6 requires Java 17 but SkyWalking is still compatible with Java 8. So, we put it in the optional plugin folder.
-* Plugin of nacos-client 2.x lib in optional plugin folder.The reason is many business irrelevant traces are generated, which cause extra payload to agents and backends, also spend more CPU, memory and network.
\ No newline at end of file
+* Plugin of nacos-client 2.x lib in optional plugin folder.The reason is many business irrelevant traces are generated, which cause extra payload to agents and backends, also spend more CPU, memory and network.
+* Plugin of netty-http 4.1.x lib in optional plugin folder.The reason is some frameworks use netty as an HTTP server and also enhanced by other plugins of SkyWalking, which can result in double trace collection
```
wu-sheng, line 27:
```suggestion
* Plugin of nacos-client 2.x lib in optional plugin folder. The reason is many business irrelevant traces are generated, which cause extra payload to agents and backends, also spend more CPU, memory and network.
* Plugin of netty-http 4.1.x lib in optional plugin folder. The reason is some frameworks use Netty HTTP as kernel, which could double the unnecessary spans and create incorrect RPC relative metrics.
```

apm-sniffer/optional-plugins/netty-http-4.1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/netty/http/ChannelPipelineRemoveLastInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+public class ChannelPipelineRemoveLastInterceptor implements InstanceMethodsAroundInterceptor {
+
+    @Override
+    public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, MethodInterceptResult result) throws Throwable {
+
+    }
+
+    @Override
+    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        // If the removed handler is a enhanced handler by skywalking, we call the method again to remove and return the user handler.
```
wu-sheng, line 39:
```suggestion
        // If the removed handler is a SkyWalking enhanced handler, we continue removing the next which is the orignal handler before the instrumentation.
```

apm-sniffer/optional-plugins/netty-http-4.1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/netty/http/RemoveHandlerInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+            return ret;
+        }
+
+        Map<ChannelHandler, ChannelHandler> map = channel.attr(AttributeKeys.HANDLER_CLASS_MAP).get();
+
+        if (map == null) {
+            return ret;
+        }
+
+        // If the removed handler has an enhanced handler associated with it,the associated handler should also be removed
```
wu-sheng, line 58:
```suggestion
        // If the removed handler has an enhanced handler associated with it, the associated handler should also be removed
```

apm-sniffer/optional-plugins/netty-http-4.1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/netty/http/config/NettyHttpPluginConfig.java
```diff
@@ -0,0 +1,48 @@
...
+        public static class NettyHttp {
+
+            /**
+             * This config item controls that whether the Netty Http plugin should collect the http body of the request.
+             */
+            public static boolean COLLECT_REQUEST_BODY = false;
+
+            /**
+             * When either {@link Plugin.NettyHttp#COLLECT_REQUEST_BODY} is enabled, how many characters to keep and send to the OAP
+             * backend, use negative values to keep and send the complete body.
```
wu-sheng, line 37:
```suggestion
             * If {@link Plugin.NettyHttp#COLLECT_REQUEST_BODY} is enabled, 
             * this limitation threshold controls the max size of collected HTTP body.
             * Use the negative value to represent the unlimited threshold, which could cause OOM.
```

docs/en/setup/service-agent/java-agent/Supported-list.md
peachisai: > 

My negligence。According to https://github.com/apache/skywalking-java/pull/625#discussion_r1369424340，It is still in http server catalog better?

apm-sniffer/optional-plugins/netty-http-4.1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/netty/http/config/NettyHttpPluginConfig.java
```diff
@@ -0,0 +1,48 @@
...
+             */
+            public static int FILTER_LENGTH_LIMIT = 1024;
+
+            /**
+             * When either {@link Plugin.NettyHttp#COLLECT_REQUEST_BODY} is enabled and content-type start with SUPPORTED_CONTENT_TYPES_PREFIX, collect the body of the request
+             * use a comma to separate multiple types
```
wu-sheng, line 43:
```suggestion
             * If {@link Plugin.NettyHttp#COLLECT_REQUEST_BODY} is enabled, 
             * this config scopes the types of HTTP content for collecting HTTP request body.
             * Use a comma to separate multiple types.
```

wu-sheng (approved):

wu-sheng (approved):

PR #634, opened 2023-10-27
Optimizing the type pool of byte-buddy (#11460)

kylixs:
### Improve the performance of type pool of byte-buddy
Add type pool cache to byte-buddy
Problem discussion here: https://github.com/apache/skywalking/issues/11460

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #11460

### DIFF
--- a/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
+++ b/apm-sniffer/apm-agent/src/main/java/org/apache/skywalking/apm/agent/SkyWalkingAgent.java
@@ -23,6 +23,8 @@ import java.security.ProtectionDomain;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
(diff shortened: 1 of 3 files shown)


lujiajing1126:
Thanks for the patch. But actually it seems no change for the startup time.

wu-sheng:
I can see the diff is hybrid doesn't support loaded classes. I think that is for attach agent?

wu-sheng:
I verified here, https://github.com/apache/skywalking-java/pull/635, and [retransform-class-tomcat-scenario](https://github.com/apache/skywalking-java/actions/runs/6665572476/job/18115743332?pr=635#logs) failed 3 minutes ago in 1m 26s, but other cases are passed. 

@kylixs How about we add a configuration to disable this as retransfer is not that popular. And we could unify them if we could find a better solution.

And we could update the configuration doc and FAQ to highlight this.

wu-sheng:
I opened https://github.com/apache/skywalking-java/pull/636 to make this configurable. 

@kylixs If you have a solution for this, welcome to override my change.

PR #635, opened 2023-10-27
Change strategy from POOL_FIRST to HYBRID

wu-sheng:


wu-sheng:
@kylixs @lujiajing1126 I just changed the strategy and want to see what happens for the retransform case. @kylixs Do you expect this would fail?

PR #636, opened 2023-10-27
Expose  config to set up the re-transform support explicitly

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/11460

I am waiting for @lujiajing1126 final report about this case in their environment.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,9 @@ Release Notes.
 9.1.0
 ------------------
 
+* **IMPORTANT** Expose `SW_ENABLE_RETRANSFORM_SUPPORT` config to set up the re-transform support explicitly, 
+  and close it by default to keep startup delay low. 
+  This was added and enabled in 9.0.0, and we confirmed it would slow down the service launch.
 * Fix hbase onConstruct NPE in the file configuration scenario
 * Fix the issue of createSpan failure caused by invalid request URL in HttpClient 4.x/5.x plugin
 * Optimize ElasticSearch 6.x 7.x plugin compatibility
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
@@ -174,6 +174,17 @@ public class Config {
          * Enable the agent kernel services and instrumentation.
          */
         public static boolean ENABLE = true;
+
+        /**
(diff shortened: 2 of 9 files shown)


lujiajing1126:
For a real workload with ~30 classes to be enhanced. 

## Local env

- OS: macOS sonoma 14.0
- JDK: 1.8.0_372
- Hardware spec: Apple M2 with 16GB memory

| Variant | premain exec time* |
|----|----|
| POOL_FIRST  | ~6 secs |
| HYBRID |  |

* specifically `agentBuilder.installOn(instrumentation)`

wu-sheng:
@lujiajing1126 I updated the doc, please take a look.

lujiajing1126 (approved):
LGTM

kylixs:
This option might be better changed to enable or disable custom byte-buddy patch, for compatibility reasons.

wu-sheng:
Compatible with which part? 
I am waiting for your PR feedback. If it is good enough, we don't need this.

kylixs:
Determine whether to use a custom `AgentBuilder` with patch or a normal one: 
```java
    private static AgentBuilder newAgentBuilder(SWDescriptionStrategy descriptionStrategy) {
        final ByteBuddy byteBuddy = new ByteBuddy()
                .with(TypeValidation.of(Config.Agent.IS_OPEN_DEBUGGING_CLASS))
                .with(new SWAuxiliaryTypeNamingStrategy(NAME_TRAIT))
                .with(new SWImplementationContextFactory(NAME_TRAIT));

        return new SWAgentBuilderDefault(byteBuddy, new SWNativeMethodStrategy(NAME_TRAIT))
                .with(descriptionStrategy);
    }
```

wu-sheng:
I don't think we need to expose. This should not be aware ny users.
Anyone doesn't read codes, don't know what this is about.

kylixs:
I mean the option is `SW_ENABLE_RETRANSFORM_SUPPORT`,  disable byte-buddy patch if it is false. Users can understand this option.

wu-sheng:
I know your point. But I doubt they will know what path or our customization mean.

wu-sheng:
Like I documented, it should be a straight forward. If there are really some issues of your patch, that is a bug, nothing more

kylixs:
I agree. I'm just wondering if there's a fallback mechanism that can be enabled if something goes wrong.

wu-sheng:
Close for now, as we seems to have a solution now.

PR #637, opened 2023-10-28
Optimize bytebuddy type description performance

kylixs:
### Improve the performance of type description of byte-buddy
The goal is to get the original class description at re-transform, so as to generate consistent results when the Skywalking agent is enhanced again (including implementing the EnhancedInstance interface, auxiliary fields and methods, etc.)

The previous type description used the `AgentBuilder.DescriptionStrategy.Default.POOL_FIRST` policy to get origin type description, which slows down the application startup, due to heavy I/O operations and parsing bytecode. 

New way is to remove dynamic fields, methods and interfaces generated by SkyWalking Agent from `TypeDescription`, and **make it as origin type descripton**.

**Key feature** :  
* No need to cache `TypeDescription` objects, less memory used.
* It only applies to the re-transform class processing flow and does not affect the startup process.

**Process flow:**
1. Find `TypeDescription` from commonly used type cache, such as primitive class.
2. Delegate to `AgentBuilder.DescriptionStrategy.Default.HYBRID`
3. Wrap `TypeDescription` by `SWTypeDescriptionWrapper` , remove fields, methods, interface generated by SkyWalking. 

**Relative Issue:**  https://github.com/apache/skywalking/issues/11460

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #11460


apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+                                 @MaybeNull JavaModule module) {
+        // use cache first
+        TypePool.CacheProvider cacheProvider = getCacheProvider(classLoader);
+        TypePool.Resolution resolution = cacheProvider.find(name);
+        if (resolution != null && resolution.isResolved()) {
+            return resolution.resolve();
+        }
+        // do as AgentBuilder.DescriptionStrategy.Default.HYBRID
+        return delegate.apply(name, type, typePool, circularityLock, classLoader, module);
+    }
```
wu-sheng, line 58: What is the diff between this and pool first?

kylixs: `POOL_FIRST`  strategy call `typePool.describe(..)` first, simply:
* find from cache
* parse bytecode if not found

But SWDescriptionStrategy is: 
* find from cache
* do as `HYBRID` strategy, create from type class if not null,  then from type pool

```java
HYBRID(true) {
      /** {@inheritDoc} */
      public TypeDescription apply(String name,
                                   @MaybeNull Class<?> type,
                                   TypePool typePool,
                                   CircularityLock circularityLock,
                                   @MaybeNull ClassLoader classLoader,
                                   @MaybeNull JavaModule module) {
          return type == null
                  ? typePool.describe(name).resolve()
                  : TypeDescription.ForLoadedType.of(type);
      }
  },

POOL_FIRST(false) {
        /** {@inheritDoc} */
        public TypeDescription apply(String name,
                                     @MaybeNull Class<?> type,
                                     TypePool typePool,
                                     CircularityLock circularityLock,
                                     @MaybeNull ClassLoader classLoader,
                                     @MaybeNull JavaModule module) {
            TypePool.Resolution resolution = typePool.describe(name);
            return resolution.isResolved() || type == null
                    ? resolution.resolve()
                    : TypeDescription.ForLoadedType.of(type);
        }
    };

```

wu-sheng:
I talked with @lujiajing1126 He will check on his local env, and feedback.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
```diff
@@ -0,0 +1,64 @@
...
+                                 @MaybeNull ClassLoader classLoader,
+                                 @MaybeNull JavaModule module) {
+        // use cache first
+        TypePool.CacheProvider cacheProvider = getCacheProvider(classLoader);
+        TypePool.Resolution resolution = cacheProvider.find(name);
+        if (resolution != null && resolution.isResolved()) {
+            return resolution.resolve();
+        }
+        // do as AgentBuilder.DescriptionStrategy.Default.HYBRID
+        return delegate.apply(name, type, typePool, circularityLock, classLoader, module);
```
lujiajing1126, line 57: What about moving cache registration here?

lujiajing1126 (changes requested):
I've tested this patch. The performance is almost the same as SW 8.

lujiajing1126:
And what about raising a PR to the ByteBuddy to add so-called CACHE_FIRST policy?

wu-sheng:
We could try, from asking what could be side effect of this.
After all, bytebuddy is a general lib, more corner cases we may not know.

I am going to close mine.

@kylixs Could you polish this and update change log as well?

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
kylixs: Many non-enhanced classes also require type descriptions during the ByteBuddy processing. These classes do not actually require a cache and can save memory. You can compare the heap memory difference between the two caching methods, if the difference is not big, I think you can also put it here, the responsibility is clearer.

wu-sheng: I think to control memory cost is more important.

kylixs:
> And what about raising a PR to the ByteBuddy to add so-called CACHE_FIRST policy?

I think so. It is currently more of a cache delegation and should be modified to avoid relying on other strategies.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
kylixs: @lujiajing1126 Can you provide the memory analysis data of these two caching ways?

kylixs:
> @kylixs Could you polish this and update change log as well?

Not quite sure what to do, more?   Add `changes-9.1.0.md` and copy all changes from https://github.com/apache/skywalking-java/milestone/11?closed=1 ?

wu-sheng:
Update changes.md should be enough.

wu-sheng (approved):
LGTM. @lujiajing1126 Do you have more questions?

lujiajing1126 (approved):
LGTM now. I suppose we could further confirm this implementation with upstream.

kylixs:
@wu-sheng Some test cases failing,  may be not caused by this PR, trigger them again?

wu-sheng:
Done, you are a committer now, you could use details->rerun by yourself.

wu-sheng:
It seems there are some issues, they are failing again

wu-sheng:
Is this relative to the latest change? Or the previous commit is the same? Could you explain why OOM?

kylixs:
Too many `CacheProvider` and tokens: 
image

Heavily `TypeDescription`  ,  method token:

wu-sheng:
How is this possible to have so many CacheProvider? 8k instances?

wu-sheng:
According to the codes of `typePoolCache#locate`, too many instances of `CacheProvider` usually means too many class loaders? This test case doesn't seem so. Could you debug it to see why they are created in a so large number?

```java
                protected TypePool.CacheProvider locate(@MaybeNull ClassLoader classLoader) {
                    classLoader = classLoader == null ? getBootstrapMarkerLoader() : classLoader;
                    TypePool.CacheProvider cacheProvider = cacheProviders.get(classLoader);
                    while (cacheProvider == null) {
                        cacheProvider = TypePool.CacheProvider.Simple.withObjectType();
                        TypePool.CacheProvider previous = cacheProviders.putIfAbsent(classLoader, cacheProvider);
                        if (previous != null) {
                            cacheProvider = previous;
                        }
                    }
                    return cacheProvider;
                }
```

kylixs:
Too many classloader instances, every one has a `CacheProvider` : 
image

The top 2 cache providers occupy 224 MB of memory, this is the key point.
image

wu-sheng:
> Too many classloader instances, every one has a `CacheProvider` : image
> 
> The top 2 cache providers occupy 224 MB of memory, this is the key point. image

But, too many classloaders should not be our expected cases. And an async HTTP client should not do this either. That is the thing confusing me.

kylixs:
> But, too many classloaders should not be our expected cases. And an async HTTP client should not do this either. That is the thing confusing me.

Every `TypePool$Default$WithLazyResolution`  has a cache provider,  they weren't created by `SWDescriptionStrategy`!

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
lujiajing1126: > @lujiajing1126 Can you provide the memory analysis data of these two caching ways?

I will check this issue soon

kylixs:
@lujiajing1126 Please check the startup time of current version.

wu-sheng:
Could you provide docs to descibe your changes?

kylixs:
> Could you provide docs to descibe your changes?

The PR description has been updated with details.

wu-sheng:
> 3\. SWTypeDescriptionWrapper

@lujiajing1126 Please let us know whether this works.

wu-sheng:
The description is very clean. Thanks. Could you list the new generated class names, field names, and method names styles?

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
```diff
@@ -0,0 +1,335 @@
...
+            }
+        }
+        // wrap result
+        return new SWTypeDescriptionWrapper(delegate.apply(name, type, typePool, circularityLock, classLoader, module), nameTrait);
+    }
+
+    /**
+     * A TypeDescription wrapper to remove fields, methods, interface generated by SkyWalking.
+     */
+    static class SWTypeDescriptionWrapper extends TypeDescription.AbstractBase implements Serializable {
```
wu-sheng, line 118: You could build a separate default class, rather than `static inner class`.

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
```diff
@@ -0,0 +1,335 @@
...
+
+        /**
+         * The class's serial version UID.
+         */
+        private static final long serialVersionUID = 1L;
```
wu-sheng, line 123:
```suggestion
        /**
         * The class' serial version UID.
         */
        private static final long serialVersionUID = 1L;
```

Is the serialVersionUID required by some places?

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
```diff
@@ -0,0 +1,335 @@
...
+                } else {
+                    methods = declaredMethods;
+                }
+            }
+            return methods;
+        }
+
+        @Override
+        public boolean isAssignableTo(Class<?> type) {
+            // ignore interface added by SkyWalking
```
wu-sheng, line 199:
```suggestion
            // Ignore the interface added by SkyWalking
```

apm-sniffer/bytebuddy-patch/src/main/java/net/bytebuddy/agent/builder/SWDescriptionStrategy.java
```diff
@@ -0,0 +1,335 @@
...
+            if (IGNORED_INTERFACES.contains(type.getName())) {
+                return false;
+            }
+            return delegate.isAssignableTo(type);
+        }
+
+        @Override
+        public boolean isAccessibleTo(TypeDescription typeDescription) {
+            // ignore interface added by SkyWalking
```
wu-sheng, line 208:
```suggestion
            // Skip the Interface added by SkyWalking
```

lujiajing1126:
> > 3. SWTypeDescriptionWrapper
> 
> @lujiajing1126 Please let us know whether this works.

The startup time is almost the same as the latest 8.x version while the heap sizes are also similar.

Thanks for your work @kylixs

wu-sheng (approved):

PR #644, opened 2023-11-02
Fix Impala Jdbc URL (including schema without properties) parsing exception

CzyerChen:
### Fix Impala Jdbc URL (including schema without properties) parsing exception
- [ ] Explain briefly why the bug exists and how to fix it.
When set up Impala jdbc url like `jdbc:impala://localhost:21050/test` without properties behind, it may cause exception `java.lang.NumberFormatException: For input string: "21050/test"`. To avoid conversion exceptions, optimized the positioning of URL params.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/ImpalaJdbcURLParser.java
+++ b/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/ImpalaJdbcURLParser.java
@@ -64,7 +64,7 @@ public class ImpalaJdbcURLParser extends AbstractURLParser {
         if (databaseStartTag == -1 && firstParamIndex == -1) {
             return null;
         } else {
-            if(firstParamIndex == -1){
+            if (firstParamIndex == -1) {
                 firstParamIndex = url.length();
             }
             String subUrl = url.substring(startSize, firstParamIndex);
--- a/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/ImpalaJdbcURLParser.java
+++ b/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/ImpalaJdbcURLParser.java
@@ -35,7 +35,10 @@ public class ImpalaJdbcURLParser extends AbstractURLParser {
     @Override
     protected URLLocation fetchDatabaseHostsIndexRange() {
         int hostLabelStartIndex = url.indexOf("//");
-        int hostLabelEndIndex = url.indexOf("/", hostLabelStartIndex + 2);
+        int hostLabelEndIndex = url.length();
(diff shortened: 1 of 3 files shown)


apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/ImpalaJdbcURLParser.java
```diff
@@ -99,7 +102,7 @@ public ConnectionInfo parse() {
         } else {
             String[] hostAndPort = hostSegment[0].split(":");
             if (hostAndPort.length != 1) {
-                return new ConnectionInfo(component, DB_TYPE, hostAndPort[0], Integer.valueOf(hostAndPort[1]), fetchDatabaseNameFromURL(location
+                return new ConnectionInfo(component, DB_TYPE, hostAndPort[0], Integer.parseInt(hostAndPort[1]), fetchDatabaseNameFromURL(location
```
wu-sheng, line 105: Could you explain why change this method?

wu-sheng:
CI fails as you don't follow the code style

> Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.2.1:check (validate) on project apm-jdbc-commons: You have 2 Checkstyle violations. -> [Help 1]

The code style file is in the root, and you should run maven verify on your local side.

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/ImpalaJdbcURLParser.java
CzyerChen: The ConnectionInfo's param `port` need type int not Integer, whether `Integer.parseInt` is better?

CzyerChen:
> maven-checkstyle-plugin:3.2.1:check (validate)

Got it!

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/ImpalaJdbcURLParser.java
wu-sheng: My question is why do you use parseInt insteads of valueOf.

CzyerChen: I see the ConnectionInfo's port need type `int`, so `Integer.parseInt` is enough, and `Integer.valueOf`(Integer.valueOf(parseInt()) here return `Integer` may be unnecessary.

CzyerChen: Revert `Integer .valueOf` changes, and fix impala url parse exception only first.

wu-sheng (approved):

PR #648, opened 2023-11-14
Optimize to final type

ruqinhu:
Optimize to final type

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/LoadedLibraryCollector.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/jvm/LoadedLibraryCollector.java
@@ -41,12 +41,12 @@ public class LoadedLibraryCollector {
 
     private static final ILog LOGGER = LogManager.getLogger(LoadedLibraryCollector.class);
     private static final  String JAR_SEPARATOR = "!";
-    private static Set<ClassLoader> CURRENT_URL_CLASSLOADER_SET = new HashSet<>();
+    private static final Set<ClassLoader> CURRENT_URL_CLASSLOADER_SET = new HashSet<>();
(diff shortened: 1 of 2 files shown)


wu-sheng:
Why do you propose this change?

ruqinhu:
> Why do you propose this change?

Minimize mutability: When a variable is declared as final, its value cannot be changed once it has been initialized. This is useful for declaring constants or other values that should not be modified.

wu-sheng:
In theory, it is. But usually, the compiler and JIT would do that automatically, as they detected no one is going to change that, it is safe to cache during process.

And the codes you changed seem to be not called in high frequency.

ruqinhu:
> In theory, it is. But usually, the compiler and JIT would do that automatically, as they detected no one is going to change that, it is safe to cache during process.
> 
> And the codes you changed seem to be not called in high frequency.

Okay, then I'll turn it of

ruqinhu:
> In theory, it is. But usually, the compiler and JIT would do that automatically, as they detected no one is going to change that, it is safe to cache during process.
> 
> And the codes you changed seem to be not called in high frequency.

org.apache.skywalking.apm.agent.core.plugin.loader.InterceptorInstanceLoader#load seem to  called in high frequency.

Remove not used check, because Class.forName not find class, will throw ClassNotFoundException instead of null.

wu-sheng:
Why ClassNotFound is a high frequency thing?

ruqinhu:
> Why ClassNotFound is a high frequency thing?

It is unnecessary to judge whether it is empty. Will JIT also optimize it?

wu-sheng:
I am not following. Instrument process only happens on booting and/or class first loaded.Nothing more.
Do you have any evidence to prove the performance has been impacted?

ruqinhu:
> I am not following. Instrument process only happens on booting and/or class first loaded.Nothing more. Do you have any evidence to prove the performance has been impacted?

i see,

wu-sheng:
We are happy to accept change if this thing really impacts perf.

PR #649, opened 2023-11-16
Fix PostgreSQL Jdbc URL parsing exception

CzyerChen:
### Fix #11552

### Fix multi hosts JDBC url parsing exception 

Fix https://github.com/apache/skywalking/issues/11552

----

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #11552.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContextTest.java
+++ b/apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContextTest.java
@@ -28,6 +28,7 @@ import org.apache.skywalking.apm.agent.core.test.tools.SegmentStorage;
 import org.apache.skywalking.apm.agent.core.test.tools.SegmentStoragePoint;
 import org.apache.skywalking.apm.agent.core.test.tools.TracingSegmentRunner;
 import org.junit.AfterClass;
+import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
@@ -73,9 +74,8 @@ public class IgnoredTracerContextTest {
         ContextManager.createLocalSpan("/test4");
         ContextManager.stopSpan();
 
-        assertThat(storage.getIgnoredTracerContexts().size(), is(3));
-        assertThat(storage.getTraceSegments().size(), is(1));
-
+        Assert.assertTrue(storage.getIgnoredTracerContexts().size() < 4);
+        Assert.assertTrue(storage.getTraceSegments().size() > 0);
     }
 
     @Test
--- a/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/PostgreSQLURLParser.java
+++ b/apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/PostgreSQLURLParser.java
@@ -37,15 +37,29 @@ public class PostgreSQLURLParser extends AbstractURLParser {
     protected URLLocation fetchDatabaseHostsIndexRange() {
         int hostLabelStartIndex = url.indexOf("//");
         int hostLabelEndIndex = url.indexOf("/", hostLabelStartIndex + 2);
+        if (hostLabelEndIndex == -1) {
+            hostLabelEndIndex = url.length();
+        }
         return new URLLocation(hostLabelStartIndex + 2, hostLabelEndIndex);
     }
 
     @Override
     protected URLLocation fetchDatabaseNameIndexRange() {
-        int databaseStartTag = url.lastIndexOf("/");
-        int databaseEndTag = url.indexOf("?", databaseStartTag);
+        int databaseLabelStartIndex = url.indexOf("//");
+        int databaseStartTag = url.indexOf("/", databaseLabelStartIndex + 2);
+        int databaseEndTag = url.indexOf("?", databaseLabelStartIndex + 2);
+        if (databaseEndTag < databaseStartTag && databaseEndTag != -1) {
+            //database parse fail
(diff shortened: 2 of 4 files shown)


wu-sheng:
I am not sure why, but the UTs on the Windows keep failing.

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/PostgreSQLURLParser.java
```diff
@@ -37,15 +37,29 @@ public PostgreSQLURLParser(String url) {
...
+        int databaseStartTag = url.indexOf("/", databaseLabelStartIndex + 2);
+        int databaseEndTag = url.indexOf("?", databaseLabelStartIndex + 2);
+        if (databaseEndTag < databaseStartTag && databaseEndTag != -1) {
+            //database parse fail
+            return new URLLocation(0, 0);
+        }
+        if (databaseStartTag == -1) {
+            //database empty
+            return new URLLocation(0, 0);
+        }
```
wu-sheng, line 58: If we fail into this, what is the database name?

CzyerChen: It will call method `fetchDatabaseNameFromURL` in `AbstractURLParser` next, the database name will be ""
```java
   protected String fetchDatabaseNameFromURL() {
        URLLocation hostsLocation = fetchDatabaseNameIndexRange(); // -->  return new URLLocation(0, 0);
        return url.substring(hostsLocation.startIndex(), hostsLocation.endIndex()); // --> substring will return ""
    }
```
Test cases matched:
```java
    @Test
    public void testParsePostgresqlJDBCURLWithHostNoPortAndDb() {
        ConnectionInfo connectionInfo = new URLParser().parser("jdbc:postgresql://localhost");
        assertThat(connectionInfo.getDBType(), is("PostgreSQL"));
        assertThat(connectionInfo.getDatabaseName(), is(""));
        assertThat(connectionInfo.getDatabasePeer(), is("localhost:5432"));
    }

    @Test
    public void testParsePostgresqlJDBCURLEmpty() {
        ConnectionInfo connectionInfo = new URLParser().parser("jdbc:postgresql://");
        assertThat(connectionInfo.getDBType(), is("PostgreSQL"));
        assertThat(connectionInfo.getDatabaseName(), is(""));
        assertThat(connectionInfo.getDatabasePeer(), is(":5432"));
    }
```

The MysqlUrlParser may be better, but it rewrite `fetchDatabaseNameFromURL` itself that not controlled by  `AbstractURLParser`.
```java
  protected String fetchDatabaseNameFromURL(int startSize) {
        URLLocation hostsLocation = fetchDatabaseNameIndexRange(startSize);
        if (hostsLocation == null) {
            return "";
        }
        return url.substring(hostsLocation.startIndex(), hostsLocation.endIndex());
    }
```

wu-sheng: Are there `jdbc:postgresql://localhost` and `jdbc:postgresql://` legal connection URLs?

CzyerChen:
> I am not sure why, but the UTs on the Windows keep failing.

all the new UTs added in URLParserTest?

wu-sheng:
> > I am not sure why, but the UTs on the Windows keep failing.
> 
> all the new UTs added in URLParserTest?

Don't worry about this. They are passed now.

apm-sniffer/apm-sdk-plugin/jdbc-commons/src/main/java/org/apache/skywalking/apm/plugin/jdbc/connectionurl/parser/PostgreSQLURLParser.java
CzyerChen: I learn from PostgreSQL documentation here `https://www.postgresql.org/docs/16/libpq-connect.html`
The following examples illustrate valid URI syntax:
```
postgresql://
postgresql://localhost
postgresql://localhost:5433
postgresql://localhost/mydb
postgresql://user@localhost
postgresql://user:secret@localhost
postgresql://other@localhost/otherdb?connect_timeout=10&application_name=myapp
postgresql://host1:123,host2:456/somedb?target_session_attrs=any&application_name=myapp
```
and  I tested 6 of them.

wu-sheng: Several things I noticed from the doc

- `postgresql://`, we should check what is the default path, and the peer can't be null for analysis
- `postgresql:///mydb?host=localhost&port=5433`, host and port could be given by parameters. We should support and connect them to make the peer.
- Please recheck on IPv6 case `postgresql://[2001:db8::1234]/database`

apm-sniffer/apm-agent-core/src/test/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContextTest.java
```diff
@@ -73,9 +74,8 @@ public void ignoredTraceContextWithSampling() {
         ContextManager.createLocalSpan("/test4");
         ContextManager.stopSpan();
 
-        assertThat(storage.getIgnoredTracerContexts().size(), is(3));
-        assertThat(storage.getTraceSegments().size(), is(1));
-
+        Assert.assertTrue(storage.getIgnoredTracerContexts().size() < 4);
+        Assert.assertTrue(storage.getTraceSegments().size() > 0);
```
wu-sheng, line 78: The original tests are not stable. Try to fix through this.

wu-sheng (approved):

PR #650, opened 2023-11-28
Bump up grpc

kezhenxu94:

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/rocketMQ-5.x-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/rocketMQ-5.x-plugin/pom.xml
@@ -39,6 +39,12 @@
 			<artifactId>rocketmq-client</artifactId>
 			<version>${rocketmq-client.version}</version>
 			<scope>provided</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.skywalking</groupId>
+			<artifactId>apm-agent-core</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
 			<exclusions>
 				<exclusion>
 					<groupId>io.grpc</groupId>
--- a/apm-sniffer/apm-sdk-plugin/rocketMQ-5.x-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/rocketMQ-5.x-plugin/pom.xml
@@ -39,6 +39,12 @@
 			<artifactId>rocketmq-client</artifactId>
 			<version>${rocketmq-client.version}</version>
 			<scope>provided</scope>
+			<exclusions>
+				<exclusion>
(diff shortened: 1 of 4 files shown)


wu-sheng:
Could you recheck the CI?

kezhenxu94:
> Could you recheck the CI?

https://github.com/apache/skywalking-java/pull/650#issuecomment-1829231131

👆

wu-sheng:
What do you mean? How a plugin blocks a lib upgrade?

kezhenxu94:
rocketmq sets a hard requirement of gRPC 1.53.0, upgrading our core grpc to >1.53.0 breaks maven dependency resolution

apm-sniffer/apm-sdk-plugin/rocketMQ-5.x-plugin/pom.xml
```diff
@@ -39,6 +39,12 @@
 			<artifactId>rocketmq-client</artifactId>
 			<version>${rocketmq-client.version}</version>
 			<scope>provided</scope>
+			<exclusions>
+				<exclusion>
+					<groupId>io.grpc</groupId>
+					<artifactId>grpc-netty</artifactId>
+				</exclusion>
+			</exclusions>
```
wu-sheng, line 47: I think we should exclude grpc from agent-core, rather than from this side?

kezhenxu94: Why? GRPC is needed in agent side, how can we exclude it there?

wu-sheng: I think the agent plugin depends on the agent core lib, which has the gRPC dependency overriding the version.
We just need the RocketMQ plugin compiled. So, about exclude, I mean for this plugin, it could depend on agent-core with gRPC excluded. In the runtime, this plugin is using the user's gRPC, our gRPC and netty would have been shaded in another package already. So, we should be good by that.

apm-sniffer/apm-sdk-plugin/rocketMQ-5.x-plugin/pom.xml
```diff
@@ -40,6 +40,18 @@
...
+			<groupId>org.apache.skywalking</groupId>
+			<artifactId>apm-agent-core</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+			<exclusions>
+				<exclusion>
+					<groupId>io.grpc</groupId>
+					<artifactId>grpc-netty</artifactId>
+				</exclusion>
+			</exclusions>
```
wu-sheng, line 53: These are my new changes.

wu-sheng (approved):
Dep upgrade works.

PR #651, opened 2023-11-30
Optimize plugin selector logic.

weixiang1862:
### Improve the performance of PluginSelector

`PluginCfg` call `PluginSelector#select` every plugin and every line in skywalking-plugin.def in load function, I move it to `PluginCfg#getPluginClassList`, then it just need call once.

I tested in my local pc (8C 16G), it can save 200ms + boot time when I exclude 4 plugins by config `plugin.exclude_plugins`.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginCfg.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginCfg.java
@@ -52,14 +52,13 @@ public enum PluginCfg {
                     LOGGER.error(e, "Failed to format plugin({}) define.", pluginDefine);
                 }
             }
-            pluginClassList = pluginSelector.select(pluginClassList);
         } finally {
             input.close();
         }
     }
 
     public List<PluginDefine> getPluginClassList() {
-        return pluginClassList;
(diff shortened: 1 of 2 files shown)


wu-sheng:
Thanks, I will check the codes later.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginCfg.java
```diff
@@ -52,14 +52,13 @@ void load(InputStream input) throws IOException {
...
             }
-            pluginClassList = pluginSelector.select(pluginClassList);
         } finally {
             input.close();
         }
     }
 
     public List<PluginDefine> getPluginClassList() {
-        return pluginClassList;
+        return pluginSelector.select(pluginClassList);
```
wu-sheng, line 61:
```suggestion
        pluginClassList = pluginSelector.select(pluginClassList);
        return pluginClassList;
```

You should make sure the excluded ones could be GC, right?

weixiang1862: Thanks, a new migic I learned.

wu-sheng (approved):

PR #656, opened 2023-12-07
Support forkjoinpool plugin in JDK11

786991884:
fix https://github.com/apache/skywalking/issues/11633

### DIFF
--- a/.github/workflows/plugins-jdk11-test.3.yaml
+++ b/.github/workflows/plugins-jdk11-test.3.yaml
@@ -1,0 +1,130 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Test
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/plugins-*.yaml'
+      - 'apm-application-toolkit/**'
+      - 'apm-commons/**'
+      - 'apm-protocol/**'
+      - 'apm-sniffer/**'
+      - 'test/plugin/**'
+      - '**/pom.xml'
+      - '!**.md'
+
+concurrency:
+  group: plugins-3-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build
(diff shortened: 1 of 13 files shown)


wu-sheng:
You need to add the test scenario to run in JDK 11, otherwise, this is only a manual test

wu-sheng:
Also, you removed the pull request template, we need that, and you missed the update in the change log.

wu-sheng:
> You need to add the test scenario to run in JDK 11, otherwise, this is only a manual test

This is still missed.

wu-sheng:
Why do you open a new test3 with so many cases?

.github/workflows/plugins-jdk11-test.3.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+  test:
+    needs: [ build ]
+    name: ${{ matrix.case }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    strategy:
+      matrix:
+        case:
+          - jdk11-forkjoinpool-scenario
```
wu-sheng, line 55: Do we need a new case? I think this is rt.jar(jdk level) change, even the codes are compiled as 1.8, the runtime method is still going to be forced to run in 11, right?

wu-sheng: Could we not copy the test scenario, but directly write the existing test in jdk11. Was that failing before and running now?

apm-sniffer/bootstrap-plugins/jdk-forkjoinpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/jdk/forkjoinpool/define/ForkJoinWorkerQueueInstrumentation.java
```diff
@@ -33,6 +33,9 @@ public class ForkJoinWorkerQueueInstrumentation extends ClassInstanceMethodsEnha
     private static final String FORK_JOIN_WORKER_QUEUE_CLASS = "java.util.concurrent.ForkJoinPool$WorkQueue";
 
     private static final String FORK_JOIN_WORKER_QUEUE_RUN_TASK_METHOD = "runTask";
+    
+    private static final String FORK_JOIN_WORKER_QUEUE_RUN_TASK_METHOD_JDK11 = "topLevelExec";
```
wu-sheng, line 37: Please add comments about why there are two methods to intercept.

wu-sheng:
From [test logs](https://github.com/apache/skywalking-java/actions/runs/7162355124/job/19504471071?pr=656), your interceptor doesn't run as expected. Still, the old method existed and being intercepted.

wu-sheng:
Tests passed, could you check this? https://github.com/apache/skywalking-java/pull/656#discussion_r1422039367

.github/workflows/plugins-jdk11-test.3.yaml
786991884: yes, you are right. I will optimize it

wu-sheng:
Is there any update here?

wu-sheng:
@786991884 Do you have any update?

wu-sheng:
@786991884 If you will be back, I hope this extra test can be removed by reusing the existing one.

I am going to merge this first because from what I saw, this fix is important.

wu-sheng (approved):

PR #657, opened 2023-12-07
Add Pulsar configuration to determine the collection of message contents

CommissarXia:
### <Feature description>
- [x] Modify the unit test to verify that the feature works.
- [x] The configuration has been verified in practice.


apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/PulsarPluginConfig.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class PulsarPluginConfig {
+
+    public static class Plugin {
+        @PluginConfig(root = PulsarPluginConfig.class)
+        public static class Pulsar {
+            /**
+             * If set to true, the message contents of the Pulsar topic would be collected.
+             */
+            public static boolean TRACE_MESSAGE_CONTENTS = true;
```
wu-sheng, line 30: The default value should be false. The impact is clear for memory, network and database.

wu-sheng: agent.config set to false is not enough.

CommissarXia: Okay, I'll make some modifications and conduct unit testing. Thank you!

apm-sniffer/apm-sdk-plugin/pulsar-common/src/main/java/org/apache/skywalking/apm/plugin/pulsar/common/PulsarConsumerInterceptor.java
```diff
@@ -73,6 +75,9 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
             Tags.MQ_BROKER.set(activeSpan, serviceUrl);
             Tags.MQ_TOPIC.set(activeSpan, consumer.getTopic());
             activeSpan.setPeer(serviceUrl);
+            if (Pulsar.TRACE_MESSAGE_CONTENTS) {
+                Tags.MQ_BODY.set(activeSpan, StringUtil.cut(new String(msg.getData()), Pulsar.MESSAGE_CONTENTS_MAX_LENGTH));
```
wu-sheng, line 79: `msg.getData()` is a byte[]. Is here OK to use the default charset?

CommissarXia: We used the default encoding for a long time without any issues, but we are using Java and are unsure if other languages can execute properly. This time, I added the UTF-8 character set and it ran normally in our environment

wu-sheng:
Please update this test, https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/pulsar-scenario

Plugin test could run locally, https://skywalking.apache.org/docs/skywalking-java/v9.1.0/en/setup/service-agent/java-agent/plugin-test/

You should activate this new config and verify new tags.

CommissarXia:
There are some issues with the environment of my computer. I will submit it after I run it. Thanks!

test/plugin/scenarios/pulsar-scenario/config/expectedData.yaml
```diff
@@ -97,6 +99,7 @@ segmentItems:
             - { key: transmission.latency, value: not blank }
             - { key: mq.broker, value: not blank }
             - { key: mq.topic, value: test }
+            - { key: mq.body, value: "1" }
```
wu-sheng, line 102: This is how the message is built, 
```java
producer.newMessage().key("testKey").value(Integer.toString(1).getBytes()).property("TEST", "TEST").send();
```

But you only set `1` as body? Could you explain?

CommissarXia: This 1 was set by the previous submitter, and I only tested the corresponding expected value based on this value. If need test message bodies in other formats, I can modify and resubmit them.

wu-sheng: But there is not only value. There are a key and properties. aren't those message body too?

wu-sheng:
Could you explain what you changed? 

Personally, I am frustrated about the communications between us. I have to read codes, understand codes, but you never relies words.

CommissarXia:
Actually, I only added a dynamic configuration item to dynamically collect the message contents of Pulsar. I changed "mq.body" to "mq.message_contents" and have already submitted it. This can more clearly reflect the content I added and the expected results.

wu-sheng:
My question is key/value, why is the key not collected?

CommissarXia:
The key is Pulsar's strategy for message routing, and its relationship with actual messages can be inclusive or independent. I just wanted to collect actual messages, so I didn't consider this key. Perhaps my consideration was a bit lacking.

wu-sheng:
Is there pulsar doc indicating key is not a part of message? I am doubt about that.

wu-sheng:
@tisonkun What is the message concept in the Pulsar?

CommissarXia:
Misunderstand my meaning, key must be passed together with the content of the message body, but the collection content I want to add does not take into account key, I will add it sometime today.

CommissarXia:
The messages in Pulsar include Value/data payload, Key, Properties, etc. My initial idea was just to collect Values.

wu-sheng:
That doesn't make sense from my understanding, unless this is from Pulsar officially.
If we declare collecting message, it should be the whole.

CommissarXia:
OK, it should be my lack of understanding. I'll readjust it again. I'm really sorry for wasting your time.

wu-sheng:
For next time, please write the proposal that you design first. It would help me align with you.

CommissarXia:
Ok, thank you very much!

wu-sheng:
Any update about this/

CommissarXia:
I was sick last week. I have time to deal with it this week.

wu-sheng:
Is this still on going?

CommissarXia:
Shengge, After careful consideration later, maybe this feature is not very necessary, and the demand for simply collecting message content is probably not too much, so I want to give up this pull request. 
Shengge, do you have any opinions? Sorry for wasting your time.

wu-sheng:
Closing upon request.

PR #658, opened 2023-12-14
Fix NoSuchMethodError in mvc-annotation-commons&change deprecated method

CzyerChen:
### Fix [NoSuchMethodError in mvc-annotation-commons&change deprecated method](https://github.com/apache/skywalking/issues/11652)
current test scenarios(spring-4.3.x-scenario, spring-6.x-scenario) passed.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/11652.

### DIFF
--- a/test/plugin/scenarios/spring-6.x-scenario/support-version.list
+++ b/test/plugin/scenarios/spring-6.x-scenario/support-version.list
@@ -15,3 +15,5 @@
 # limitations under the License.
 
 6.0.4
+6.1.0
+6.1.1
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@ Release Notes.
 9.2.0
 ------------------
 
+* Fix NoSuchMethodError in mvc-annotation-commons and change deprecated method 
+
 
 #### Documentation
 
--- a/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java
@@ -168,11 +168,11 @@ public abstract class AbstractMethodInterceptor implements InstanceMethodsAround
                         next.setHeadValue(serverHttpRequest.getHeaders().getFirst(next.getHeadKey()));
                     }
 
-                    String operationName = this.buildOperationName(method, serverHttpRequest.getMethodValue(),
+                    String operationName = this.buildOperationName(method, serverHttpRequest.getMethod().name(),
                                                                    (EnhanceRequireObjectCache) objInst.getSkyWalkingDynamicField());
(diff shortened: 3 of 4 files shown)


wu-sheng:
I want to double-check, [this test](https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/spring-3.1.x-scenario) has covered 3.2.18.RELEASE, so, only 3.2.0 had the bug, but 3.2.8 is OK?

CzyerChen:
> I want to double-check, [this test](https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/spring-3.1.x-scenario) has covered 3.2.18.RELEASE, so, only 3.2.0 had the bug, but 3.2.8 is OK?

I can write a scenario with version `3.2.18.RELEASE` and test again locally.

wu-sheng:
The link I provided, it has tested 3.1.4.RELEASE, 3.2.18.RELEASE, 4.0.9.RELEASE, check this, https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/spring-3.1.x-scenario/support-version.list.

What I want to confirm with you is, is this test covers your scenario if we add 3.2.0.xx into that? Or 3.2.18 is still having this bug just not covered by the test.

CzyerChen:
> The link I provided, it has tested 3.1.4.RELEASE, 3.2.18.RELEASE, 4.0.9.RELEASE, check this, https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/spring-3.1.x-scenario/support-version.list.
> 
> What I want to confirm with you is, is this test covers your scenario if we add 3.2.0.xx into that? Or 3.2.18 is still having this bug just not covered by the test.

- SpringBoot version upgrade to 3.2.0 and Spring version upgrade to 6.1.1, then `getMethodValue` removed and caused `NoSuchMethodError`, so i want to commit this change.
- I add Spring version 6.1.0/6.1.1 support-version into `spring-6.x-scenario`, change dependency versions in pom.xml to support Spring 6.x([springboot2+spring5 dependency management for 2.7.x with springboot3+spring6 dependency management for 3.0.x ](https://docs.spring.io/spring-boot/docs/3.0.x/reference/html/dependency-versions.html#appendix.dependency-versions).
- Upgrade from springboot3.1.x+spring6.0.x to springboot3.2.x+spring6.1.x, our testcase `spring-6.x-scenario` is a spring project, not a springboot project, so i follow the issue [http request parameter error](https://github.com/spring-projects/spring-boot/issues/38541#issuecomment-1827462871) to change maven compile config, and it works. 
- I passed `spring-6.x-scenario` locally.

test/plugin/scenarios/spring-6.x-scenario/support-version.list
```diff
@@ -15,3 +15,5 @@
 # limitations under the License.
 
 6.0.4
+6.1.0
+6.1.1
```
wu-sheng, line 19: So, only this is using SpringBoot 3.2.0, which was failing before the fix, am I getting your idea correctly?

CzyerChen: yes, I tracd my own springboot project in 3.2.0(Spring 6.1.1), and find SpringMVC node missing, but springboot project in 3.1.6(Spring 6.0.14) works.

wu-sheng (approved):

PR #660, opened 2023-12-18
Support for tracing Spring-Webflux request in Webflux 6.x plugin

CzyerChen:
### Fix #11654 and add support for spring-webflux-6.x

### Add an agent plugin to support Spring-Webflux-6.x

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #11654.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,8 +5,8 @@ Release Notes.
 9.2.0
 ------------------
 
-* Fix NoSuchMethodError in mvc-annotation-commons and change deprecated method 
-
+* Fix NoSuchMethodError in mvc-annotation-commons and change deprecated method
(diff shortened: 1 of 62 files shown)


wu-sheng:
What happened here? 

A tip, if you are adding new toolkit, please
1. Make sure it is necessary, such as WebFlux 5.x and 6.x expose different APIs for toolkit compilings.
2. Docs should be updated accordingly.

CzyerChen:
> What happened here?
> 
> A tip, if you are adding new toolkit, please
> 
> 1. Make sure it is necessary, such as WebFlux 5.x and 6.x expose different APIs for toolkit compilings.
> 2. Docs should be updated accordingly.

I can think about this again, and redo the pull request.

PR #661, opened 2023-12-19
Support for tracing webflux-6.x and gateway-4.x

CzyerChen:
### Add support for spring-cloud-gateway 4.x 

why:
spring-cloud-starter-gateway 4.0.0 depends on  spring-boot-starter-webflux 3.0.0
spring-boot-starter-webflux 3.0.0  depends on spring-webflux 6.0.2
spring-webflux-5.x plugin can not trace the request
how:
add support for spring-webflux-6.x diff from 5.x by witnessmethod
upgrade spring-cloud-gateway to 4.x with spring-webflux-6.x diff from 3.x by witnessclass

### Add an agent plugin to support spring-cloud-gateway 4.x 

### DIFF
--- a/docs/en/setup/service-agent/java-agent/Supported-list.md
+++ b/docs/en/setup/service-agent/java-agent/Supported-list.md
@@ -15,7 +15,7 @@ metrics based on the tracing data.
   * Resin 3 (Optional¹)
   * Resin 4 (Optional¹)
   * [Jetty Server](http://www.eclipse.org/jetty/) 9.x -> 11.x
-  * [Spring WebFlux](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) 5.x (Optional¹)
+  * [Spring WebFlux](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) 5.x (Optional¹) -> 6.x
   * [Undertow](http://undertow.io/)  1.3.0.Final -> 2.0.27.Final
   * [RESTEasy](https://resteasy.github.io/)  3.1.0.Final -> 6.2.4.Final
   * [Play Framework](https://www.playframework.com/) 2.6.x -> 2.8.x
--- a/docs/en/setup/service-agent/java-agent/Supported-list.md
+++ b/docs/en/setup/service-agent/java-agent/Supported-list.md
@@ -15,7 +15,7 @@ metrics based on the tracing data.
   * Resin 3 (Optional¹)
   * Resin 4 (Optional¹)
   * [Jetty Server](http://www.eclipse.org/jetty/) 9.x -> 11.x
-  * [Spring WebFlux](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) 5.x (Optional¹) -> 6.x
+  * [Spring WebFlux](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) 5.x (Optional¹) -> 6.x (Optional¹)
   * [Undertow](http://undertow.io/)  1.3.0.Final -> 2.0.27.Final
   * [RESTEasy](https://resteasy.github.io/)  3.1.0.Final -> 6.2.4.Final
   * [Play Framework](https://www.playframework.com/) 2.6.x -> 2.8.x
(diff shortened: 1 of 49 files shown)


test/plugin/scenarios/gateway-4.x-scenario/gateway-projectA-scenario/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+    <pluginRepositories>
+        <pluginRepository>
+            <id>spring-snapshots</id>
+            <url>https://repo.spring.io/snapshot</url>
+        </pluginRepository>
+        <pluginRepository>
+            <id>spring-milestones</id>
+            <url>https://repo.spring.io/milestone</url>
+        </pluginRepository>
+    </pluginRepositories>
```
wu-sheng, line 66: I think we don't need this. Maven central should all versions?

test/plugin/scenarios/gateway-4.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,155 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+segmentItems:
+- serviceName: gateway-projectB-scenario
+  segmentSize: nq 0
+  segments:
+  - segmentId: not null
+    spans:
+    - operationName: /provider/timeout/error
```
wu-sheng, line 22: Is this timeout built intentionally?

test/plugin/scenarios/gateway-4.x-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - {parentEndpoint: SpringCloudGateway/RoutingFilter, networkAddress: 'localhost:18070',
+        refType: CrossProcess, parentSpanId: 1, parentTraceSegmentId: not null, parentServiceInstance: not
+          null, parentService: not null, traceId: not null}
+      skipAnalysis: 'false'
+- serviceName: gateway-projectA-scenario
+  segmentSize: nq 0
+  segments:
+  - segmentId: not null
+    spans:
+    - operationName: /provider/timeout/error
```
wu-sheng, line 66: Why is a timeout for entry? How an entry span timeout?

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -15,7 +15,7 @@ metrics based on the tracing data.
   * Resin 3 (Optional¹)
   * Resin 4 (Optional¹)
   * [Jetty Server](http://www.eclipse.org/jetty/) 9.x -> 11.x
-  * [Spring WebFlux](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) 5.x (Optional¹)
+  * [Spring WebFlux](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) 5.x (Optional¹) -> 6.x
```
wu-sheng, line 18:
```suggestion
  * [Spring WebFlux](https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html) 5.x (Optional¹) -> 6.x (Optional¹)
```

wu-sheng: I think `[Spring Cloud Gateway](https://spring.io/projects/spring-cloud-gateway) 2.0.2.RELEASE -> 3.x (Optional²)` is not updated.

CzyerChen: got it.

test/plugin/scenarios/gateway-4.x-scenario/config/expectedData.yaml
CzyerChen: in order to check error case.
image

test/plugin/scenarios/gateway-4.x-scenario/gateway-projectA-scenario/pom.xml
CzyerChen: got it.

test/plugin/scenarios/gateway-4.x-scenario/config/expectedData.yaml
wu-sheng: My question is, are the codes causing the timeout on testing purpose?

wu-sheng:
Could you confirm this https://github.com/apache/skywalking-java/pull/661#discussion_r1433534466? If the codes don't seem to trigger error, we should not have them in the expected data.

test/plugin/scenarios/gateway-4.x-scenario/config/expectedData.yaml
CzyerChen: according to gateway3.x, in order to check request timeout case, so add the testcasse, [issue](https://github.com/apache/skywalking-java/pull/175)

wu-sheng (approved):
Thanks for adding this.

PR #664, opened 2024-01-11
add support for HttpExchange along with webflux-webclient-6.x

CzyerChen:
### Add an agent plugin to support <SpringBoot-HttpExchange>
- [ ] Add a component id in [the component-libraries.yml](https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/component-libraries.yml), component name exist

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #11715.

### DIFF
--- a/docs/en/setup/service-agent/java-agent/Plugin-list.md
+++ b/docs/en/setup/service-agent/java-agent/Plugin-list.md
@@ -174,3 +174,4 @@
 - websphere-liberty-23.x
 - spring-cloud-gateway-4.x
 - spring-webflux-6.x
+- spring-webflux-6.x-webclient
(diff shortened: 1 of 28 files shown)


docs/en/setup/service-agent/java-agent/Plugin-list.md
```diff
@@ -174,3 +174,4 @@
 - websphere-liberty-23.x
 - spring-cloud-gateway-4.x
 - spring-webflux-6.x
+- spring-webflux-6.x-webclient
```
wu-sheng, line 177: I think both for webclient 5.x and 6.x should be updated in https://skywalking.apache.org/docs/skywalking-java/next/en/setup/service-agent/java-agent/supported-list/

CzyerChen: add `  * [Spring Webflux WebClient](https://github.com/spring-projects/spring-framework/tree/main/spring-webflux) 5.x -> 6.x` in supported-list

wu-sheng (approved):
Thanks for adding this.

PR #665, opened 2024-01-19
Support tracing for async producing, batch sync consuming, and batch async consuming in rocketMQ-client-java-5.x-plugin

CzyerChen:
### Enhance an agent plugin to support RocketMQ

Based on `rocketMQ-client-java-5.x-plugin`, add interceptors to tracing async producing, batch sync consuming, and batch async consuming  cases. Batch consuming messages can refer to the producer. Async producing can record messageId.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/11752.


wu-sheng:
Could you share how batch consuming works? Is UI working as expected to see multiple relative trace ID from downstream?

CzyerChen:
> Could you share how batch consuming works? Is UI working as expected to see multiple relative trace ID from downstream?

As the testcase:
1. api
2. produce 1  normal message
3. PushConsumer consumes 1 message by listener
4.5. produce 2 normal messages async
6. SimpleConsumer consumes 2 messages by triggered manually
7.8. produce 2 normal messages
9. SimpleConsumer consumes 2 messages async by triggered manually

image

wu-sheng:
So, from reading your graph, the same number of spans are going to be created based on the number of messages in one bulk. Am I right?

Q, what will happen if two messages are produced from different contexts (two trace IDs), and consumed in one bulk from one client, is this possible?

wu-sheng:
BTW, at here, you need to use the whole URL, rather than issue ID.

```
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/11752.
```

CzyerChen:
> So, from reading your graph, the same number of spans are going to be created based on the number of messages in one bulk. Am I right?
> 
> Q, what will happen if two messages are produced from different contexts (two trace IDs), and consumed in one bulk from one client, is this possible?

Testcase:
ProjectA:
1.  produce 1 message 
2. produce 1 message

ProjectB:
1. consumer 2 message in a batch

Then, the graph like this:
image
image


1 producer can be referred,  the other can not.

wu-sheng:
What does the VNode mean in this case?

CzyerChen:
> What does the VNode mean in this case?

image
 The VNode is another ref.

wu-sheng:
OK, I will work with @Fine0830 about UI part, this should not be your concern. Thanks for your explanation.

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageListenerInterceptor.java
```diff
@@ -37,30 +38,38 @@
...
         MessageView messageView = (MessageView) allArguments[0];
 
         ContextCarrier contextCarrier = getContextCarrierFromMessage(messageView);
 
         AbstractSpan span = ContextManager.createEntrySpan(CONSUMER_OPERATION_NAME_PREFIX + messageView.getTopic()
-                + "/Consumer", contextCarrier);
+                                                               + "/Consumer", contextCarrier);
         Tags.MQ_TOPIC.set(span, messageView.getTopic());
-
+        span.tag(Tags.ofKey("mq.message.id"), messageView.getMessageId().toString());
```
wu-sheng, line 56: `Tags#ofKey` is very expensive. The tag should be a static field, rather than created repeatedly.

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageListenerInterceptor.java
```diff
@@ -72,7 +81,11 @@ public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allA
     }
 
     @Override
-    public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Throwable t) {
+    public void handleMethodException(EnhancedInstance objInst,
+                                      Method method,
+                                      Object[] allArguments,
+                                      Class<?>[] argumentsTypes,
+                                      Throwable t) {
```
wu-sheng, line 88: Please don't reformat codes if you don't change relative things.

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageSendAsyncInterceptor.java
```diff
@@ -0,0 +1,148 @@
...
+        ContextCarrier contextCarrier = new ContextCarrier();
+        String namingServiceAddress = producerImpl.getClientConfiguration().getEndpoints();
+        AbstractSpan span = ContextManager.createExitSpan(
+            buildOperationName(message.getTopic()), contextCarrier, namingServiceAddress);
+        span.setComponent(ComponentsDefine.ROCKET_MQ_PRODUCER);
+        Tags.MQ_BROKER.set(span, namingServiceAddress);
+        Tags.MQ_TOPIC.set(span, message.getTopic());
+        Collection<String> keys = message.getKeys();
+        if (!CollectionUtil.isEmpty(keys)) {
+            span.tag(Tags.ofKey("mq.message.keys"), String.join(",", keys));
```
wu-sheng, line 72: Same suggestion for `Tags#ofKey`.

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageSendAsyncInterceptor.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+        Optional<String> tag = message.getTag();
+        tag.ifPresent(s -> span.tag(Tags.ofKey("mq.message.tags"), s));
```
wu-sheng, line 75: What is the typical use of tags? Would those be useful somehow?

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/SimpleConsumerImplAsyncInterceptor.java
```diff
@@ -0,0 +1,125 @@
...
+        final ContextSnapshot finalCapture = capture;
+        return futureList.whenCompleteAsync((messages, throwable) -> {
+            String topics = messages.stream().map(MessageView::getTopic).distinct().collect(Collectors.joining(","));
+            AbstractSpan span = ContextManager.createEntrySpan(
+                CONSUMER_OPERATION_NAME_PREFIX + topics + "/Consumer", null);
+            if (finalCapture != null) {
+                ContextManager.continued(finalCapture);
+            }
+            if (null != throwable) {
+                span.errorOccurred();
```
wu-sheng, line 72: We flag span as error occurred, but you don't log the error(throwable), could you explain why?

CzyerChen: have to log the error, I missed that.

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageSendAsyncInterceptor.java
CzyerChen: In general, different topics are used to separate different services, and different tags are used to separate different service segments. 

usages may be:
- 1 topic
- 1 topic and several tags
- several topics and  several tags 

It may be easily to know the service segments by recording and showing the tag of message.

wu-sheng: Could you make that optional? I am not sure whether this is an overload to tag everything. Afterall, if user needs details to debug, message body is required as well. But we can't afford to tag/log all.

CzyerChen: added & resubmitted

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageSendAsyncInterceptor.java
CzyerChen: updated & resubmitted

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/MessageListenerInterceptor.java
CzyerChen: revert & resubmitted

apm-sniffer/config/agent.config
```diff
@@ -316,3 +316,7 @@ plugin.nettyhttp.collect_request_body=${SW_PLUGIN_NETTYHTTP_COLLECT_REQUEST_BODY
 plugin.nettyhttp.filter_length_limit=${SW_PLUGIN_NETTYHTTP_FILTER_LENGTH_LIMIT:1024}
 #  When `HTTP_COLLECT_REQUEST_BODY` is enabled and content-type start with HTTP_SUPPORTED_CONTENT_TYPES_PREFIX, collect the body of the request , multiple paths should be separated by `,`
 plugin.nettyhttp.supported_content_types_prefix=${SW_PLUGIN_NETTYHTTP_SUPPORTED_CONTENT_TYPES_PREFIX:application/json,text/}
+# If set to true, the keys of RocketMQ mesaages would be collected by RocketMQ client java agent.
```
wu-sheng, line 319:
```suggestion
# If set to true, the keys of messages would be collected by the plugin for RocketMQ client Java agent.
```

apm-sniffer/config/agent.config
```diff
@@ -316,3 +316,7 @@ plugin.nettyhttp.collect_request_body=${SW_PLUGIN_NETTYHTTP_COLLECT_REQUEST_BODY
...
+plugin.rocketmqclient.collect_message_keys=${SW_PLUGIN_ROCKETMQCLIENT_COLLECT_MESSAGE_KEYS:false}
+# If set to true, the tags of RocketMQ mesaages would be collected by RocketMQ client java agent.
```
wu-sheng, line 321:
```suggestion
# If set to true, the tags of messages would be collected by the plugin for RocketMQ Java client.
```

apm-sniffer/apm-sdk-plugin/rocketMQ-client-java-5.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/rocketMQ/client/java/v5/RocketMqClientJavaPluginConfig.java
```diff
@@ -0,0 +1,37 @@
...
+        public static class Rocketmqclient {
+            /**
+             * This config item controls that whether the RocketMqClientJava plugin should collect the keys of the message.
+             */
+            public static boolean COLLECT_MESSAGE_KEYS = false;
+            /**
+             * This config item controls that whether the RocketMqClientJava plugin should collect the tags of the message.
+             */
+            public static boolean COLLECT_MESSAGE_TAGS = false;
+        }
```
wu-sheng, line 35: You should activate these two options in the test scenarios, to make sure they could work.

CzyerChen: ok

wu-sheng (approved):

PR #667, opened 2024-01-20
Convert the Redisson lock span into an async span

peachisai:
Refer https://github.com/apache/skywalking/discussions/11799


apm-sniffer/apm-sdk-plugin/redisson-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/redisson/v3/RedissonHighLevelLockInterceptor.java
```diff
@@ -48,7 +50,19 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
 
     @Override
     public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
+        AbstractSpan span = ContextManager.activeSpan();
+        span.prepareForAsync();
         ContextManager.stopSpan();
+
+        RFuture<Object> future = (RFuture) ret;
+        CompletableFuture<Object> completableFuture = future.toCompletableFuture();
+        completableFuture.whenCompleteAsync((res, ex) -> {
```
wu-sheng, line 59: AFAIK, as you picked `#whenCompleteAsync`, you can't guarantee the given execution(`span#asyncFinish`) called in time. The execution time could be longer than expected.

peachisai: > AFAIK, as you picked `#whenCompleteAsync`, you can't guarantee the given execution(`span#asyncFinish`) called in time. The execution time could be longer than expected.

How about this
```
        RFuture<Object> future = (RFuture) ret;
        CompletableFuture<Object> completableFuture = future.toCompletableFuture();

        CompletableFuture.runAsync(()->{
            completableFuture.join();
            span.asyncFinish();
        });
```

wu-sheng: Isn't this simpler?
```java
completableFuture.whenComplete((result, error) -> { span.asyncFinish(); });
```

peachisai: > Isn't this simpler?
> 
> ```java
> completableFuture.whenComplete((result, error) -> { span.asyncFinish(); });
> ```

Okay, I thought you mean the ```whenComplete``` may not notify listener as soon, consequently, i choose ```join``` method。so the question is the thread context change could cause another cost time

wu-sheng: `whenComplete` should be a sync call when the future completed. Isn't it? Am I getting anything wrong?

peachisai: > `whenComplete`

Yes,you are right.This is a sync call
)1CCYEXAHY}~_RU2~M4%YO5
As ```Redisson```  also use ```whenComplete ```  frequently, i think it is ok

wu-sheng (approved):

PR #670, opened 2024-01-26
Support for ActiveMQ-Artemis messaging tracing.

CzyerChen:
### Add an agent plugin to support ActiveMQ-Artemis

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes [#11830](https://github.com/apache/skywalking/issues/11830).


test/plugin/scenarios/activemq-artemis-2.x-scenario/support-version.list
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# lists your version here (Contains only the last version number of each minor version.)
+2.30.0
+2.31.2
```
wu-sheng, line 19: Could we support more versions? I saw versions in the docs are 2.30 and 2.31 only

CzyerChen: for the testcase depends on `spring-boot-3.2.0` and `artemis-jakarta-client` , it can support 
- not 2.19.0, and 2.19.0before not tested
- not 2.24.0
- not 2.25.0
- 2.26.0
- 2.27.0
- 2.28.0
- not 2.29.0
- 2.30.0
- 2.31.x
----
the version not supprt because of the dependency conflicts with `SpringBoot`, and `springboot-stater-artemis` officially support 3 versions of Artemis with jakarta-client (2.26.0,2.28.0,2.31.2)

wu-sheng: From you description, I think we should say we support 
- Spring Starter Artemis + ActiveMQ-Artemis a/b/c versions
- ActiveMQ-Artemis a/b/c/e/f/g versions.

We should have two test scenarios to cover two version lists.

wu-sheng:
```
-activemq-artemis-jakata-client-2.x
+activemq-artemis-jakarta-client-2.x
```

Please notice the name.

test/plugin/scenarios/activemq-artemis-2.x-scenario/support-version.list
CzyerChen: `artemis-jakarta-client` has many minor versions, do the support-list.md have to record all the last version number of each minor version.

image

wu-sheng: `support-list.md` should list all the range. `support-version.list` should pick the latest patch of every minor for verification. Such as, we could say, we support 2.19 - 2.31, and test 2.19.1 for 2.19.x

CzyerChen: It can support all `artemis-jakarta-client` versions from 2.17.0 to 2.31.2 by running the testcase project, but the official docker image [`apache/activemq-artemis` just provide from 2.30.0,](https://hub.docker.com/r/apache/activemq-artemis/tags). Can I written `support 2.17.0 ~ 2.31.2, and test 2.30.0/2.31.2` in `support-list.md`?

wu-sheng: Yes, you can put it that way if you have verified locally or have checked from source codes level.

```suggestion
# You should put comments here about why others are not tested in the plugin test pipeline.
2.30.0
2.31.2
```

wu-sheng: One more question, [Images](https://hub.docker.com/r/apache/activemq-artemis/tags) you provided are server side versions, could you use various client versions to connect 2.31 server?
What we need to verify is only client version compatibility, I believe the server version doesn't matter.

CzyerChen: Ok, I can fix the server version, to check the client version.

wu-sheng:
Please fix the javadoc. Warning is fine, notice the error ones, which should be from your codes.

wu-sheng:
Is anything wrong for RocketMQ grpc case?

CzyerChen:
> Is anything wrong for RocketMQ grpc case?

I try to check again.

PR #672, opened 2024-03-06
Fix duplicate traceId when use forward scheme in spring.cloud.gateway.routes.uri properties

cylx3126:
### Fix duplicate traceId when use forward scheme in spring.cloud.gateway.routes.uri properties

I provide a [Minimal, Reproducible Example](https://github.com/cylx3126/spring-cloud-gateway-skywalking-duplicate-tradeid-demo) to explain when and why this bug happens.


apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v21x/NettyRoutingFilterInterceptor.java
```diff
@@ -41,6 +45,15 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
...
 
+        if (isAlreadyRouted(exchange) || !isHttpScheme(exchange)) {
+            // In these cases, NettyRoutingFilter will not send request and HttpClientFinalizer#send will not be
+            // invoked.
+            // If HttpClientFinalizer#send don't invoke, "SpringCloudGateway/RoutingFilter" span created below will
+            // not close and produce a duplicate traceId bug.
+            // Return before create span to avoid this bug.
+            return;
+        }
+
```
wu-sheng, line 56: If there is no span created, will the trace be broken? In the `forward` case, how the tracing context is going to propagate? I don't see the new workflow in your provided MVP. The bug description is clear about why tracing context was leaked.

cylx3126: These conditions are as same as [NettyRoutingFilter](https://github.com/spring-cloud/spring-cloud-gateway/blob/dc327b79cf27cf904e9bfe4cf7e71e1db1c3fe8a/spring-cloud-gateway-server/src/main/java/org/springframework/cloud/gateway/filter/NettyRoutingFilter.java#L115-L117).

If the request is a http/https request, tracing context is going to propagate like NettyRoutingFilter#filter  -> HttpClientFinalizer#send -> HttpClientFinalizer#responseConnection .

If the request is not a http/https request, tracing context is going to propagate like NettyRoutingFilter#filter  -> ForwardRoutingFilter#filter -> DispatcherHandler#handle, in this case, HttpClientFinalizer will not participate in the propagation.

By the way, in the Spring Cloud Gateway releases before 2021/05/24, conditions there are use equals instead of equalsIgnoreCase. Tracing context will still leaked when use a older version Spring Cloud Gateway and the scheme of request is HTTP/HTTPS ,

wu-sheng: > By the way, in the Spring Cloud Gateway releases before 2021/05/24, conditions there are use equals instead of equalsIgnoreCase.

2021 seems not long ago, could you try to fix that case too?
Could we use witness class to flag those versions and use a different interceptor, in order to match the target codes?

cylx3126: No witness class around that Spring Cloud Gateway release,

How about we add some code in NettyRoutingFilterInterceptor#afterMethod instead of the check at NettyRoutingFilterInterceptor#beforeMethod, the code in NettyRoutingFilterInterceptor#afterMethod may like this:
```java
    @Override
    public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
                              Object ret) throws Throwable {
        if (ContextManager.isActive()) {
            // if HttpClientFinalizerSendInterceptor does not invoke, we will stop the span there to avoid context leak.
            ContextManager.stopSpan();
        }
        return ret;
    }
```

That is the way apm-spring-cloud-gateway-2.0.x-plugin to stop the span.

Consider two propagates above, have a check and stop the span in NettyRoutingFilterInterceptor#afterMethod may better than have a check at NettyRoutingFilterInterceptor#beforeMethod.

wu-sheng: If we can't find a way(witness class or method) to separate the two cases, it is better we could close the span in the after to avoid context leak.

cylx3126:
Ok, I will modify and test the code that close the span in afterMethod hook.

cylx3126:
Could you review this PR? I have done a refactor on my code.

wu-sheng:
impala case seems to have some issues, I am not sure the reason.

cylx3126:
>    Pulling postgres (parrotstream/postgres:10.5)...
  The image for the service you're trying to recreate has been removed. If you continue, volume data could be lost. Consider backing up your data before continuing.
  
>  pull access denied for parrotstream/postgres, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
  docker startup failure!

It seems like docker pull failed in impala test case.

wu-sheng:
I am trying to remove that case. This PR will be merged after I fix this.

wu-sheng (approved):

PR #674, opened 2024-03-12
Fix not tracing in HttpClient v5 when HttpHost(arg[0]) is null but `RoutingSupport#determineHost` works.

cylx3126:
[MRE here](https://github.com/cylx3126/skywalking-hc5-losespan-mre).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@ Release Notes.
 * Support for ActiveMQ-Artemis messaging tracing.
 * Archive the expired plugins `impala-jdbc-2.6.x-plugin`.
 * Fix a bug in Spring Cloud Gateway if HttpClientFinalizer#send does not invoke, the span created at NettyRoutingFilterInterceptor can not stop.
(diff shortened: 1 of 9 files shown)


CHANGES.md
```diff
@@ -17,6 +17,7 @@ Release Notes.
 * Support for ActiveMQ-Artemis messaging tracing.
 * Archive the expired plugins `impala-jdbc-2.6.x-plugin`.
 * Fix a bug in Spring Cloud Gateway if HttpClientFinalizer#send does not invoke, the span created at NettyRoutingFilterInterceptor can not stop.
+* Fix a bug in hc5: when the arg HttpHost of InternalHttpClient#doExecute is null, hc5 would send request but the agent don't create span
```
wu-sheng, line 20:
```suggestion
* Fix not tracing in HttpClient v5 when HttpHost(arg[0]) is null but `RoutingSupport#determineHost` works.
```

wu-sheng (approved):

PR #675, opened 2024-03-20
Support across thread tracing for SOFA-RPC

OrezzerO:
### <Feature description>
Now SOFA-PRC plugin only support in thread tracing. It does not support across thread tracing. `InvokeCallback` class is the key class which does the async callback work. This pr add the ability of across thread tracing by enhancing `InvokeCallback` class the way like `org.apache.skywalking.apm.toolkit.activation.trace.CallableOrRunnableActivation` do.


wu-sheng:
Please explain your scenario. When this happens, and what was not working.

wu-sheng:
I think you should add `callback` in https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/sofarpc-scenario and verify the callback spans through the expected file.

wu-sheng:
@OrezzerO Is https://github.com/apache/skywalking-java/pull/676 somehow affected this plugin? Is SOFA requiring JDK21? It seems not. Could you change your local JDK to JDK8 or 11 to finish this?

OrezzerO:
> @OrezzerO Is #676 somehow affected this plugin? Is SOFA requiring JDK21? It seems not. Could you change your local JDK to JDK8 or 11 to finish this?

I can change my jdk. I am not familiar with skywalking scenario test, need some time to finish this work.

wu-sheng:
I think it is not hard to run. The basic workflow is, the test is using a demo app with a compiled latest Java agent up and running. Then there is a mock traffic generator to trigger traces to the mock collector. The last step is to write expected spans to verify.

OrezzerO:
I have commit some code to verify the callback spans through the expected file. However I have some thing not very sure: 
1. Is `expectedData.yaml` file strongly depends on the stop order of each span? I am using async invoke ,so the stop order is not fixed. It may affect CI success rate.
2. I am not sure if "my expect of cross thread tracing" is suitable for skywalking. Please review `test/plugin/scenarios/sofarpc-scenario/config/expectedData.yaml` carefully.

test/plugin/scenarios/sofarpc-scenario/config/expectedData.yaml
```diff
@@ -17,6 +17,23 @@ segmentItems:
 - serviceName: sofarpc-scenario
   segmentSize: gt 2
   segments:
+  - segmentId: not null
+    spans:
+      - operationName: HEAD:/sofarpc-scenario/case/healthCheck
```
wu-sheng, line 22: `/healthCheck` should not be included in the file.

test/plugin/scenarios/sofarpc-scenario/config/expectedData.yaml
```diff
@@ -34,6 +51,37 @@ segmentItems:
         refType: CrossProcess, parentSpanId: 1, parentTraceSegmentId: not null, parentServiceInstance: not
           null, parentService: sofarpc-scenario, traceId: not null}
       skipAnalysis: 'false'
+  - segmentId: not null
```
wu-sheng, line 54: > Is expectedData.yaml file strongly depends on the stop order of each span? I am using async invoke ,so the stop order is not fixed. It may affect CI success rate.

Considering this is a new segment, which part of the span order are you referring to?

test/plugin/scenarios/sofarpc-scenario/config/expectedData.yaml
```diff
@@ -34,6 +51,37 @@ segmentItems:
...
+      spanLayer: RPCFramework
+      startTime: nq 0
+      endTime: nq 0
+      componentId: 43
+      isError: false
+      spanType: Exit
+      peer: 127.0.0.1:12200
+      skipAnalysis: false
+      tags:
+      - {key: url, value: 'bolt://127.0.0.1:12200/org.apache.skywalking.apm.testcase.sofarpc.interfaces.SofaRpcDemoService.onAppResponse(java.lang.String)'}
```
wu-sheng, line 68: By reading this, the callback is calling another RPC?

OrezzerO: Yes callback is calling another RPC. I will simplify it.

test/plugin/scenarios/sofarpc-scenario/config/expectedData.yaml
OrezzerO: image

I mean main thread segment and callback thread segment.  Which segment will stop first is not fixed.

I try to sleep in main thread to make main thread segment stop after callback thread segment. The test runs successfully. So, segment order in expectedData.yaml is ignored?

wu-sheng: Segment order is not a part of validation in the file. You don't need to worry about it.

test/plugin/scenarios/sofarpc-scenario/config/expectedData.yaml
OrezzerO: Removed it.

apm-sniffer/apm-sdk-plugin/sofarpc-plugin/src/main/java/org/apache/skywalking/apm/plugin/sofarpc/SofaBoltCallbackConstructInterceptor.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceConstructorInterceptor;
+
+public class SofaBoltCallbackConstructInterceptor implements InstanceConstructorInterceptor {
+    @Override
+    public void onConstruct(EnhancedInstance objInst, Object[] allArguments) {
+        if (ContextManager.isActive()) {
+            objInst.setSkyWalkingDynamicField(ContextManager.capture());
```
wu-sheng, line 29: There is a risk of using this field to propagate tracing context. The input(inherit InvokeCallback) is a object, which could be reused by many callback invocations. 
If SOFA RPC indicates this callback service should never be shared with different RPCs? Especially if this callback could be statusless.

An alternative solution is, you could use a wrapper, a plugin level InvokeCallback implementation to wrap the original one. We could make sure, that is one time used, and add a field(not need to be dynamic field) to hold the context.

OrezzerO: In the context of SOFARPC, InvokeCallback is a wrapper. Users directly use the SofaResponseCallback class. Each SOFARPC callback request corresponds uniquely to an InvokeCallback.

However, in the context of SOFA BOLT (BOLT being a wrapper for Netty which SOFARPC relies on), InvokeCallback may be reused.




It's a good idea to create a  plugin level wrapper, but I am afraid I will introduce an error because of  cannot dealing with classloader correctly. 

I plan to create a class `InvokeCallbackWrapper` in plugin code. How to make it loaded by business classloader? 
```java
public class InvokeCallbackWrapper implements InvokeCallback {

    private ContextSnapshot contextSnapshot;
    private InvokeCallback invokeCallback;

    public InvokeCallbackWrapper(InvokeCallback invokeCallback) {
        this.contextSnapshot = ContextManager.capture();
        this.invokeCallback = invokeCallback;

    }

    @Override
    public void onResponse(final Object o) {
        ContextManager.createLocalSpan("Thread/" + invokeCallback.getClass().getName() + "/onResponse");
        if (contextSnapshot != null) {
            ContextManager.continued(contextSnapshot);
        }
        try {
            invokeCallback.onResponse(o);
        } catch (Throwable t) {
            ContextManager.activeSpan().log(t);
            throw t;
        } finally {
            ContextManager.stopSpan();
            contextSnapshot = null;
        }

    }

    @Override
    public void onException(final Throwable throwable) {
        ContextManager.createLocalSpan("Thread/" + invokeCallback.getClass().getName() + "/onException");
        if (contextSnapshot != null) {
            ContextManager.continued(contextSnapshot);
        }
        if (throwable != null) {
            AbstractSpan abstractSpan = ContextManager.activeSpan();
            if (abstractSpan != null) {
                abstractSpan.log(throwable);
            }
        }
        try {
            invokeCallback.onException(throwable);
        } catch (Throwable t) {
            ContextManager.activeSpan().log(t);
            throw t;
        } finally {
            ContextManager.stopSpan();
            contextSnapshot = null;
        }
    }

    @Override
    public Executor getExecutor() {
        return invokeCallback.getExecutor();
    }
}

```

wu-sheng: > I plan to create a class InvokeCallbackWrapper in plugin code. How to make it loaded by business classloader?

All plugin interceptor relative codes are loaded in your target class loader, otherwise, all existing logic will fail. I think in this case, users of SOFA would not be aware of the wrapper, as it is only running internally.

wu-sheng: Classloader doesn't need to be worried, AFAIK.

apm-sniffer/apm-sdk-plugin/sofarpc-plugin/src/main/java/org/apache/skywalking/apm/plugin/sofarpc/InvokeCallbackWrapper.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private ContextSnapshot contextSnapshot;
+    private final InvokeCallback invokeCallback;
+
+    public InvokeCallbackWrapper(InvokeCallback invokeCallback) {
+        if (ContextManager.isActive()) {
+            this.contextSnapshot = ContextManager.capture();
+        }
+        this.invokeCallback = invokeCallback;
+
```
wu-sheng, line 37:
```suggestion
```

apm-sniffer/apm-sdk-plugin/sofarpc-plugin/src/main/java/org/apache/skywalking/apm/plugin/sofarpc/InvokeCallbackWrapper.java
```diff
@@ -0,0 +1,93 @@
...
+            ContextManager.continued(contextSnapshot);
+        }
+        try {
+            invokeCallback.onResponse(o);
+        } catch (Throwable t) {
+            ContextManager.activeSpan().log(t);
+            throw t;
+        } finally {
+            ContextManager.stopSpan();
+            contextSnapshot = null;
```
wu-sheng, line 53:
```suggestion
            contextSnapshot = null;
            ContextManager.stopSpan();
```

wu-sheng: `contextSnapshot = null` is 100% safe operation. The `stop Span` could throw exceptions.

apm-sniffer/apm-sdk-plugin/sofarpc-plugin/src/main/java/org/apache/skywalking/apm/plugin/sofarpc/InvokeCallbackWrapper.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+        try {
+            invokeCallback.onException(throwable);
+        } catch (Throwable t) {
+            ContextManager.activeSpan().log(t);
+            throw t;
+        } finally {
+            ContextManager.stopSpan();
+            contextSnapshot = null;
```
wu-sheng, line 77:
```suggestion
            contextSnapshot = null;
            ContextManager.stopSpan();
```

apm-sniffer/apm-sdk-plugin/sofarpc-plugin/src/main/java/org/apache/skywalking/apm/plugin/sofarpc/InvokeCallbackWrapper.java
```diff
@@ -0,0 +1,93 @@
...
+        return invokeCallback.getExecutor();
+    }
+
+    protected ContextSnapshot getContextSnapshot() {
+        return contextSnapshot;
+    }
+
+    protected InvokeCallback getInvokeCallback() {
+        return invokeCallback;
+    }
```
wu-sheng, line 92: These could be replaced by annotations.

wu-sheng:
Generally, this PR is good, just some nits. Please fix them, then it is good to merge.

apm-sniffer/apm-sdk-plugin/sofarpc-plugin/src/main/java/org/apache/skywalking/apm/plugin/sofarpc/InvokeCallbackWrapper.java
OrezzerO: Fixed it.

apm-sniffer/apm-sdk-plugin/sofarpc-plugin/src/main/java/org/apache/skywalking/apm/plugin/sofarpc/InvokeCallbackWrapper.java
```diff
@@ -20,13 +20,16 @@
...
 import com.alipay.remoting.InvokeCallback;
 import java.util.concurrent.Executor;
+import lombok.Getter;
 import org.apache.skywalking.apm.agent.core.context.ContextManager;
 import org.apache.skywalking.apm.agent.core.context.ContextSnapshot;
 import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
 
 public class InvokeCallbackWrapper implements InvokeCallback {
 
+    @Getter
```
wu-sheng, line 30: I think you should keep it in the `default` scope? This getter is generating public method.

wu-sheng:
```suggestion
    @Getter(AccessLevel.PACKAGE)
```

apm-sniffer/apm-sdk-plugin/sofarpc-plugin/src/main/java/org/apache/skywalking/apm/plugin/sofarpc/InvokeCallbackWrapper.java
```diff
@@ -0,0 +1,87 @@
...
+import lombok.Getter;
+import org.apache.skywalking.apm.agent.core.context.ContextManager;
+import org.apache.skywalking.apm.agent.core.context.ContextSnapshot;
+import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
+
+public class InvokeCallbackWrapper implements InvokeCallback {
+
+    @Getter
+    private ContextSnapshot contextSnapshot;
+    @Getter
```
wu-sheng, line 32:
```suggestion
    @Getter(AccessLevel.PACKAGE)
```

wu-sheng:
I tried to apply the change, but it seems not work.

OrezzerO:
> I tried to apply the change, but it seems not work.

Missed `import lombok.AccessLevel;` statement;  added it.

wu-sheng (approved):

PR #676, opened 2024-03-20
Bump up lombok version to 1.18.30 to support JDK21

OrezzerO:

Fix compile problem when run scenario test.  

Lombok version smaller than 1.18.30 is not support JDK21. This pr bump up lombok version to 1.18.30 to support JDK21.

Related PR: [skywalking-agent-test-tool#57](https://github.com/apache/skywalking-agent-test-tool/pull/57)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ Release Notes.
 * Archive the expired plugins `impala-jdbc-2.6.x-plugin`.
 * Fix a bug in Spring Cloud Gateway if HttpClientFinalizer#send does not invoke, the span created at NettyRoutingFilterInterceptor can not stop.
 * Fix not tracing in HttpClient v5 when HttpHost(arg[0]) is null but `RoutingSupport#determineHost` works.
+* Bump up lombok version to 1.18.30 to support JDK21.
 
 #### Documentation
 * Update docs to describe `expired-plugins`.
--- a/test/e2e/base/pom.xml
+++ b/test/e2e/base/pom.xml
@@ -53,7 +53,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <spring.boot.version>2.2.5.RELEASE</spring.boot.version>
         <guava.version>30.1.1-jre</guava.version>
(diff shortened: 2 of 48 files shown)


Superskyyy:
Eh, but how did the agent pass GitHub action testing though..?

lujiajing1126:
> Fix compile problem when run scenario test.

I suppose you have to use JDK8 or 17 for those scenarios based on the workflows defined in the repo.

For other cases, the situation is unknown. Even if you've bumped up lombok version, other dependencies may not work.

wu-sheng:
Agree, I have the same question, is this bump up necessary for all tests. Because basically, different tests are not expected to run on locked JDK versions. Some of the lib, especially old released versions, will never be able to run on JDK 21. Even if it can be, it is either transparent or relies on a new release.

wu-sheng:
> Eh, but how did the agent pass GitHub action testing though..? or java agent hasn't been tested on jdk21 yet

The agent is ready, but the target libs may or may not, depending on the cases. 
This bump up is harmless, if you have to say. But this doesn't mean, all plugins can magically run on JDK 21 when the target libs can't by themselves.

OrezzerO:
I follow the advice of [Java Agent Compiling porject](https://skywalking.apache.org/docs/skywalking-java/next/en/contribution/compiling/) and use JDK 21. I checked Github Action config,  even if  i use JDK 17, it is not all test case can run in  JDK 17, most cases are in JDK1.8 environment.  There is a little problem between documents and actual situation. 

Back to this pr, is there some others i need to do?

lujiajing1126:
> I follow the advice of [Java Agent Compiling porject](https://skywalking.apache.org/docs/skywalking-java/next/en/contribution/compiling/) and use JDK 21. There is a little problem between documents and actual situation.

The doc does not say anything about E2E. You **MUST** follow the steps in the workflow to run all E2E cases.

> I checked Github Action config, even if i use JDK 17, it is not all test case can run in JDK 17, most cases are in JDK1.8 environment.

Exactly

wu-sheng:
Close for now

PR #677, opened 2024-03-30
Update Jedis 4.x plugin to support Sharding and Cluster models.

pg-yang:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,7 +19,7 @@ Release Notes.
 * Fix a bug in Spring Cloud Gateway if HttpClientFinalizer#send does not invoke, the span created at NettyRoutingFilterInterceptor can not stop.
 * Fix not tracing in HttpClient v5 when HttpHost(arg[0]) is null but `RoutingSupport#determineHost` works.
 * Support across thread tracing for SOFA-RPC.
-* Update Jedis 4.x plugin to support Sharding/Cluster model
+* Update Jedis 4.x plugin to support Sharding and Cluster models.
 
 #### Documentation
 * Update docs to describe `expired-plugins`.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ Release Notes.
 * Fix a bug in Spring Cloud Gateway if HttpClientFinalizer#send does not invoke, the span created at NettyRoutingFilterInterceptor can not stop.
 * Fix not tracing in HttpClient v5 when HttpHost(arg[0]) is null but `RoutingSupport#determineHost` works.
 * Support across thread tracing for SOFA-RPC.
+* Update Jedis 4.x plugin to support Sharding/Cluster model
 
 #### Documentation
 * Update docs to describe `expired-plugins`.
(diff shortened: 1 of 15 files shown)


apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/ConnectionProviderConstructorInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+    public void onConstruct(final EnhancedInstance objInst, final Object[] allArguments) throws Throwable {
+        if (objInst.getSkyWalkingDynamicField() != null) {
+            return;
+        }
+        Object arg = allArguments[0];
+        if (arg instanceof Iterable<?>) {
+            Iterable<?> iterable = (Iterable<?>) arg;
+            StringBuilder sb = new StringBuilder();
+            for (Object o : iterable) {
+                sb.append(o.toString()).append(",");
```
wu-sheng, line 35: I think you should use `StringUtil#join`.

CHANGES.md
```diff
@@ -19,6 +19,7 @@ Release Notes.
 * Fix a bug in Spring Cloud Gateway if HttpClientFinalizer#send does not invoke, the span created at NettyRoutingFilterInterceptor can not stop.
 * Fix not tracing in HttpClient v5 when HttpHost(arg[0]) is null but `RoutingSupport#determineHost` works.
 * Support across thread tracing for SOFA-RPC.
+* Update Jedis 4.x plugin to support Sharding/Cluster model
```
wu-sheng, line 22:
```suggestion
* Update Jedis 4.x plugin to support Sharding and Cluster models.
```

apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/ConnectionProviderConstructorInterceptor.java
pg-yang: `StringUtil#join` doesn't support `Iterable` as parameter.

wu-sheng: I think it can support that easily.

pg-yang: I have used `Collection<?>` instead of `Iterable<?> `, and then converted the collection to String array, that can be directly applied to `StringUtil#join(final char delimiter, final String... strings)`. Anyway, the type converting is required

wu-sheng (approved):

PR #679, opened 2024-04-10
remove idleCount in druid plugin

AlchemyDing:
In fact, in the druid, `PoolingCount` and `idleCount` represent the same meaning.

associate [issue](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/11072) in opentelemetry-java-instrumentation

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/druid-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/druid/v1/PoolingAddDruidDataSourceInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/druid-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/druid/v1/PoolingAddDruidDataSourceInterceptor.java
@@ -58,10 +58,9 @@ public class PoolingAddDruidDataSourceInterceptor implements StaticMethodsAround
     }
 
     private Map<String, Function<DruidDataSourceMBean, Supplier<Double>>> getMetrics() {
-        Map<String, Function<DruidDataSourceMBean, Supplier<Double>>> metricMap = new HashMap();
+        Map<String, Function<DruidDataSourceMBean, Supplier<Double>>> metricMap = new HashMap<>();
         metricMap.put("activeCount", (DruidDataSourceMBean druidDataSource) -> () -> (double) druidDataSource.getActiveCount());
         metricMap.put("poolingCount", (DruidDataSourceMBean druidDataSource) -> () -> (double) druidDataSource.getPoolingCount());
-        metricMap.put("idleCount", (DruidDataSourceMBean druidDataSource) -> () -> (double) (druidDataSource.getPoolingCount() - druidDataSource.getActiveCount()));
         metricMap.put("lockQueueLength", (DruidDataSourceMBean druidDataSource) -> () -> (double) druidDataSource.getLockQueueLength());
         metricMap.put("maxWaitThreadCount", (DruidDataSourceMBean druidDataSource) -> () -> (double) druidDataSource.getMaxWaitThreadCount());
         metricMap.put("commitCount", (DruidDataSourceMBean druidDataSource) -> () -> (double) druidDataSource.getCommitCount());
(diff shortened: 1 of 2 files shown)


wu-sheng:
Pending on OTEL side discussion, as I want to keep the whole context continuous.

AlchemyDing:
`idleCount` should be  `druidDataSource.getPoolingCount()`
But this way it's the same as `poolingCount`.
[https://github.com/apache/skywalking-java/blob/main/apm-sniffer/apm-sdk-plugin/druid-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/druid/v1/PoolingAddDruidDataSourceInterceptor.java#L63](https://github.com/apache/skywalking-java/blob/main/apm-sniffer/apm-sdk-plugin/druid-1.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/druid/v1/PoolingAddDruidDataSourceInterceptor.java#L63)
Because the `idleCount` is not present in the `druidDataSource`, I removed the `idleCount`.

wu-sheng:
So, Druid pool doesn't have an idle concept? Then what is the status of the connection after it is used?

AlchemyDing:
> > Poolingcount represents the concept of idle
> 
> Then why don't put `Poolingcount` to `idle`, and `idle` + `activeCount`(maybe?) as the real `poolingCount`? The reason I am asking about this is, that I want to make metrics meanings aligned with general understanding.
> 
> If this is able to apply, please add comments in the codes why we change like this.

Of course, this is possible.
But before that, I also have a question. I searched for the code of skywalking-java and found that only the druid plugin has attributes called poolingCount and idleCount. Do we need to change the original meaning of the druid for these two attributes in APM, perhaps we need a standardized semantic convention?
WDYT?

wu-sheng:
> Of course, this is possible.
> But before that, I also have a question. I searched for the code of skywalking-java and found that only the druid plugin has attributes called poolingCount and idleCount. Do we need to change the original meaning of the druid for these two attributes in APM, perhaps we need a standardized semantic convention?

Good question. It should be good to remove that, rather than recalculate.

__

Meanwhile, to make CI passed, you should fix the expected file.

https://github.com/apache/skywalking-java/blob/eb1041675cafb31dcc8da1fc0a45bc796d5b912c/test/plugin/scenarios/druid-1.x-scenario/config/expectedData.yaml#L223-L227

AlchemyDing:
Therefore, I have a proposal to standardize our semantic conventions, which can avoid potential semantic ambiguity in the future.

AlchemyDing:
> > 当然，这是可能的。
> > 但在此之前，我还有一个问题。我搜索了skywalking-java的代码，发现只有druid插件有叫做poolingCount和idleCount的属性。我们是否需要在APM中改变druid对于这两个属性的原始含义，也许我们需要一个标准化的语义约定？
> 
> 好问题。最好删除它，而不是重新计算。
> 
> __
> 
> 同时，为了让CI通过，你应该修复预期的文件。
> 
> https://github.com/apache/skywalking-java/blob/eb1041675cafb31dcc8da1fc0a45bc796d5b912c/test/plugin/scenarios/druid-1.x-scenario/config/expectedData.yaml#L223-L227

Done.

AlchemyDing:
> > Of course, this is possible.
> > But before that, I also have a question. I searched for the code of skywalking-java and found that only the druid plugin has attributes called poolingCount and idleCount. Do we need to change the original meaning of the druid for these two attributes in APM, perhaps we need a standardized semantic convention?
> 
> Good question. It should be good to remove that, rather than recalculate.
> 
> __
> 
> Meanwhile, to make CI passed, you should fix the expected file.
> 
> https://github.com/apache/skywalking-java/blob/eb1041675cafb31dcc8da1fc0a45bc796d5b912c/test/plugin/scenarios/druid-1.x-scenario/config/expectedData.yaml#L223-L227

Done.

wu-sheng:
> Therefore, I have a proposal to standardize our semantic conventions, which can avoid potential semantic ambiguity in the future.

We should consider that in a way.

wu-sheng (approved):

wu-sheng (approved):

PR #682, opened 2024-04-18
Polish NPE fix in handleMethodException method of apm-jdk-threadpool-plugin

ForrestWang123:
### Fix Fix NPE in handleMethodException method of apm-jdk-threadpool-plugin

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #12128


wu-sheng:
What is the new change about?

wu-sheng (approved):
Look like a polish for the previous PR.

PR #683, opened 2024-04-25
add Support for C3P0 connection pool tracing

CzyerChen:
### Add an agent plugin to support <C3P0> version: 0.9.0.x ~ 0.10.0(latest)
- [x] Add two test case for the new plugin because of a groupId difference in the dependency package
one for 0.9.0.x ~ 0.9.1.x :  <groupId>c3p0</groupId>
one for 0.9.2.x ~ 0.10.0 :  <groupId>com.mchange</groupId>
- [x] Add a component id in [the component-libraries.yml](https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/component-libraries.yml)  https://github.com/apache/skywalking/pull/12162

report connection pool metrics:
image

about the trace(getConnection, closeConnection):
image
image

### DIFF
--- a/docs/en/setup/service-agent/java-agent/Supported-list.md
+++ b/docs/en/setup/service-agent/java-agent/Supported-list.md
@@ -167,6 +167,10 @@ The meter plugin provides the advanced metrics collections, which are not a part
   * [Dubbo](https://github.com/apache/dubbo) 2.5.x -> 2.7.x
   * [Jetty](https://github.com/eclipse/jetty.project) 9.1.x -> 11.x
   * [Grizzly](https://github.com/eclipse-ee4j/grizzly) 2.3.x -> 4.x
+  * [Apache Commons DBCP](https://github.com/apache/commons-dbcp) 2.x
+  * [Alibaba Druid](https://github.com/alibaba/druid) 1.x
+  * [HikariCP](https://github.com/brettwooldridge/HikariCP) 3.x -> 4.x
+  * [C3P0](https://github.com/swaldman/c3p0) 0.9.0 -> 0.10.0
 ___
 ¹Due to license incompatibilities/restrictions these plugins are hosted and released in 3rd part repository,
  go to [SkyAPM java plugin extension repository](https://github.com/SkyAPM/java-plugin-extensions) to get these.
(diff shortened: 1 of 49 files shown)


wu-sheng:
Please fix CI.

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -148,6 +148,7 @@ metrics based on the tracing data.
   * [Apache Commons DBCP](https://github.com/apache/commons-dbcp) 2.x
   * [Alibaba Druid](https://github.com/alibaba/druid) 1.x
   * [HikariCP](https://github.com/brettwooldridge/HikariCP) 3.x -> 4.x
+  * [C3P0](https://github.com/swaldman/c3p0) 0.9.0 -> 0.10.0
```
wu-sheng, line 151: I think you should update meter plugin list as well. I noticed, it seems other `Connection Pool` plugins are not in the list as well, could you check?

wu-sheng: Same page here, https://github.com/apache/skywalking-java/blob/main/docs/en/setup/service-agent/java-agent/Supported-list.md#meter-plugins

CzyerChen: ok

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -166,6 +167,10 @@ The meter plugin provides the advanced metrics collections, which are not a part
   * [Dubbo](https://github.com/apache/dubbo) 2.5.x -> 2.7.x
   * [Jetty](https://github.com/eclipse/jetty.project) 9.1.x -> 11.x
   * [Grizzly](https://github.com/eclipse-ee4j/grizzly) 2.3.x -> 4.x
+  * [Apache Commons DBCP](https://github.com/apache/commons-dbcp) 2.x
```
wu-sheng, line 170:
```suggestion
* Connection Pool
  * [Apache Commons DBCP](https://github.com/apache/commons-dbcp) 2.x
```

I think should be this.

wu-sheng (approved):

PR #688, opened 2024-05-11
fix TracingSegmentRunner

JoeCqupt:
TracingSegmentRunner  override method withAfters made Junit @after annotation not work

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/12210


lujiajing1126:
Nice catch. But you have to verify all test cases work as expected. I notice we've declared many `cleanup`s or `teardown`s. But they are actually not running at all.

CodePrometheus:
I think `org.apache.skywalking.apm.agent.core.test.tools.TracingSegmentRunner#withAfters` also needs to be modified.

CodePrometheus (approved):
Now I think all cleanup and teardown could work correctly, the code LGTM, thanks @JoeCqupt for your attentive discovery and don't forget to update CHANGES, let's wait for other teachers to review.

wu-sheng:
> Now I think all cleanup and teardown could work correctly, the code LGTM, thanks @JoeCqupt for your attentive discovery and don't forget to update CHANGES, let's wait for other teachers to review.

As no feature code change, no changelog required.

PR #689, opened 2024-05-13
Add a forceIgnoring mechanism and apply it to the plugins (Spring Clo…

gzlicanyi:
Add a [forceIgnoring mechanism](https://github.com/apache/skywalking/discussions/12161) and apply it to the plugins (Spring Cloud Gateway, HttpClient, JDK-ForkJoinPool, JDK-Threading, JDK-ThreadPool, Toolkit-Trace, Toolkit-WebFlux)

# proposal
## Motivation 
context: https://github.com/apache/skywalking/discussions/12161.

Generally speaking, if the source Span is ignored, the downstream Span also needs to be ignored, such as in same-thread operations and cross-process operations. However, for Spans created by cross-thread operations, such as in the scenario of Spring Cloud Gateway, there is currently no mechanism to make them ignored along with the parent Span.

The force-ignoring mechanism is a supplement to the ignoring mechanism for cross-thread operations.

## Architecture Graph
image

## Proposed Changes
The reason why you can only modify within the ContextManager#createLocalSpan method, and not within the ContextManager#continued method, is because the latter cannot change an already created Context and Span.

I haven't found a way to support this feature without disrupting the existing API, because the tracing context is created before the snapshot is continued.

Currently modified plugins are:
httpasyncclient-4.x
httpclient-5.x
jdk-forkjoinpool
jdk-threading
jdk-threadpool
spring-cloud-gateway-2.0.x
spring-cloud-gateway-2.1.x
spring-cloud-gateway-3.x
spring-cloud-gateway-4.x
toolkit-trace-activation
toolkit-webflux-activation

## Imported Dependencies libs and their licenses.
No new dependency.

## Compatibility
In scenarios where it is necessary to keep the span node of the entire trace intact, such as in Spring Cloud Gateway, the call to the ContextManager#continued method needs to be replaced with the ContextManager#createLocalSpan method that includes a snapshot.


wu-sheng:
Please write a proposal about what is a force-ignoring mechanism. The discussion we did was only about the context.

gzlicanyi:
> Please write a proposal about what is a force-ignoring mechanism. The discussion we did was only about the context.

Do you have a proposal template?

wu-sheng:
For an official proposal, you could refer to SWIP(https://skywalking.apache.org/docs/main/next/en/swip/readme/), which is mandatory for OAP, but not for the agent.
A key part of a proposal is about why you want to change, how you design looks like, what are the impacts. Then we could discuss for more details.

A reminder for you, the way you changes, is making a significant impact for existing APIs, this may not be a good way.

gzlicanyi:
> Please write a proposal about what is a force-ignoring mechanism. The discussion we did was only about the context.

Added.

wu-sheng:
Add what?

gzlicanyi:
> Add what?

here: https://github.com/apache/skywalking-java/pull/689#issue-2292115894

wu-sheng:
I think we need to separate the context. Cross thread is different from cross process. 
Cross process is defined here, https://skywalking.apache.org/docs/main/latest/en/api/x-process-propagation-headers-v3/. This can't be changed as it crosses all language agents.

About crosee thread, the status should only documented in the snapshot. But I am not clear about how you changed a sampled tracing context into a ignoring context?

wu-sheng:
I am feeling you are doing a hijack only, which should not be the official way.

gzlicanyi:
> I think we need to separate the context. Cross thread is different from cross process. Cross process is defined here, https://skywalking.apache.org/docs/main/latest/en/api/x-process-propagation-headers-v3/. This can't be changed as it crosses all language agents.
> 
> About crosee thread, the status should only documented in the snapshot. But I am not clear about how you changed a sampled tracing context into a ignoring context?

This question can be divided into two parts:
1. In the scenario of cross-thread, is it necessary to continue the upstream context?
2. If the first point is necessary, how can we modify it to better align with the official way?

wu-sheng:
I don't have direct answer, if I had, I already told you in the last discussion.

The reason it doesn't exist is, the tracing context is created before the snapshot is continued.

You proposal should cover about how.

The cross process is more complex. And sampling is service(process) oriented. There is a flag in the protocol about sampling, but you have to evaluate all agent codes(all languages), about whether there will be an impact.

wu-sheng:
When this is about ContextManager, it should not affect TracingContext#CreateLocal span. Only snapshot should be changed, and ContextManager asked TracingContext switching to IgnoringTracingCntext with the same stack. Also, the plugins should not be changed, and all span and context manager APIs should not be changed.

gzlicanyi:
> When this is about ContextManager, it should not affect TracingContext#CreateLocal span. Only snapshot should be changed, and ContextManager asked TracingContext switching to IgnoringTracingCntext with the same stack. Also, the plugins should not be changed, and all span and context manager APIs should not be changed.

If the TracingContext switches to IgnoringTracingContext in the continued method, the span reference created by the ContextManager#createLocalSpan method cannot be changed, and the plugin still uses the span before the switch.
Additionally, the already created but useless TracingContext will increment the sampling count of the operationName by one.

wu-sheng:
I think you got it wrong. It should be in ContextManager#continued method, which controls the context.

wu-sheng:
This is a force sampling(unsampling), we don't need to worry about sampling counter.

gzlicanyi:
> This is a force sampling(unsampling), we don't need to worry about sampling counter.



> This is a force sampling(unsampling), we don't need to worry about sampling counter.

ForceSampling might not necessarily be true. I feel that it could still impact the count of the SamplingService.

wu-sheng:
> > This is a force sampling(unsampling), we don't need to worry about sampling counter.
> 
> ForceSampling might not necessarily be true. I feel that it could still impact the count of the SamplingService.

It just counts one more, so what? Sampling is never to be accurate.

wu-sheng:
> > I think you got it wrong. It should be in ContextManager#continued method, which controls the context.
> 
> Understood. How can the [span](https://github.com/apache/skywalking-java/blob/1a010470076ba7f206f4fd6559d04136bea73417/apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.0.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v20x/NettyRoutingFilterInterceptor.java#L44) used by the plugin be changed to a NoopSpan?

Consider creating `AbstractTracerContext#forceIgnoring. `TracingContext#forceIgnoring` could return a `IgnoredTracerContext` with the same stack depth, to make sure no tracing context stack leak. `IgnoredTracerContext` just simply returns **this** as no change is required.

gzlicanyi:
> > > I think you got it wrong. It should be in ContextManager#continued method, which controls the context.
> > 
> > 
> > Understood. How can the [span](https://github.com/apache/skywalking-java/blob/1a010470076ba7f206f4fd6559d04136bea73417/apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.0.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v20x/NettyRoutingFilterInterceptor.java#L44) used by the plugin be changed to a NoopSpan?
> 
> Consider creating `AbstractTracerContext#forceIgnoring`. `TracingContext#forceIgnoring` could return a `IgnoredTracerContext` with the same stack depth, to make sure no tracing context stack leak. `IgnoredTracerContext` just simply returns **this** as no change is required.

IgnoredTracerContext retains the activeSpanStack of TracingContext, do you think there's no problem with it?

gzlicanyi:
> > > > I think you got it wrong. It should be in ContextManager#continued method, which controls the context.
> > > 
> > > 
> > > Understood. How can the [span](https://github.com/apache/skywalking-java/blob/1a010470076ba7f206f4fd6559d04136bea73417/apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.0.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v20x/NettyRoutingFilterInterceptor.java#L44) used by the plugin be changed to a NoopSpan?
> > 
> > 
> > Consider creating `AbstractTracerContext#forceIgnoring`. `TracingContext#forceIgnoring` could return a `IgnoredTracerContext` with the same stack depth, to make sure no tracing context stack leak. `IgnoredTracerContext` just simply returns **this** as no change is required.
> 
> IgnoredTracerContext retains the activeSpanStack of TracingContext, do you think there's no problem with it?

If there is no problem, indeed the 'continued' method can complete the transition of the context. However, for the leftover spans, I am not entirely sure if they might cause other issues.

wu-sheng:
Which problem do you mean  No matter what spans they are. As they are not reported, they are fine to be GCed.

gzlicanyi:
> Which problem do you mean No matter what spans they are. As they are not reported, they are fine to be GCed.

I have made alterations using a new way and submitted another [PR](https://github.com/apache/skywalking-java/pull/692). Can you check if there are any problems?

wu-sheng:
Please update here.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
```diff
@@ -34,14 +34,21 @@ public class IgnoredTracerContext implements AbstractTracerContext {
     private static final NoopSpan NOOP_SPAN = new NoopSpan();
     private static final String IGNORE_TRACE = "Ignored_Trace";
 
+    private LinkedList<AbstractSpan> activeSpanStack;
```
wu-sheng, line 37: Why need this?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
```diff
@@ -34,14 +34,21 @@ public class IgnoredTracerContext implements AbstractTracerContext {
...
     public IgnoredTracerContext() {
-        this.stackDepth = 0;
+        this.activeSpanStack = new LinkedList<>();
+        this.correlationContext = new CorrelationContext();
+        this.extensionContext = new ExtensionContext();
+        this.profileStatusContext = ProfileStatusContext.createWithNone();
+    }
+
+    public IgnoredTracerContext(LinkedList<AbstractSpan> activeSpanStack) {
+        this.activeSpanStack = activeSpanStack;
```
wu-sheng, line 51: We only need the depth to be initialized based on stack depth, rather than holding all

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
```diff
@@ -34,14 +34,21 @@ public class IgnoredTracerContext implements AbstractTracerContext {
...
     private final ProfileStatusContext profileStatusContext;
 
-    private int stackDepth;
-
     public IgnoredTracerContext() {
-        this.stackDepth = 0;
+        this.activeSpanStack = new LinkedList<>();
+        this.correlationContext = new CorrelationContext();
+        this.extensionContext = new ExtensionContext();
+        this.profileStatusContext = ProfileStatusContext.createWithNone();
```
wu-sheng, line 47: You just need to call new constructor.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
```diff
@@ -84,34 +91,35 @@ public int getSpanId() {
 
     @Override
     public AbstractSpan createEntrySpan(String operationName) {
-        stackDepth++;
+        activeSpanStack.addLast(NOOP_SPAN);
```
wu-sheng, line 94: All logic here should not be changed.

wu-sheng: We just need depth, no further.

wu-sheng (changes requested):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/IgnoredTracerContext.java
gzlicanyi: Because the old span is still in use, if it's discarded, [span#prepareForAsync](https://github.com/apache/skywalking-java/blob/1a010470076ba7f206f4fd6559d04136bea73417/apm-sniffer/optional-plugins/optional-spring-plugins/optional-spring-cloud/gateway-2.0.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/spring/cloud/gateway/v20x/NettyRoutingFilterInterceptor.java#L45) will throw an error.

gzlicanyi: image
image
image

wu-sheng: This is not friendly for GC. We could add `ingored` flag to the span and skip this check.

wu-sheng: Another option maybe to create a ForceIgnoreTracerContext to deal with created span cases. 
And this only happens in very rare case, so we should process that only when we checked all active spans, which have one running in async mode.

gzlicanyi: Do you prefer replacing the span's context with IgnoredTracerContext for old spans with old [context references](https://github.com/apache/skywalking-java/blob/4dfc1e85b56e75a1bf2bae11d34d8c0fe234bed0/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java#L68), or would you rather ignore any operations involving context within the span?

wu-sheng: I didn't dig in so deeply, as I am feeling this change is not easy.

I just provided two possibilities, no preference. One key part is, what you are adding is not happening as always, you should not make the agent kernel costs too much.

I don't whether this is possible.

wu-sheng: That was why I said, this could be done easily if this only happens in your private fork. But in general, I don't have a good idea could be done easily. 

But this has to be very good, as your changes are going to affect every user for sure.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/ContextManager.java
```diff
@@ -163,7 +163,12 @@ public static void continued(ContextSnapshot snapshot) {
             throw new IllegalArgumentException("ContextSnapshot can't be null.");
         }
         if (!snapshot.isFromCurrent()) {
-            get().continued(snapshot);
+            if (snapshot.isValid()) {
```
wu-sheng, line 166:
```suggestion
            // Invalid snapshot is only created by {@link IgnoredTracerContext#capture()}.
            // When the snapshot is not valid, need to force ignoring the current tracing context.
            if (snapshot.isValid()) {
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/AbstractTracerContext.java
```diff
@@ -138,4 +138,10 @@ public interface AbstractTracerContext {
      * Get current primary endpoint name
      */
     String getPrimaryEndpointName();
+
+    /**
+     *  Change the current context to IgnoredTracerContext.
```
wu-sheng, line 143:
```suggestion
     *  Change the current context to be in ignoring status.
```

It doesn't have to be `IgnoredTracerContext`. The only effect is to make the spans not being reported anymore.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/trace/AbstractTracingSpan.java
```diff
@@ -63,7 +63,8 @@ public abstract class AbstractTracingSpan implements AbstractSpan {
     private volatile boolean isAsyncStopped = false;
 
     /**
-     * The context to which the span belongs
+     * The context to which the span belongs.
+     * This should not be called when ignored is set to true.
```
wu-sheng, line 67:
```suggestion
     * This should not be called when {@link #ignored} is true.
```

wu-sheng:
It seems the UTs failing. Please recheck and fix.

wu-sheng:
Please follow the comments to polish the codes. Generally, this PR is good.

wu-sheng (approved):

wu-sheng:
Thanks for adding this.

PR #691, opened 2024-05-13
Fix method name missing in spring-plugins:scheduled-annotation-plugin with spring 6.1.x

CzyerChen:
### Fix method name missing in spring-plugins:scheduled-annotation-plugin with spring 6.1.x

Since Spring 6.1.x,  the intercepted method deprecated

```java
  @Deprecated(
        since = "6.1"
    )
    @Nullable
    protected Runnable createRunnable(Object target, Method method) {
        return null;
    }
```

use the constructor method instead:

```java
  public ScheduledMethodRunnable(Object target, Method method, @Nullable String qualifier, Supplier<ObservationRegistry> observationRegistrySupplier) {
        this.target = target;
        this.method = method;
        this.qualifier = qualifier;
        this.observationRegistrySupplier = observationRegistrySupplier;
    }
```

----
Before adjustment:

image

the method name missing wiith Spring 6.1.x

Now:

image

Meanwhile, optimize test cases of `spring-scheduled`.


CHANGES.md
```diff
@@ -11,6 +11,7 @@ Release Notes.
 * Use a daemon thread to flush logs.
 * Fix typos in `URLParser`.
 * Add support for `Derby`/`Sybase`/`SQLite`/`DB2`/`OceanBase` jdbc url format in `URLParser`.
+* Optimize spring-plugins:scheduled-annotation-plugin compatibility.
```
wu-sheng, line 14:
```suggestion
* Optimize spring-plugins:scheduled-annotation-plugin compatibility about Spring 6.1.x support.
```

wu-sheng:
I can see RocketMQ case fails again.

CzyerChen:
> I can see RocketMQ case fails again.

OK，I will check RocketMQ case first.

CzyerChen:
> I can see RocketMQ case fails again.

The problem is not reproduced locally, and the sending and receiving of async messages are enhanced.

Please do inform me when the recurrence of the problem.

wu-sheng:
Are we sending too frequently? The test GHA env is only 3u8g for all runtime. It could be super slow.

wu-sheng (dismissed):

CzyerChen:
> Are we sending too frequently? The test GHA env is only 3u8g for all runtime. It could be super slow.

May have an impact.

wu-sheng:
I can see RocketMQ fails again. From what I see, 50% chance fails

wu-sheng (approved):

PR #692, opened 2024-05-14
Add a forceIgnoring mechanism in a CROSS_THREAD scenario.

gzlicanyi:

context: https://github.com/apache/skywalking-java/pull/689

Implement the forceIgnoring mechanism in a CROSS_THREAD scenario by switching the context in the 'continued' method.


wu-sheng:
Why another? Please hold on the old one. We should keep context continuously.

PR #694, opened 2024-05-25
Support for tracing the callbacks of async methods in elastiicsearch-6.x/7.x-plugin

CzyerChen:
### Support for tracing the callbacks of async methods in elastiicsearch-6.x/7.x-plugin

callbacks(onResponse/onFailure) for 
- createAsync
- deleteAsync
- analyzeAsync
- refreshAsync
- getAsync
- searchAsync
- indexAsync
- updateAsync
- scrollAsync
- searchTemplateAsync
- clearScrollAsync
- deleteByQueryAsync

image

Trace details:
ElasticsearchDeleteRequestonResponse
ElasticsearchGetReguest
ElasticsearchactionGet

record `transmission.latency` when callback:

image

7.0.0 - 7.2.1 not included in the test because analyze method appears since 7.3.0

### DIFF
--- a/test/plugin/scenarios/elasticsearch-7.x-scenario/src/main/java/org/apache/skywalking/apm/testcase/elasticsearch/RestHighLevelClientCase.java
+++ b/test/plugin/scenarios/elasticsearch-7.x-scenario/src/main/java/org/apache/skywalking/apm/testcase/elasticsearch/RestHighLevelClientCase.java
@@ -25,6 +25,7 @@ import java.util.UUID;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.skywalking.apm.testcase.elasticsearch.controller.CaseController;
+import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthRequest;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthResponse;
 import org.elasticsearch.action.admin.cluster.settings.ClusterGetSettingsRequest;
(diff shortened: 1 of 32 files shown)


test/plugin/scenarios/elasticsearch-7.x-scenario/support-version.list
```diff
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# 7.0.0 - 7.17.12 have been tested, and 7.0.0 - 7.2.1 not included in the test
```
wu-sheng, line 17: It is good to see these are all verified.

test/plugin/scenarios/elasticsearch-7.x-scenario/src/main/java/org/apache/skywalking/apm/testcase/elasticsearch/RestHighLevelClientCase.java
```diff
@@ -97,27 +100,37 @@ public String elasticsearch() throws Exception {
...
             createIndex(indexName);
+            createIndexAsync(indexName2);
 
             // index
             index(indexName);
+            indexAsync(indexName2);
 
             // refresh
             client.indices().refresh(new RefreshRequest(indexName), RequestOptions.DEFAULT);
+            Thread.sleep(1000);
```
wu-sheng, line 111: Do we have another way to verify the request has been verified? 
Such as, do it through query or callback?

wu-sheng: `Sleep` usually is not stable when doing this, when resource is limited in GHA scenario.

CzyerChen: OK, I will improve that.

wu-sheng:
Could you verify that, is virtual database working as expected?

CzyerChen:
> Could you verify that, is virtual database working as expected?

Looks it's working as well.
image
image
image

wu-sheng (dismissed):

CzyerChen:
ES-7.x case may fail again due to the limited resources.

wu-sheng:
Test cases fail. You may need some recheck.

CzyerChen:
> Test cases fail. You may need some recheck.

Let me think twice.

wu-sheng (approved):

PR #695, opened 2024-05-31
Fix to Ensure All Rules Satisfy not isInterface

wengyingjian:
    According to the code principles of `net.bytebuddy.matcher.ElementMatcher.Junction`:
    In the current code, the running effect can only ensure that the first `NamedMatch` satisfies `not isInterface`:
   ` (((NamedMatch) and (not(isInterface))) or IndirectMatch) or...IndirectMatch.`
    
    After moving `not isInterface` to the end, the running effect can ensure that all rules satisfy `not isInterface`:
   ` (((NamedMatch) or (IndirectMatch)) or...IndirectMatch) and not(isInterface).`

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginFinder.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginFinder.java
@@ -95,13 +95,13 @@ public class PluginFinder {
                 return nameMatchDefine.containsKey(target.getActualName());
             }
         };
-        judge = judge.and(not(isInterface()));
         for (AbstractClassEnhancePluginDefine define : signatureMatchDefine) {
             ClassMatch match = define.enhanceClass();
             if (match instanceof IndirectMatch) {
                 judge = judge.or(((IndirectMatch) match).buildJunction());
             }
         }
+        judge = judge.and(not(isInterface()));
         return new ProtectiveShieldMatcher(judge);
     }
 
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginFinder.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginFinder.java
@@ -101,7 +101,8 @@ public class PluginFinder {
                 judge = judge.or(((IndirectMatch) match).buildJunction());
             }
         }
-        judge = judge.and(not(isInterface()));
+        // Filter out all matchers returns to exclude pure interface types.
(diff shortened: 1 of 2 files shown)


CodePrometheus:
> In the current code, the running effect can only ensure that the first NamedMatch satisfies not isInterface

@wengyingjian I think the current `Junction` expression is
~~~
(A and not(isInterface())) or (B and not(isInterface())) or ...
~~~
Is my understanding right?

wu-sheng:
I think original request is all things should be not an interface.

First or last are not accurate.

It should use and with all other condition or before all other checks run.

wengyingjian:
Yes, because I didn't find a place that required modifying the bytecode for interfaces, I think all matching rules should be non-interface based.

 If that's the case, adding and (not isInterface) in the for loop is indeed a safer approach.

Additionally, I've noticed that some custom IndirectMatch instances include not interface, while some do not.
I'm not sure if this means there is already a place that requires modifying the interface bytecode, or if it's just to maintain flexibility, allowing subclasses to implement it themselves.

wu-sheng:
In theory, there is no point to intrument an interface, especially we require to add new field to the target.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/PluginFinder.java
```diff
@@ -95,13 +95,13 @@ public boolean matches(NamedElement target) {
...
             }
         };
-        judge = judge.and(not(isInterface()));
         for (AbstractClassEnhancePluginDefine define : signatureMatchDefine) {
             ClassMatch match = define.enhanceClass();
             if (match instanceof IndirectMatch) {
                 judge = judge.or(((IndirectMatch) match).buildJunction());
             }
         }
+        judge = judge.and(not(isInterface()));
```
wu-sheng, line 104:
```suggestion
        // Filter out all matchers returns to exclude pure interface types.
        judge = not(isInterface()).and(judge);
```

I think this should be the way you propose. The result will be the same as your wrote, but the logic is more clear.

`not an interface` and `all other pre-built checker`.

wu-sheng:
> I've noticed that some custom IndirectMatch instances include not interface, while some do not.

I think those are not necessary, but as the bug you are fixing, they may be added as it is.

wu-sheng:
> > In the current code, the running effect can only ensure that the first NamedMatch satisfies not isInterface
> 
> @wengyingjian I think the current `Junction` expression is
> 
> ```
> (A and not(isInterface())) or (B and not(isInterface())) or ...
> ```
> 
> Is my understanding right?

I think the logic error is, that the `is not interface` could be overridden by the `or` from the 2nd condition, as it is flat.

PR #696, opened 2024-06-06
Fix the opentracing toolkit SPI config

xiaqi1210:
1. The directory for SPI should be `/META-INF/services`, not `/META-INF.services`.
2. There is an issue with the letter casing in the class name. It should be `SkywalkingTracer`, not `SkyWalkingTracer`.


wu-sheng (approved):
I am OK to merge, but, OT is archived over 2 years(since January 31, 2022). 

It is the time to get rid of it.

PR #703, opened 2024-07-04
Fix problem in the jetty-client-9.x which leading original listener reading empty content

mei-zengxin:
### Fix problem in the jetty-client-9.x which leading original listener reading empty content

when using jetty-client 9.x send async request, reponse body is always empty

because the onContent callback could not be executed

```java
        request.send(new BufferingResponseListener() {
            @Override
            public void onComplete(Result result) {
                try { 
                    //this is always empty when use agent
                    byte[] responseBytes = getContent();
                    if (responseBytes == null || responseBytes.length == 0) {
                        System.out.println("responseBytes is empty");
                    }
                    System.out.println(new String(responseBytes, "UTF-8"));
                } catch (UnsupportedEncodingException e) {
                    throw new RuntimeException(e);
                }
            }
        });
```

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/jetty-plugin/jetty-client-9.x-plugin/pom.xml
+++ b/apm-sniffer/apm-sdk-plugin/jetty-plugin/jetty-client-9.x-plugin/pom.xml
@@ -31,7 +31,7 @@
     <url>http://maven.apache.org</url>
 
     <properties>
-        <jetty-client.version>9.1.0.v20131115</jetty-client.version>
+        <jetty-client.version>9.4.55.v20240627</jetty-client.version>
     </properties>
 
     <dependencies>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ Release Notes.
 * Improve 4x performance of ContextManagerExtendService.createTraceContext()
 * Add a plugin that supports the Solon framework.
 * Fixed issues in the MySQL component where the executeBatch method could result in empty SQL statements .
+* Fix problem in the jetty-client-9.x which leading original listener reading empty content
 
 
 All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/213?closed=1)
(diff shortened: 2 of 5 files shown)


apm-sniffer/apm-sdk-plugin/jetty-plugin/jetty-client-9.x-plugin/pom.xml
```diff
@@ -31,7 +31,7 @@
     <url>http://maven.apache.org</url>
 
     <properties>
-        <jetty-client.version>9.1.0.v20131115</jetty-client.version>
+        <jetty-client.version>9.4.55.v20240627</jetty-client.version>
```
wu-sheng, line 34: Why change this? Would this break preview compatibility?

wu-sheng (changes requested):

apm-sniffer/apm-sdk-plugin/jetty-plugin/jetty-client-9.x-plugin/pom.xml
mei-zengxin: because org.eclipse.jetty.client.api.Response.AsyncContentListener cound not be found in 9.1.x version
if the wrapper not implementing it, application will also get empty result when using jetty client 9.2.x or later, 

by testing, it is  compatible with preview version 9.1.0.v20131115
, 9.2.12.v20150709, 9.3.30.v20211001 and 9.4.55.v20240627

wu-sheng:
You should fix CI, and add more versions of newly supported to verify.

wu-sheng:
You should update the following test scenarios and expected files.

- https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/jetty-scenario
- https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/jetty-10.x-scenario
- https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/jetty-11.x-scenario

apm-sniffer/apm-sdk-plugin/jetty-plugin/jetty-client-9.x-plugin/pom.xml
wu-sheng: If it doesn't exist in `9.1.0.v20131115`, how do we make sure of compatibility? This doesn't make sense.

wu-sheng:
Close as no update.

PR #704, opened 2024-07-10
Optimize Spring and Tomcat version recognization problem

honganan:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/12425 .

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/ClassUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/ClassUtil.java
@@ -21,21 +21,9 @@ package org.apache.skywalking.apm.agent.core.util;
 import java.util.HashSet;
 import java.util.Set;
 
-/**
- * @author honganan
- */
 public class ClassUtil {
 
-    public static boolean isAssignableFrom(String superClassName, Class<?> subClass) {
-        if (superClassName == null || subClass == null) {
-            return false;
-        }
-
-        Set<String> allSuperClassesAndInterfaces = getAllSuperClassesAndInterfaces(subClass);
-        return allSuperClassesAndInterfaces.contains(superClassName);
-    }
-
-    private static Set<String> getAllSuperClassesAndInterfaces(Class<?> clazz) {
+    public static Set<String> getAllSuperClassesAndInterfaces(Class<?> clazz) {
         Set<String> result = new HashSet<>();
         while (clazz != null) {
             result.add(clazz.getName());
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/ClassUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/ClassUtil.java
@@ -38,15 +38,12 @@ public class ClassUtil {
     private static Set<String> getAllSuperClassesAndInterfaces(Class<?> clazz) {
         Set<String> result = new HashSet<>();
         while (clazz != null) {
-            // 添加当前类
             result.add(clazz.getName());
 
-            // 添加所有接口
             for (Class<?> iface : clazz.getInterfaces()) {
                 result.addAll(getAllSuperClassesAndInterfaces(iface));
             }
 
-            // 获取超类
             clazz = clazz.getSuperclass();
         }
         return result;
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/ClassUtil.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/ClassUtil.java
@@ -1,0 +1,54 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.agent.core.util;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * @author honganan
+ */
+public class ClassUtil {
+
+    public static boolean isAssignableFrom(String superClassName, Class<?> subClass) {
+        if (superClassName == null || subClass == null) {
+            return false;
+        }
+
+        Set<String> allSuperClassesAndInterfaces = getAllSuperClassesAndInterfaces(subClass);
+        return allSuperClassesAndInterfaces.contains(superClassName);
+    }
+
+    private static Set<String> getAllSuperClassesAndInterfaces(Class<?> clazz) {
+        Set<String> result = new HashSet<>();
+        while (clazz != null) {
+            // 添加当前类
+            result.add(clazz.getName());
+
+            // 添加所有接口
+            for (Class<?> iface : clazz.getInterfaces()) {
+                result.addAll(getAllSuperClassesAndInterfaces(iface));
+            }
+
+            // 获取超类
+            clazz = clazz.getSuperclass();
+        }
+        return result;
+    }
+}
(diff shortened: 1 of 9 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/util/ClassUtil.java
```diff
@@ -0,0 +1,54 @@
...
+            // 添加当前类
+            result.add(clazz.getName());
+
+            // 添加所有接口
+            for (Class<?> iface : clazz.getInterfaces()) {
+                result.addAll(getAllSuperClassesAndInterfaces(iface));
+            }
+
+            // 获取超类
+            clazz = clazz.getSuperclass();
```
wu-sheng, line 50: Chinese is not allowed in the source codes.

apm-sniffer/apm-sdk-plugin/spring-plugins/mvc-annotation-commons/src/main/java/org/apache/skywalking/apm/plugin/spring/mvc/commons/interceptor/AbstractMethodInterceptor.java
```diff
@@ -111,7 +84,7 @@ public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allAr
             if (stackDepth == null) {
                 final ContextCarrier contextCarrier = new ContextCarrier();
 
-                if (IN_SERVLET_CONTAINER && IS_JAVAX && HttpServletRequest.class.isAssignableFrom(request.getClass())) {
+                if (isAssignableFrom(JAVAX_SERVLET_REQUEST_CLASS, request.getClass())) {
```
wu-sheng, line 87: If isAssignableFrom executes every time, you will slow the whole process.

wu-sheng: `static` block is an important mechanism. You should not remove them all.

wu-sheng: Meanwhile, I think from performance perspective, use the witness class directly should be quicker too.

honganan: > If isAssignableFrom executes every time, you will slow the whole process.

I've considered the need for a cache before, but then I decided that a class has a very limited number of superclasses, and this is a completely purely in-memory operation that should be very fast

wu-sheng: No, it doesn't. You could run benchmark, metadata access is not performance oriented.

honganan: > `static` block is an important mechanism. You should not remove them all.

The top part of the static block looks a bit redundant. Because when it's used, for example:
```java
if (IS_SERVLET_GET_STATUS_METHOD_EXIST && HttpServletResponse.class.isAssignableFrom(response.getClass()))
```
If response's class is a subclass of`HttpServletResponse`, it certainly have `getStatus()` method.

honganan: > Meanwhile, I think from performance perspective, use the witness class directly should be quicker too.

Using the witness class mechanism to recognize the SDK version is better and should be clearer to read. Since it's an abstract class used by multiple versions of plugins, we would need to do some refactoring to it. Is this consistent with your thoughts?

wu-sheng: I think for any plugin using jaxax and/or jakata, they need to deal with this kind of conflict, right?
It is better we have a unified way to do that.

honganan: I believe there should be no conflicts if each plugin defines an accurate witness class. Perhaps removing this cumbersome abstraction and pushing the logic down into each version of the plugins would look more natural. ^^

lujiajing1126:
After checking the issue itself, I don't think the scenario is a real case,

1. Normally, you should not use old servlet-api with Spring 6.x/Tomcat 10.x
2. `servlet-api` should be declared as `provided` as this is provided by the servlet container via runtime.

Pls correct me if I'm wrong

wu-sheng:
> Normally, you should not use old servlet-api together with Spring 6.x/Tomcat 10.x which is using Jakarta API,

Could we find some information about this? Such as whether Spring 6 or Tomcat 10 docs mentioned this?
As this could run, whether Spring and Tomcat embed compatible logic.

lujiajing1126:
> > Normally, you should not use old servlet-api together with Spring 6.x/Tomcat 10.x which is using Jakarta API,
> 
> Could we find some information about this? Such as whether Spring 6 or Tomcat 10 docs mentioned this? As this could run, whether Spring and Tomcat embed compatible logic.

From the official migration guide, quoted,

"Due to the Jakarta EE migration, make sure to upgrade to Tomcat 10, Jetty 11, or Undertow 2.2.19 with the undertow-servlet-jakarta artifact, alongside switching your javax.servlet imports to jakarta.servlet (Jakarta EE 9). For the latest server generations, consider Tomcat 10.1 and Undertow 2.3 (Jakarta EE 10).

Several outdated Servlet-based integrations have been dropped: e.g. Apache Commons FileUpload (org.springframework.web.multipart.commons.CommonsMultipartResolver), and Apache Tiles as well as FreeMarker JSP support in the corresponding org.springframework.web.servlet.view subpackages. We recommend org.springframework.web.multipart.support.StandardServletMultipartResolver for multipart file uploads and regular FreeMarker template views if needed, and a general focus on REST-oriented web architectures."

https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-6.x/10747c0695774ef4dd3c3bbe4f6d4dde96168b08

wu-sheng:
@honganan It seems we should skip this compatibility, as dependence itself should be fixed? WDYT?

We could add FAQ about this for sure.

honganan:
> @honganan It seems we should skip this compatibility, as dependence itself should be fixed? WDYT?
> 
> We could add FAQ about this for sure.

All of your discussions have been very professional, @lujiajing1126. From the actual situation I've encountered (we upgraded to JDK 17 and found this to be the case with more than 10 services), I don't want the system we provide to rely on users‘ exact correct usage. If a user over-references or indirectly references an unused dependency, the application itself may still work fine, but suddenly, Skywalking crashes.

Forgive me if my thoughts are a bit extreme. In any case, adding a FAQ will also be very helpful.

lujiajing1126:
> All of your discussions have been very professional, @lujiajing1126. From the actual situation I've encountered (we upgraded to JDK 17 and found this to be the case with more than 10 services), I don't want the system we provide to rely on users‘ exact correct usage. 

I agree that the Tomcat / Spring MVC plugins could provide a more accurate way to detect the runtime dependencies. But the current implementation in this PR can be hardly met the requirement from from a performance perspective.

> If a user over-references or indirectly references an unused dependency, the application itself may still work fine, but suddenly, Skywalking crashes.

Well, this is a good question. IMHO, you should run dependencies check before the artifact is going to be **promoted** to the production artifactory, e.g. Nexus Repository for Maven.

honganan:
> I highly doubt about "the application itself may still work fine".

We have many services running in production environment. And the easiest way to test is adding a `javax.servlet-api` dependency to 'Spring-6.x-scenario' ：
```xml
        <dependency>
            <groupId>javax.servlet</groupId>
            <artifactId>javax.servlet-api</artifactId>
            <version>3.1.0</version>
        </dependency>
```
and access in browser with: `http://localhost:8080/spring-6.x-scenario/case/resttemplate`


> But the current implementation in this PR can be hardly met the requirement from a performance perspective

About the performance, I think we can add a cache map with limited length here.

lujiajing1126:
> We have many services running in production environment. And the easiest way to test is adding a `javax.servlet-api` dependency to 'Spring-6.x-scenario' ：
> 
> ```
>         <dependency>
>             <groupId>javax.servlet</groupId>
>             <artifactId>javax.servlet-api</artifactId>
>             <version>3.1.0</version>
>         </dependency>
> ```
> 
> and access in browser with: `http://localhost:8080/spring-6.x-scenario/case/resttemplate`

I did not refer to this specific case. I meant if you add dependencies in an unrestricted way, it is highly possible that there will be dependency conflicts.

Since the E2E test targets only very "simple" and correct scenarios, users may actually face many incompatibilities. And do we have responsibilities to solve all those unreasonable use cases?

honganan:
Alright, then I'll close it. If anyone encounters the same issue, feel free to @ me.

PR #707, opened 2024-07-21
Feature/repairkafka3.7

darkness-2nd:

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/Kafka37ConsumerInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/Kafka37ConsumerInterceptor.java
@@ -20,8 +20,6 @@ package org.apache.skywalking.apm.plugin.kafka;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
-import org.apache.kafka.clients.consumer.internals.Fetch;
-import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.header.Header;
 import org.apache.skywalking.apm.agent.core.context.CarrierItem;
 import org.apache.skywalking.apm.agent.core.context.ContextCarrier;
@@ -35,10 +33,8 @@ import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
 import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
 import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
 
-public class Kafka3_7ConsumerInterceptor extends KafkaConsumerInterceptor {
+public class Kafka37ConsumerInterceptor extends KafkaConsumerInterceptor {
 
     @Override
     public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes, Object ret) throws Throwable {
--- a/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/KafkaConsumerInstrumentation.java
+++ b/apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/KafkaConsumerInstrumentation.java
@@ -50,6 +50,8 @@ public class KafkaConsumerInstrumentation extends AbstractKafkaInstrumentation {
     public static final String INTERCEPTOR_CLASS = "org.apache.skywalking.apm.plugin.kafka.KafkaConsumerInterceptor";
     public static final String INTERCEPTOR_CLASS_KAFKA3_2 = "org.apache.skywalking.apm.plugin.kafka.Kafka3ConsumerInterceptor";
     public static final String ENHANCE_METHOD = "pollOnce";
+    public static final String ENHANCE_METHOD_KAFKA3_7 = "poll";
+    public static final String INTERCEPTOR_CLASS_KAFKA3_7 = "org.apache.skywalking.apm.plugin.kafka.Kafka3_7ConsumerInterceptor";
     public static final String ENHANCE_COMPATIBLE_METHOD = "pollForFetches";
     public static final String ENHANCE_CLASS = "org.apache.kafka.clients.consumer.KafkaConsumer";
     public static final String SUBSCRIBE_METHOD = "subscribe";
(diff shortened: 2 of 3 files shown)


wu-sheng:
Please explain what is changing here, and if this is a plugin update, you need to update test scenarios too.

wu-sheng:
If the new use cases are same here, you should add more versions to support your new implementations for newer versions, and keep compatibility with previous.

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/KafkaConsumerInstrumentation.java
```diff
@@ -62,104 +64,121 @@ public class KafkaConsumerInstrumentation extends AbstractKafkaInstrumentation {
...
-                @Override
-                public ElementMatcher<MethodDescription> getConstructorMatcher() {
-                    return takesArgumentWithType(0, CONSTRUCTOR_INTERCEPT_TYPE);
-                }
+        return new ConstructorInterceptPoint[]{
+                new ConstructorInterceptPoint() {
+                    @Override
+                    public ElementMatcher<MethodDescription> getConstructorMatcher() {
+                        return takesArgumentWithType(0, CONSTRUCTOR_INTERCEPT_TYPE);
+                    }
```
wu-sheng, line 72: You should not reformat, thr code style file is in the source codes.

darkness-2nd:
> ase explain what is changing here, and if th



> If the new use cases are same here, you should add more versions to support your new implementations for newer versions, and keep compatibility with previous.

Ok, I'll commit later

wu-sheng:
About https://github.com/apache/skywalking-java/pull/708#issuecomment-2242063529 I mean the test versions

demo codes https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/kafka-scenario
tested versions https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/kafka-scenario/support-version.list

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/KafkaConsumerInstrumentation.java
darkness-2nd: > You should not reformat, thr code style file is in the source codes.

I reset the java file which I reformat

wu-sheng:
Please pick one for 3.4, 3.5, 3.6 as well.

darkness-2nd:
> Please pick one for 3.4, 3.5, 3.6 as well.

3.4, 3.5, 3.6 are same to 3.2.3

wu-sheng:
Logically same. But we should test.

darkness-2nd:
> Logically same. But we should test.

OK, Should I create modules kafka-3.6.x-scenario and kafka-3.7.x-scenario like kafka-2.3.x-scenario in dir /test?

wu-sheng:
> > Logically same. But we should test.
> 
> 
> 
> OK, Should I create modules kafka-3.6.x-scenario and kafka-3.7.x-scenario like kafka-2.3.x-scenario in dir /test?

Those exist because the test codes are different. If you just need to change dependency versions, you don't need to do that.

wu-sheng:
And please fix CIs.

wu-sheng:
You should verify all locally rather than costing too much on CI.

darkness-2nd:
> You should verify all locally rather than costing too much on CI.
Sry, it's my first time to create PR in github, I'm doing test CI in my private branch now

docs/en/setup/service-agent/java-agent/Bootstrap-plugins.md
```diff
@@ -19,4 +19,3 @@ The plugin of JDK HttpURLConnection depended on `sun.net.*`. When using Java 9+,
 
 For more information 
 1. [JEP 403: Strongly Encapsulate JDK Internals](https://openjdk.org/jeps/403)
-2. [A peek into Java 17: Encapsulating the Java runtime internals](https://blogs.oracle.com/javamagazine/post/a-peek-into-java-17-continuing-the-drive-to-encapsulate-the-java-runtime-internals)
```
wu-sheng, line 22: I think you made a mistake to remove this line?

wu-sheng:
> --- /home/runner/work/skywalking-java/skywalking-java/tools/plugin/md-javaagent-plugin-list.txt	2024-07-25 03:04:25.186078041 +0000
+++ /home/runner/work/skywalking-java/skywalking-java/tools/plugin/genernate-javaagent-plugin-list.txt	2024-0[7](https://github.com/apache/skywalking-java/actions/runs/10087411570/job/27891606111?pr=707#step:5:8)-25 03:04:25.178078015 +0000

I think you did not update docs accordingly, Plugin-list.md and Supported-list.md

docs/en/setup/service-agent/java-agent/Bootstrap-plugins.md
darkness-2nd: > I think you made a mistake to remove this line?

No, the CI thiks this line is a deadlink, though it can be access in browser

wu-sheng: The blog is there. What is the deadlink check saying?

wu-sheng: You should not remove the docs, just because a CI took says otherwise.

darkness-2nd: > You should not remove the docs, just because a CI took says otherwise.

OK, I will rollback it, thank you!

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/KafkaConsumerInstrumentation.java
```diff
@@ -63,103 +63,103 @@ public class KafkaConsumerInstrumentation extends AbstractKafkaInstrumentation {
...
+                    @Override
+                    public ElementMatcher<MethodDescription> getConstructorMatcher() {
+                        return takesArgumentWithType(0, CONSTRUCTOR_INTERCEPT_MAP_TYPE);
+                    }
+
+                    @Override
+                    public String getConstructorInterceptor() {
+                        return MAP_CONSTRUCTOR_INTERCEPTOR_CLASS;
+                    }
+                },
```
wu-sheng, line 87: Please revert re-format.

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/Kafka37ConsumerInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+ * When the kafka version < 3.7, the enhance method could be: org.apache.kafka.clients.consumer.KafkaConsumer#pollForFetches.
+ * But when the kafka version >= 3.7, the enhance method should be: org.apache.kafka.clients.consumer.KafkaConsumer#poll(java.time.Duration).
+ * And the return type was also changed, from org.apache.kafka.clients.consumer.internals.Fetch to org.apache.kafka.clients.consumer.ConsumerRecords.
+ */
+public class Kafka37ConsumerInterceptor extends KafkaConsumerInterceptor {
+
+    @Override
+    protected Map<TopicPartition, List<ConsumerRecord<?, ?>>> fetchRecords(Object retObj) {
+        Map<TopicPartition, List<ConsumerRecord<?, ?>>> rsp = new HashMap<>();
+        if (retObj instanceof ConsumerRecords) {
```
wu-sheng, line 41: Why `instanceof`? Is there another case to have another parameter? The parent class has `Map<TopicPartition, List<ConsumerRecord<?, ?>>>` type as the parameter, but it should not affect this new Kafka37ConsumerInterceptor, right?

darkness-2nd: This class should be removed, I missed it, I'll delete it at next commit

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/KafkaConsumerInstrumentation.java
darkness-2nd: > Please revert re-format.

I have already recerted it, this class dosen't change, I have already rollback it to the original version

darkness-2nd:
> > --- /home/runner/work/skywalking-java/skywalking-java/tools/plugin/md-javaagent-plugin-list.txt	2024-07-25 03:04:25.186078041 +0000
> > +++ /home/runner/work/skywalking-java/skywalking-java/tools/plugin/genernate-javaagent-plugin-list.txt	2024-0[7](https://github.com/apache/skywalking-java/actions/runs/10087411570/job/27891606111?pr=707#step:5:8)-25 03:04:25.178078015 +0000
> 
> I think you did not update docs accordingly, Plugin-list.md and Supported-list.md



> > --- /home/runner/work/skywalking-java/skywalking-java/tools/plugin/md-javaagent-plugin-list.txt	2024-07-25 03:04:25.186078041 +0000
> > +++ /home/runner/work/skywalking-java/skywalking-java/tools/plugin/genernate-javaagent-plugin-list.txt	2024-0[7](https://github.com/apache/skywalking-java/actions/runs/10087411570/job/27891606111?pr=707#step:5:8)-25 03:04:25.178078015 +0000
> 
> I think you did not update docs accordingly, Plugin-list.md and Supported-list.md



> And please fix CIs.



> And please fix CIs.



> And please fix CIs.

All the CIs were successfully executed but the DeadLink check

wu-sheng:
@kezhenxu94 As 403 respond code, should we add this into ignore list?

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/KafkaConsumerInstrumentation.java
lujiajing1126: > I have already recerted it, this class dosen't change, I have already rollback it to the original version

No, you didn't. Change is still there.

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/Kafka37AsyncConsumerInstrumentation.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin.kafka.define;
+
+import org.apache.skywalking.apm.agent.core.plugin.match.ClassMatch;
+import static org.apache.skywalking.apm.agent.core.plugin.match.NameMatch.byName;
+
+/**
```
lujiajing1126, line 24: For the comment, I think we can focus on the difference instead of repeating what is already written in the parent class.

darkness-2nd: > For the comment, I think we can focus on the difference instead of repeating what is already written in the parent class.

Yes, the difference is the method named pollForFetchs was removed from KafkaConsumer to another two classes, so the original interceptor can not intercept it. Because of the enhance class is changed, so I create two new classes to repair the uncompatible problem.

lujiajing1126: > Yes, the difference is the method named pollForFetchs was removed from KafkaConsumer to another two classes, so the original interceptor can not intercept it. Because of the enhance class is changed, so I create two new classes to repair the uncompatible problem.

I meant you can modify the comment for this class......

darkness-2nd: You are right, I forget to change the comment

apm-sniffer/apm-sdk-plugin/kafka-plugin/src/main/java/org/apache/skywalking/apm/plugin/kafka/define/KafkaConsumerInstrumentation.java
darkness-2nd: > > I have already recerted it, this class dosen't change, I have already rollback it to the original version
> 
> No, you didn't. Change is still there.

Those changes are compare to the changes which are not correct I commit before...

darkness-2nd: > > I have already recerted it, this class dosen't change, I have already rollback it to the original version
> 
> No, you didn't. Change is still there.

I dont't know why.... I check out the original branch and copy the codes into the class file, but the IDE says that there is no difference.....

wu-sheng (dismissed):

wu-sheng:
One more. changes.md should be updated as well.

darkness-2nd:
> One more. changes.md should be updated as well.
Changed

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -76,7 +76,7 @@ metrics based on the tracing data.
 * MQ
   * [RocketMQ](https://github.com/apache/rocketmq) 3.x-> 5.x
   * [RocketMQ-gRPC](http://github.com/apache/rocketmq-clients) 5.x
-  * [Kafka](http://kafka.apache.org) 0.11.0.0 -> 3.2.3
+  * [Kafka](http://kafka.apache.org) 0.11.0.0 -> 3.2.3 -> 3.7.1
```
wu-sheng, line 79:
```suggestion
  * [Kafka](http://kafka.apache.org) 0.11.0.0 -> 3.7.1
```

wu-sheng (approved):

PR #708, opened 2024-07-22
Feature/repair kafka3.7.x

darkness-2nd:


wu-sheng:
Why 3.7 plugin becomes a new plugin? Could you explain what you did?

darkness-2nd:
> Why 3.7 plugin becomes a new plugin? Could you explain what you did?

I did not get your "you should add more versions to support your new implementations for newer versions", you mean I should only add version into skywalking-plugin.def and do impl in module kafka-plugin?

wu-sheng:
I mean the test versions
- demo codes https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/kafka-scenario
- tested versions https://github.com/apache/skywalking-java/blob/main/test/plugin/scenarios/kafka-scenario/support-version.list

PR #711, opened 2024-07-27
Upgrade nats plugin to support 2.16.5

pg-yang:

Support nats 2.16.5.  
There is still no proper method and opportunity to change the message header and re-calculate the message length. From what I tested recently, up to 2.16.5 works correctly.


wu-sheng:
Please fix CI. You changed the plugin name.
And document(supported list) should be update.

More important, you need to add new version to test scenarios.

wu-sheng (approved):

PR #712, opened 2024-07-27
fix redission has not parameter problem

CodingMakeWordBetter:
when i use redission plugin,i found even agent.config plugin.redisson.trace_redis_parameters set to true but skywalking ui do not have parameter. so fix this problem.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/redisson-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/redisson/v3/RedisConnectionMethodInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/redisson-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/redisson/v3/RedisConnectionMethodInterceptor.java
@@ -140,6 +140,9 @@ public class RedisConnectionMethodInterceptor implements InstanceMethodsAroundIn
             return Optional.empty();
         }
         Object argument = allArguments[0];
+        if (argument instanceof byte[]) {
+            return Optional.of(new String((byte[]) argument));
+        }
         // include null
         if (!(argument instanceof String)) {
             return Optional.empty();


wu-sheng:
We had discussed this, I think. How can you sure the bytes could be encoded as string?

CodingMakeWordBetter:
Firstly, I packaged this code into a jar package for the Redis plugin and tested that Skywalking can accept commands normally in my environment. Secondly, my project uses Redis 3.15.6, while Redis 3.20.0 is used in the plugin. If the developer of the old plugin has no problem with this, then I believe that the element 0 of Object [] params in the CommandData object returned byte [] in the old version and String [] in the new version

wu-sheng:
That is what want to confirm, how could make sure the old client wouldn't be impacted if it is using bytes can't be encoded?

wu-sheng:
You can't say you don't have the problem, then other users won't.
Unless the official doc(redis) confirmed this is a string, we should not assume that.

CodingMakeWordBetter:
关于这个问题，以我个人能力，我只能去询问redission官方找寻答案，但是我认为目前情况属于有问题的，插件是3.x，但是对于3.15.6并没有通过测试且是可复现的，说明插件本身对于版本并没有兼容。

wu-sheng:
If there is higher version has string as parameters, we could consider separating two interceptors.

wu-sheng:
> Regarding this issue, based on my personal ability, I can only consult with the official Redis to find the answer. However, I believe that the current situation is problematic. The plugin is 3. x, but it did not pass the testing for 3.15.6 and is reproducible, indicating that the plugin itself is not compatible with the version.

That is not a surprise. It's common that newer versions are not supported. That is how iteration begins.

wu-sheng:
> Regarding this issue, based on my personal ability, I can only consult with the official Redis to find the answer. 

We don't have to know the answer. We just need to support versions with string as parameters, and keep the bytes as what it is. If I understand your point correctly, I think this should be enough.

CodingMakeWordBetter:
> If there is higher version has string as parameters, we could consider separating two interceptors.



> > Regarding this issue, based on my personal ability, I can only consult with the official Redis to find the answer. However, I believe that the current situation is problematic. The plugin is 3. x, but it did not pass the testing for 3.15.6 and is reproducible, indicating that the plugin itself is not compatible with the version.
> 
> That is not a surprise. It's common that newer versions are not supported. That is how iteration begins.

But we are not sure about the specific version number, so it feels a bit strange to use version to split. In the worst case scenario of this code, the necessary condition is that the user has turned on the report switch, and the version is a lower version, and the content of byte [] cannot be converted into a string like 3.15.6, resulting in garbled parameters being reported. If we do not consider compatibility with older versions of redistriction, then I will only do local repairs

wu-sheng:
> But we are not sure about the specific version number, so it feels a bit strange to use version to split.

No, we don't. SkyWalking used class name and/or method signatures to identify versions. We documented it clearly in plugin dev doc.
Please don't guess.

CodingMakeWordBetter:
> We don't have to know the answer. We just need to support versions with string as parameters, and keep the bytes as what it is. If I understand your point correctly, I think this should be enough.



> We don't have to know the answer. We just need to support versions with string as parameters, and keep the bytes as what it is. If I understand your point correctly, I think this should be enough.

No problem, that's because even if the command switch is set to true for the low version (byte [] version) of the authorization report, the reported command cannot be obtained

wu-sheng:
What do you mean switch to true?

CodingMakeWordBetter:
> What do you mean switch to true?

in the agent.config plugin.redisson.trace_redis_parameters set true,but skywalking did not revice redis command.

wu-sheng:
You could update config document to indicate this. bytes are not supported.

I don't think this is a big issue. Encoding unexpected data could impact the target app, that is a serious issue.

CodingMakeWordBetter:
> You could update config document to indicate this. bytes are not supported.
> 
> I don't think this is a big issue. Encoding unexpected data could impact the target app, that is a serious issue.



> You could update config document to indicate this. bytes are not supported.
> 
> I don't think this is a big issue. Encoding unexpected data could impact the target app, that is a serious issue.

Okay, I will use the local version to solve this problem. I will try upgrading from 3.15.6 to 3.20.0 and it will directly report an error

wu-sheng:
I think yoi didn't read the document and understand how the things really work.
What you faced is a super common case in the agent.

PR #714, opened 2024-08-12
bugfix for httpasyncclient

lu-xiaoshuang:
HttpHost info is missing in URL tag, if `public <T> Future<T> execute(final HttpAsyncRequestProducer requestProducer, ...` and/or `public Future<HttpResponse> execute(final HttpHost target, final HttpRequest request, ...` are/is used.

```
    closeableHttpAsyncClient.execute(
        HttpHost.create("http://address:port"),
        new HttpGet("/path"),
        new FutureCallbackImpl<HttpResponse>());
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Release Notes.
 9.4.0
 ------------------
 
+* bugfix for httpasyncclient
 * Upgrade nats plugin to support 2.16.5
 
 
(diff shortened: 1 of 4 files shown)


wu-sheng:
Could you update tesr scenario of http async client to add one call of using this way and add an expected span in the expected file?

wu-sheng:
Also, changelogs should be updated too. You didn't follow the templ of pull request.

wu-sheng:
I think you didn't follow plugin dev doc.

- https://skywalking.apache.org/docs/skywalking-java/next/en/setup/service-agent/java-agent/plugin-test/
- This should be async http plugin test, https://github.com/apache/skywalking-java/tree/main/test/plugin/scenarios/asynchttpclient-scenario

We didn't trust UTs, as most data are mocked, which means nothing to the agent use. We built and run real application to verify.

wu-sheng:
Close as no update accordingly.

PR #715, opened 2024-08-19
[plugin][pulsar] Handle the SendCallback method sign changed.

dao-jun:
In https://github.com/apache/pulsar/pull/22940, Pulsar will change the `org.apache.pulsar.client.impl.SendCallback#sendComplete` method sign from

```java
public void sendComplete(Exception e)
```
to

```java
public void sendComplete(Throwable t, OpSendMsgStats stats)
```

This PR is to address it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Release Notes.
 ------------------
 
 * Upgrade nats plugin to support 2.16.5
+* [pulsar] Handle the SendCallback class method signature change.
 
 
 All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/222?closed=1)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ Release Notes.
 ------------------
 
 * Upgrade nats plugin to support 2.16.5
-* [pulsar] Handle the SendCallback class method signature change.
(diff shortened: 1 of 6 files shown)


wu-sheng:
Please fix CI and we need to add the new versions to the test scenarios to verify.

Also, you need to make CI passed.

CHANGES.md
```diff
@@ -6,6 +6,7 @@ Release Notes.
 ------------------
 
 * Upgrade nats plugin to support 2.16.5
+* [pulsar] Handle the SendCallback class method signature change.
```
wu-sheng, line 9:
```suggestion
* Support newer Pulsar versions to adopt SendCallback class method signature change.
```

I can't see which versions will include this. The PR you refer to doesn't have the version information.

dao-jun: The code implementation PR: https://github.com/apache/pulsar/pull/23196.
It will available in the next release.

wu-sheng: We can't accept this PR until it is released and verified through CI tests.

We have to hold this until then.

wu-sheng:
Could you fix the CI first? Then we need to wait for Pulsar's next release.

wu-sheng:
No update for a month. Close for now.

PR #717, opened 2024-09-23
fix ClassCircularityError

lujiajing1126:
We are also suffering from this issue. But it is not possible to reproduce it.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #12572.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@ Release Notes.
 
 * Upgrade nats plugin to support 2.16.5
 * Add agent self-observability.
-
+* Update ByteBuddy to 1.15.1 and resolve the ClassCircularityError.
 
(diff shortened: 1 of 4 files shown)


lujiajing1126:
Since the discussion has not been updated for days. I would like to cherry-pick the commit from dd-trace-java.

The only issue for me is how to verify?

wu-sheng:
Please revert the unexpected changes.

wu-sheng:
> The only issue for me is how to verify?

There isn't an easy way to reproduce. Only in the environment it happened and waited.

lujiajing1126:
> Please revert the unexpected changes.

Fixed

wu-sheng:
Please recheck CI, or rerun? Others, I am good.

lujiajing1126:
#718 preferred

PR #720, opened 2024-09-29
Support async profiler feature

zhengziyi0117:
Integrate async profiler performance analysis function in  Java


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -252,6 +252,25 @@ public static class Profile {
...
+        /**
+         * Max monitor time(second), if async profiler monitor time out of limit, then stop it.
+         * default 1h.
+         */
+        public static int MAX_DURATION = 600;
+
+        /**
+         * jfr directory generated by async profiler
+         */
+        public static String OUTPUT_PATH = "";
```
lujiajing1126, line 271: It seems we don't have any default value for the output path

lujiajing1126 (changes requested):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
zhengziyi0117: This parameter is used when creating jfr files. If it is empty, Files.createTemp will be used. Otherwise, the file will be created in the specified directory.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -252,6 +252,25 @@ public static class Profile {
         public static int SNAPSHOT_TRANSPORT_BUFFER_SIZE = 500;
     }
 
+    public static class AsyncProfiler {
+        /**
+         * If true, skywalking agent will enable profile when user create a new async profiler task.
```
lujiajing1126, line 257:
```suggestion
         * If true, Async Profiler will be enabled when user creates a new async profiler task.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -252,6 +252,25 @@ public static class Profile {
...
+         * Otherwise disable it.
```
lujiajing1126, line 258:
```suggestion
         * Otherwise it is disabled.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -252,6 +252,25 @@ public static class Profile {
...
+         */
+        public static boolean ACTIVE = true;
+
+        /**
+         * Max monitor time(second), if async profiler monitor time out of limit, then stop it.
```
lujiajing1126, line 263:
```suggestion
         * Max execution time(second) for the Async Profiler. The task will be stopped even if a longer time is specified.
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/conf/Config.java
```diff
@@ -252,6 +252,25 @@ public static class Profile {
...
+         * default 1h.
+         */
+        public static int MAX_DURATION = 600;
+
+        /**
+         * jfr directory generated by async profiler
```
lujiajing1126, line 269:
```suggestion
         * Path for the JFR outputs from the Async profiler
```

apm-sniffer/config/agent.config
```diff
@@ -164,6 +164,12 @@ profile.duration=${SW_AGENT_PROFILE_DURATION:10}
 profile.dump_max_stack_depth=${SW_AGENT_PROFILE_DUMP_MAX_STACK_DEPTH:500}
 # Snapshot transport to backend buffer size
 profile.snapshot_transport_buffer_size=${SW_AGENT_PROFILE_SNAPSHOT_TRANSPORT_BUFFER_SIZE:4500}
+# If true, skywalking agent will enable profile when user create a new asyncprofile task. Otherwise disable it.
```
lujiajing1126, line 167: Same suggestion for the configurations

lujiajing1126 (changes requested):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTask.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * execArgument from oap server
+     */
+    private String execArgs;
+    /**
+     * run profiling for duration seconds
+     */
+    private int duration;
+    /**
+     * run profiling for duration seconds
```
lujiajing1126, line 49: The same meaning?

lujiajing1126 (changes requested):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
```diff
@@ -0,0 +1,185 @@
...
+        AsyncProfilerMetaData metaData = AsyncProfilerMetaData.newBuilder()
+                .setService(Config.Agent.SERVICE_NAME)
+                .setServiceInstance(Config.Agent.INSTANCE_NAME)
+                .setType(AsyncProfilerCollectType.PROFILING_SUCCESS)
+                .setContentSize(size)
+                .setTaskId(task.getTaskId())
+                .build();
+        AsyncProfilerData asyncProfilerData = AsyncProfilerData.newBuilder().setMetaData(metaData).build();
+        dataStreamObserver.onNext(asyncProfilerData);
+        // todo wait for server ?
```
lujiajing1126, line 130: I suppose we have reached agreement how the Agent and OAP should negotiate by the first communication

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
```diff
@@ -0,0 +1,185 @@
...
+                    LOGGER.error(
+                            t, "Send async profiler task data to collector fail with a grpc internal exception."
+                    );
+                }
+                ServiceManager.INSTANCE.findService(GRPCChannelManager.class).reportError(t);
+            }
+
+            @Override
+            public void onCompleted() {
+                status.finished();
```
lujiajing1126, line 118: Shall we delete the file if upload has succeeded

lujiajing1126 (changes requested):

lujiajing1126:
Pls fix CI

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
```diff
@@ -0,0 +1,190 @@
...
+                .setContentSize(size)
+                .setTaskId(task.getTaskId())
+                .build();
+        AsyncProfilerData asyncProfilerData = AsyncProfilerData.newBuilder().setMetaData(metaData).build();
+        dataStreamObserver.onNext(asyncProfilerData);
+        /**
+         * Wait briefly to see if the server can receive the jfr. If the server cannot receive it, onError will be triggered.
+         * Then we wait for a while (waiting for the server to send onError) and then decide whether to send the jfr file.
+         */
+        Thread.sleep(500);
```
lujiajing1126, line 135: @wu-sheng In the current design, the Java Agent has to wait after the first request in order to determine whether the OAP accepts the following data transfer. This is a kind of negotiation.

WDYT?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
```diff
@@ -0,0 +1,190 @@
...
+
+        // Is it possible to upload jfr?
```
wu-sheng, line 137: What does this comment mean?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
wu-sheng: `500ms` can't guarantee anything AFAIK.

pom.xml
```diff
@@ -213,6 +214,11 @@
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
+                <groupId>tools.profiler</groupId>
+                <artifactId>async-profiler</artifactId>
+                <version>${async-profiler.version}</version>
+            </dependency>
```
wu-sheng, line 221: Please move this the last one rather than the first.

pom.xml
```diff
@@ -87,6 +87,7 @@
 
         <!-- core lib dependency -->
         <bytebuddy.version>1.14.9</bytebuddy.version>
+        <async-profiler.version>3.0</async-profiler.version>
```
wu-sheng, line 90: Please move this the last one rather than the first.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
wu-sheng: Note, in this case, Command should not be the response anymore.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
```diff
@@ -0,0 +1,194 @@
...
+            ClientCallStreamObserver<AsyncProfilerData> requestStream;
+
+            @Override
+            public void beforeStart(ClientCallStreamObserver<AsyncProfilerData> requestStream) {
+                this.requestStream = requestStream;
+            }
+
+            @Override
+            public void onNext(AsyncProfilerCollectionResponse value) {
+                if (!AsyncProfilingStatus.TERMINATED_BY_OVERSIZE.equals(value.getType())) {
```
lujiajing1126, line 110: I suppose positive check is better here. For example,

```java
switch (value.getType()) {
case AsyncProfilingStatus.TERMINATED_BY_OVERSIZE:
    // ???
    break;
default:
    // ...
```

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTask.java
```diff
@@ -0,0 +1,146 @@
...
+    private long createTime;
+    /**
+     * tempFile generated by async-profiler execution
+     */
+    private Path tempFile;
+
+    private static String execute(AsyncProfiler asyncProfiler, String args)
+            throws IllegalArgumentException, IOException {
+        LOGGER.info("async profiler execute args:{}", args);
+        String result = asyncProfiler.execute(args);
```
lujiajing1126, line 61: What is the meaning of this return val?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTaskExecutionService.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    public void onComplete() throws Throwable {
+
+    }
+
+    @Override
+    public void shutdown() throws Throwable {
+        ASYNC_PROFILER_EXECUTOR.shutdown();
```
lujiajing1126, line 136: Shall we cancel the future before shutdown?

lujiajing1126 (changes requested):

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTask.java
zhengziyi0117: This return value is used by async-profiler to indicate whether the execution is successful. For example, the start command will normally return "Profiling start". For details, see https://github.com/async-profiler/async-profiler/blob/116504c9f75721911b2f561e29eda065c224caf6/src/profiler.cpp#L1695

lujiajing1126: So why we have to trim the result? Is it possible that the result is NULL?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTaskExecutionService.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public void onComplete() throws Throwable {
+
+    }
+
+    @Override
+    public void shutdown() throws Throwable {
+        scheduledFuture.cancel(true);
```
lujiajing1126, line 136: you have to check NULL!

lujiajing1126 (changes requested):
LGTM if all the current comments are resolved

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTask.java
zhengziyi0117: From the result of his method, I don't think it will return null, but I think "String.trim" is not necessary, so I updated it.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTaskChannelService.java
```diff
@@ -0,0 +1,115 @@
...
+                    .findService(AsyncProfilerTaskExecutionService.class).getLastCommandCreateTime();
+
+            AsyncProfilerTaskCommandQuery query = AsyncProfilerTaskCommandQuery.newBuilder()
+                    .setServiceInstance(Config.Agent.INSTANCE_NAME)
+                    .setService(Config.Agent.SERVICE_NAME)
+                    .setLastCommandTime(lastCommandCreateTime)
+                    .build();
+            Commands commands = asyncProfilerTaskBlockingStub.withDeadlineAfter(GRPC_UPSTREAM_TIMEOUT, TimeUnit.SECONDS)
+                    .getAsyncProfilerTaskCommands(query);
+            ServiceManager.INSTANCE.findService(CommandService.class).receiveCommand(commands);
```
wu-sheng, line 69: You should check `service not implemented` error of gRPC, you could try that by connecting v < 10.1 OAP.

wu-sheng: Once it is not implemented, the agent should shutdown this part automatically.

apm-sniffer/config/agent.config
```diff
@@ -164,6 +164,12 @@ profile.duration=${SW_AGENT_PROFILE_DURATION:10}
 profile.dump_max_stack_depth=${SW_AGENT_PROFILE_DUMP_MAX_STACK_DEPTH:500}
 # Snapshot transport to backend buffer size
 profile.snapshot_transport_buffer_size=${SW_AGENT_PROFILE_SNAPSHOT_TRANSPORT_BUFFER_SIZE:4500}
+# If true, Async Profiler will be enabled when user creates a new async profiler task, Otherwise it is disabled.
```
wu-sheng, line 167: This comment is not accurate. You hard coded this value as true, so, the default value is TRUE, not disabled. Please update your default value in the codes.

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/command/AsyncProfilerTaskCommand.java
```diff
@@ -0,0 +1,113 @@
...
+
+package org.apache.skywalking.apm.network.trace.component.command;
+
+import org.apache.skywalking.apm.network.common.v3.Command;
+import org.apache.skywalking.apm.network.common.v3.KeyStringValuePair;
+
+import java.util.List;
+import java.util.Objects;
+
+public class AsyncProfilerTaskCommand extends BaseCommand implements Serializable, Deserializable<AsyncProfilerTaskCommand> {
```
wu-sheng, line 27: Please explain the command format in the code comments.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
```diff
@@ -0,0 +1,194 @@
...
+import java.nio.channels.FileChannel;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+
+import static org.apache.skywalking.apm.agent.core.conf.Config.Collector.GRPC_UPSTREAM_TIMEOUT;
+
+@DefaultImplementor
+public class AsyncProfilerDataSender implements BootService, GRPCChannelListener {
+    private static final ILog LOGGER = LogManager.getLogger(ProfileSnapshotSender.class);
+    private static final int DATA_CHUNK_SIZE = 1024 * 1024;
```
wu-sheng, line 54: This should move into config. Don't have to expose through agent.config, but we don't need to hard code this.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
```diff
@@ -0,0 +1,194 @@
...
+                GRPC_UPSTREAM_TIMEOUT, TimeUnit.SECONDS
+        ).collect(new StreamObserver<AsyncProfilerCollectionResponse>() {
+            @Override
+            public void onNext(AsyncProfilerCollectionResponse value) {
+            }
+
+            @Override
+            public void onError(Throwable t) {
+                status.finished();
+                ServiceManager.INSTANCE.findService(GRPCChannelManager.class).reportError(t);
```
wu-sheng, line 171: When `onError`, we should output logs.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerDataSender.java
```diff
@@ -0,0 +1,194 @@
...
+            public void onCompleted() {
+                status.finished();
+            }
+        });
+        AsyncProfilerMetaData metaData = AsyncProfilerMetaData.newBuilder()
+                .setService(Config.Agent.SERVICE_NAME)
+                .setServiceInstance(Config.Agent.INSTANCE_NAME)
+                .setTaskId(task.getTaskId())
+                .setType(AsyncProfilingStatus.EXECUTION_TASK_ERROR)
+                .setContentSize(0)
```
wu-sheng, line 184:
```suggestion
                .setContentSize(-1)
```

Usually negative value indicates null for length.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTaskExecutionService.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+@DefaultImplementor
+public class AsyncProfilerTaskExecutionService implements BootService {
+
+    private static final ILog LOGGER = LogManager.getLogger(AsyncProfilerTaskChannelService.class);
+
+    private static final AsyncProfiler ASYNC_PROFILER = AsyncProfiler.getInstance();
```
wu-sheng, line 44: Please use lazy initialization as much as possible. I am not sure how much effects would this getInstance takes, but as service list is loaded before agent instrumentation, we should load as less code as possible. Otherwise, the instrumentation could fail due to classes have been loaded.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTaskExecutionService.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private static final String SUCCESS_RESULT = "Profiling started\n";
+
+    // profile executor thread pool, only running one thread
+    private static final ScheduledExecutorService ASYNC_PROFILER_EXECUTOR = Executors.newSingleThreadScheduledExecutor(
```
wu-sheng, line 49: Same here, please initialize later in the process.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTaskExecutionService.java
```diff
@@ -0,0 +1,160 @@
...
+                        () -> stopWhenSuccess(task), task.getDuration(), TimeUnit.SECONDS
+                );
+            } catch (IOException e) {
+                LOGGER.error("AsyncProfilerTask executor error:" + e.getMessage(), e);
+            }
+        });
+    }
+
+    private void stopWhenError(AsyncProfilerTask task, String errorMessage) {
+        LOGGER.error("AsyncProfilerTask start fail result:" + errorMessage);
```
wu-sheng, line 87:
```suggestion
        LOGGER.error("AsyncProfilerTask fails to start: {}"， errorMessage);
```

Correcting the sentence and use log text formater.

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTaskExecutionService.java
```diff
@@ -0,0 +1,160 @@
...
+            try (FileInputStream fileInputStream = new FileInputStream(dumpFile)) {
+                // upload file
+                FileChannel channel = fileInputStream.getChannel();
+
+                AsyncProfilerDataSender dataSender = ServiceManager.INSTANCE.findService(AsyncProfilerDataSender.class);
+                dataSender.sendData(task, channel);
+            }
+
+            if (!dumpFile.delete()) {
+                LOGGER.warn("delete async profiler dump file failed");
```
wu-sheng, line 106:
```suggestion
                LOGGER.warn("Fail to delete the dump file of async profiler.");
```

lujiajing1126 (approved):
LGTM

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/asyncprofiler/AsyncProfilerTaskChannelService.java
```diff
@@ -0,0 +1,131 @@
...
+                Commands commands = asyncProfilerTaskBlockingStub.withDeadlineAfter(GRPC_UPSTREAM_TIMEOUT, TimeUnit.SECONDS)
+                        .getAsyncProfilerTaskCommands(query);
+                ServiceManager.INSTANCE.findService(CommandService.class).receiveCommand(commands);
+            } catch (Throwable t) {
+                if (!(t instanceof StatusRuntimeException)) {
+                    LOGGER.error(t, "fail to query async-profiler task from backend");
+                    return;
+                }
+                final StatusRuntimeException statusRuntimeException = (StatusRuntimeException) t;
+                if (statusRuntimeException.getStatus().getCode() == Status.Code.UNIMPLEMENTED) {
```
wu-sheng, line 79: Please use `Status.Code.UNIMPLEMENTED.equals`. It is safer.

wu-sheng (approved):

PR #727, opened 2024-11-05
fix: shade async-profiler packages

zhengziyi0117:
- [ ] Associated with issue https://github.com/apache/skywalking/issues/12748

### DIFF
--- a/apm-sniffer/apm-agent-core/pom.xml
+++ b/apm-sniffer/apm-agent-core/pom.xml
@@ -53,6 +53,8 @@
         <shade.org.slf4j.target>${shade.package}.${shade.org.slf4j.source}</shade.org.slf4j.target>
         <ststem-rules.version>1.18.0</ststem-rules.version>
         <slf4j.version>1.7.25</slf4j.version>
+        <shade.one.profiler.source>one.profiler</shade.one.profiler.source>
+        <shade.one.profiler.target>${shade.package}.${shade.one.profiler.source}</shade.one.profiler.target>
     </properties>
 
     <dependencies>
@@ -226,6 +228,7 @@
                                     <exclude>org.checkerframework:checker-compat-qual:jar:</exclude>
                                     <exclude>org.checkerframework:checker-qual:jar:</exclude>
                                     <exclude>org.codehaus.mojo:animal-sniffer-annotations:jar:</exclude>
+<!--                                    <exclude>tools.profiler:async-profiler:jar:</exclude>-->
                                 </excludes>
                             </artifactSet>
                             <relocations>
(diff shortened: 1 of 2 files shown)


wu-sheng:
Could you post the screenshot of new binary package here?

And, have you tested the feature again? Shading sometimes breaks the function.

zhengziyi0117:
> Could you post the screenshot of new binary package here?你能在这里发布新的二进制包的屏幕截图吗？
> 
> And, have you tested the feature again? Shading sometimes breaks the function.而且，您再次测试过该功能吗？阴影有时会破坏功能。

I have tested it locally and the original function is normal. 

The following is the package structure diagram: 
image

wu-sheng:
About the so files, are they able to move as well?

zhengziyi0117:
> About the so files, are they able to move as well?

The .so file has not been moved yet. 

I think it should not be possible to move the .so file because the async-profiler code gets the dynamic library path based on the following code 
`String resourceName = "/" + getPlatformTag() + "/libasyncProfiler.so";`

wu-sheng (dismissed):

lujiajing1126 (commented):
It is a solution as long as you understand that **shading in the agent package instead of agent-core will make it impossible to be used from other packages** like plugins

wu-sheng:
> It is a solution as long as you understand that **shading in the agent package instead of agent-core will make it impossible to be used from other packages** like plugins

Do you mean you prefer to move this(shade) into core module? I think plugins would not access this part of codes.

lujiajing1126:
> > It is a solution as long as you understand that **shading in the agent package instead of agent-core will make it impossible to be used from other packages** like plugins
> 
> Do you mean you prefer to move this(shade) into core module? I think plugins would not access this part of codes.

I am OK with this one.

lujiajing1126 (approved):

wu-sheng:
@zhengziyi0117 Could you try that?
I think we don't need async profilor accessible from.plugins.

apm-sniffer/apm-agent-core/pom.xml
```diff
@@ -226,6 +228,7 @@
                                     <exclude>org.checkerframework:checker-compat-qual:jar:</exclude>
                                     <exclude>org.checkerframework:checker-qual:jar:</exclude>
                                     <exclude>org.codehaus.mojo:animal-sniffer-annotations:jar:</exclude>
+<!--                                    <exclude>tools.profiler:async-profiler:jar:</exclude>-->
```
wu-sheng, line 231: I think you should have this?

wu-sheng (approved):

wu-sheng (approved):

PR #731, opened 2024-11-28
Upgrade Kafka clients dependency version to 3.8.0

Super-Lu:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/12800.

### DIFF
--- a/apm-sniffer/optional-reporter-plugins/pom.xml
+++ b/apm-sniffer/optional-reporter-plugins/pom.xml
@@ -42,7 +42,7 @@
         <ant-contrib.version>1.0b3</ant-contrib.version>
         <ant-nodeps.version>1.8.1</ant-nodeps.version>
 
-        <kafka-clients.version>2.4.1</kafka-clients.version>
+        <kafka-clients.version>3.8.0</kafka-clients.version>
         <spring-kafka-test.version>2.4.6.RELEASE</spring-kafka-test.version>
     </properties>
 


wu-sheng:
As discussed, I don't think updating this works. Have you tested with a Kafka cluster and OAP?

Super-Lu:
> As discussed, I don't think updating this works. Have you tested with a Kafka cluster and OAP?

Yes, I have already tested the Kafka client version 3.8.0 locally to connect to the Kafka cluster and report link data to OAP. Currently, the data reporting to OAP is normal.

wu-sheng:
Let's see how the e2e says.

wu-sheng:
If all are working, remember to update the change log accordingly.

wu-sheng:
image

The e2e reports failure. Please recheck this e2e.

wu-sheng:
Maybe because of Kafka server version? Or something?

Super-Lu:
> Maybe because of Kafka server version? Or something?



> image
> The e2e reports failure. Please recheck this e2e.
Okay, I'll check it out.

lujiajing1126:
Starting from Kafka Clients 3.0, acks
 is changed from 1 to all, and idempotent is enabled by default.

It will lead to increasing latency. I suppose it is not necessary for telemetry cases.

lujiajing1126:
> Maybe we could change that configuration? Can we?
> 
> Personally, the issue is we can't stay in an old and not-releasing version, as suffering CVEs.

I think we can keep the current configurations by setting default properites

Super-Lu:
> > Maybe we could change that configuration? Can we?
> > Personally, the issue is we can't stay in an old and not-releasing version, as suffering CVEs.
> 
> I think we can keep the current configurations by setting default properites

 Could you please tell me what modifications I need to make in order to pass the test cases, and then I can merge my code into the main repository?

wu-sheng:
Any update? we should make the tests passed, then we talk about default configurations.

Super-Lu:
> Any update? we should make the tests passed, then we talk about default configurations.

Sorry to bother you, I need some help. How can I run test cases locally? Are there any guide documents available?

wu-sheng:
Which part? e2e? You could follow this, https://skywalking.apache.org/docs/skywalking-infra-e2e/latest/en/setup/run-e2e-tests/

Super-Lu:
> Which part? e2e? You could follow this, https://skywalking.apache.org/docs/skywalking-infra-e2e/latest/en/setup/run-e2e-tests/

Thanks, this is the guide document I needed.

PR #732, opened 2024-11-29
Add empty judgment for constructorInterceptPoint

CodePrometheus:

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
@@ -112,17 +112,20 @@ public abstract class ClassEnhancePluginDefine extends AbstractClassEnhancePlugi
          */
         if (existedConstructorInterceptPoint) {
             for (ConstructorInterceptPoint constructorInterceptPoint : constructorInterceptPoints) {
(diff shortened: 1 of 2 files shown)


apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
```diff
@@ -112,17 +112,20 @@ protected DynamicType.Builder<?> enhanceInstance(TypeDescription typeDescription
          */
         if (existedConstructorInterceptPoint) {
             for (ConstructorInterceptPoint constructorInterceptPoint : constructorInterceptPoints) {
+                String constructorInterceptor = constructorInterceptPoint.getConstructorInterceptor();
+                if (StringUtil.isEmpty(constructorInterceptor)) {
+                    throw new EnhanceException("no InstanceConstructorInterceptor define to enhance class " + enhanceOriginClassName);
+                }
```
wu-sheng, line 118: Someone added an intercept point but no interceptor? What happens if not check? It will still be an error.

CodePrometheus: If ConstructorInterceptor is empty but there is no check, no error message during runtime, but the expected data cannot be obtained because it is not intercepted, so I want to add a check here to inform the user.

wu-sheng (approved):

PR #734, opened 2024-12-04
fix: rm tomcat-10x-plugin version

fengyouliang:
tomcat-10x-plugin version incorrect
image


wu-sheng (approved):

fengyouliang:
"Error: The operation was canceled." 

Hi @wu-sheng  This looks like it timed out. What could be the reason for this? And can this check be triggered again?

wu-sheng:
This is merged. That error should be CI network issue.

PR #739, opened 2024-12-14
Fix CreateAopProxyInterceptor in the Spring core-patch indeed changes the implementation of the Spring AOP proxy

youjie23:
Fix https://github.com/apache/skywalking/issues/12858 The `CreateAopProxyInterceptor` in the Spring core-patch indeed changes the implementation of the Spring AOP proxy
- [X] Add a unit test to verify that the fix works.
- [X] Explain briefly why the bug exists and how to fix it.
The current version of `CreateAopProxyInterceptor` does not account for scenarios where the proxy target class implements `UserSuppliedInterface` and should work correctly with only `JdkDynamicAopProxy`. This can lead to inconsistencies in behavior when the SkyWalking agent is used.

This PR fixes the issue by adding restrictions on when the behavior changes, ensuring that the proxy mechanism reverts to using `JdkDynamicAopProxy` when appropriate. Specifically, it ensures that the application's behavior remains consistent both with and without the SkyWalking agent enhancement.

### DIFF
--- a/apm-sniffer/apm-sdk-plugin/spring-plugins/core-patch/src/main/java/org/apache/skywalking/apm/plugin/spring/patch/CreateAopProxyInterceptor.java
+++ b/apm-sniffer/apm-sdk-plugin/spring-plugins/core-patch/src/main/java/org/apache/skywalking/apm/plugin/spring/patch/CreateAopProxyInterceptor.java
@@ -18,11 +18,13 @@
 
 package org.apache.skywalking.apm.plugin.spring.patch;
 
-import java.lang.reflect.Method;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.springframework.aop.SpringProxy;
 import org.springframework.aop.framework.AdvisedSupport;
+
+import java.lang.reflect.Method;
 
 /**
  * <code>CreateAopProxyInterceptor</code> check that the bean has been implement {@link EnhancedInstance}.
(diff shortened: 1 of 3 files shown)


wu-sheng:
You missed the update of changes.md. Please fix it if all tests can pass.

apm-sniffer/apm-sdk-plugin/spring-plugins/core-patch/src/main/java/org/apache/skywalking/apm/plugin/spring/patch/CreateAopProxyInterceptor.java
```diff
@@ -32,25 +34,45 @@ public class CreateAopProxyInterceptor implements InstanceMethodsAroundIntercept
...
     }
 
+    private boolean maybeHasUserSuppliedProxyInterfaces(Object ret) {
+        return !(Boolean) ret;
+    }
+
+    private boolean onlyImplementsEnhancedInstanceAndSpringProxy(AdvisedSupport advisedSupport) {
+        Class<?>[] ifcs = advisedSupport.getProxiedInterfaces();
+        Class targetClass = advisedSupport.getTargetClass();
+        return ifcs.length == 2 && EnhancedInstance.class.isAssignableFrom(targetClass) && SpringProxy.class.isAssignableFrom(targetClass);
```
wu-sheng, line 64: When is `SpringProxy.class` added into the hierarchy tree?

youjie23: Thank you for the review.
As shown below, the `hasNoUserSuppliedProxyInterfaces` method in `DefaultAopProxyFactory` is defined as follows:
**spring4/spring5**
```
         /**
	 * Determine whether the supplied {@link AdvisedSupport} has only the
	 * {@link org.springframework.aop.SpringProxy} interface specified
	 * (or no proxy interfaces specified at all).
	 */
	private boolean hasNoUserSuppliedProxyInterfaces(AdvisedSupport config) {
		Class<?>[] ifcs = config.getProxiedInterfaces();
		return (ifcs.length == 0 || (ifcs.length == 1 && SpringProxy.class.isAssignableFrom(ifcs[0])));
	}
```
**spring3**
```
         /**
	 * Determine whether the supplied {@link AdvisedSupport} has only the
	 * {@link org.springframework.aop.SpringProxy} interface specified
	 * (or no proxy interfaces specified at all).
	 */
	private boolean hasNoUserSuppliedProxyInterfaces(AdvisedSupport config) {
		Class[] interfaces = config.getProxiedInterfaces();
		return (interfaces.length == 0 || (interfaces.length == 1 && SpringProxy.class.equals(interfaces[0])));
	}
```

wu-sheng: OK, this seems ok. Could you update the changelog? Then I could merge this PR.

youjie23: Thank you for the review. 
The `Changes.md` file has been updated.

wu-sheng (approved):
LGTM. Wait for CI passed.

apm-sniffer/apm-sdk-plugin/spring-plugins/core-patch/src/test/java/org/apache/skywalking/apm/plugin/spring/patch/CreateAopProxyInterceptorTest.java
```diff
@@ -48,18 +49,69 @@ public void setUp() {
...
+        // doReturn(MockClassImplementsSpringProxy.class).when(advisedSupport).getTargetClass();
+        // doReturn(MockClassImplementsSpringProxy.class.getInterfaces()).when(advisedSupport).getProxiedInterfaces();
+        assertThat(true, is(interceptor.afterMethod(enhancedInstance, null, new Object[] {advisedSupport}, new Class[] {Object.class}, true)));
+    }
+
+    @Test
+    public void testInterceptClassImplementsEnhancedInstance() throws Throwable {
+        doReturn(MockClassImplementsEnhancedInstance.class).when(advisedSupport).getTargetClass();
+        doReturn(MockClassImplementsEnhancedInstance.class.getInterfaces()).when(advisedSupport).getProxiedInterfaces();
         assertThat(true, is(interceptor.afterMethod(enhancedInstance, null, new Object[] {advisedSupport}, new Class[] {Object.class}, false)));
```
lujiajing1126, line 76: https://stackoverflow.com/questions/45124660/the-bean-could-not-be-injected-as-a-type-because-it-is-a-jdk-dynamic-proxy-tha

According to the answer in this SO thread, we basically should not expect classes that both use `@Transactional` annotation and being enhanced by SkyWalking Agent to use JDK Proxy.

We are facing startup failures after upgrading to SkyWalking 9.4.

PR #743, opened 2024-12-28
Add Caffeine plugin as optional

CodePrometheus:
- Bump up github/codeql-action/init & github/codeql-action/analyze to v3
- Fix SW_PLUGIN_GUAVACACHE_OPERATION_MAPPING_WRITE/READ contents reversed in configurations.md
- Add Caffeine plugin as optional


wu-sheng:
As you are adding a new component ID, please add it to the main repo and reserve it.

.github/workflows/plugins-jdk17-test.1.yaml
```diff
@@ -63,6 +63,7 @@ jobs:
           - c3p0-0.9.0.x-0.9.1.x-scenario
           - c3p0-0.9.2.x-0.10.x-scenario
           - spring-scheduled-6.x-scenario
+          - caffeine-3.x-scenario
```
wu-sheng, line 66: Is there any reason running this in JDK17?

apm-sniffer/optional-plugins/caffeine-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/caffeine/v3/CaffeineInterceptor.java
```diff
@@ -0,0 +1,86 @@
...
+        span.setComponent(ComponentsDefine.CAFFEINE);
+        if (allArguments != null && allArguments.length > 0 && allArguments[0] != null) {
+            if (allArguments[0] instanceof String) {
+                Tags.CACHE_KEY.set(span, allArguments[0].toString());
+            } else if (allArguments[0] instanceof Map) {
+                String keys = ((Map<?, ?>) allArguments[0])
+                    .keySet().stream().map(String::valueOf)
+                    .collect(Collectors.joining(","));
+                Tags.CACHE_KEY.set(span, keys);
+            } else if (allArguments[0] instanceof Set) {
```
wu-sheng, line 55: We should consider multiple interceptors to map different parameter type cases. It provides better performance.

test/plugin/scenarios/caffeine-3.x-scenario/src/main/java/org/apache/skywalking/apm/testcase/caffeine/controller/CaffeineController.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class CaffeineController {
+
+    @GetMapping("/case/caffeine")
+    public void testCase() {
+        try {
+            Cache<Object, Object> caffeine = Caffeine.newBuilder().build();
```
wu-sheng, line 34: As a local memory cache, shouldn't this be initialized once only?

.github/workflows/plugins-jdk17-test.1.yaml
CodePrometheus: Because I chose to use springboot3.x, no other special reason.

apm-sniffer/optional-plugins/caffeine-3.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/caffeine/v3/CaffeineIterableInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+public class CaffeineIterableInterceptor extends AbstractCaffeineInterceptor {
+
+    @Override
+    public void beforeMethod(final EnhancedInstance objInst,
+                             final Method method,
+                             final Object[] allArguments,
+                             final Class<?>[] argumentsTypes,
+                             final MethodInterceptResult result) throws Throwable {
+        AbstractSpan span = generateSpanInfo(method.getName());
+        if (allArguments != null && allArguments.length > 0) {
```
wu-sheng, line 38: Is this necessary? In which case, `allArguments` is unexpected?

CodePrometheus: I prefer to add, not sure all inputs meet the expectations before this function is processed by upstream.

wu-sheng (approved):

wu-sheng (approved):

PR #748, opened 2025-01-17
Fix ConsumeDriver running status

JoeCqupt:

### DIFF
--- a/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumeDriver.java
+++ b/apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumeDriver.java
@@ -88,7 +88,7 @@ public class ConsumeDriver<T> implements IDriver {
         }
         lock.lock();
         try {
-            if (running){
+            if (running) {
                 return;
             }
             this.allocateBuffer2Thread();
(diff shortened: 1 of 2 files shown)


wu-sheng:
Could you explain why do this change?

apm-commons/apm-datacarrier/src/main/java/org/apache/skywalking/apm/commons/datacarrier/consumer/ConsumeDriver.java
```diff
@@ -88,6 +88,9 @@ public void begin(Channels channels) {
         }
         lock.lock();
         try {
+            if (running){
```
JoeCqupt, line 91: Double Checked Locking

JoeCqupt:
ConsumeDriver use a `lock`  + `running ` status  to protect  `begin` method.  but there is a problem with code. it should use `Double Checked Locking`

wu-sheng:
> ConsumeDriver use a `lock`  + `running ` status  to protect  `begin` method.  but there is a problem with code. it should use `Double Checked Locking` 

Did you face some issues? Could you paste the logs here?
Or just from code review perspective?

wu-sheng:
You should update changes.md and describe what is fixed.

wu-sheng:
> You should update changes.md and describe what is fixed.

CI seems strange. Please update this and push again.

JoeCqupt:
> > ConsumeDriver use a `lock`  + `running ` status  to protect  `begin` method.  but there is a problem with code. it should use `Double Checked Locking`
> 
> Did you face some issues? Could you paste the logs here? Or just from code review perspective?

I just found it when I am studying the skywalking-java source code

wu-sheng:
It seems you broke the code style, please check before submit.

wu-sheng (approved):

PR #751, opened 2025-03-08
Add the virtual thread executor plugin

peachisai:

### DIFF
--- a/apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
+++ b/apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
@@ -260,4 +260,6 @@ public class ComponentsDefine {
     public static final OfficialComponent SOLON_MVC = new OfficialComponent(158, "SolonMVC");
 
     public static final OfficialComponent CAFFEINE = new OfficialComponent(160, "Caffeine");
+
+    public static final OfficialComponent THREAD_PER_TASK_EXECUTOR = new OfficialComponent(161, "ThreadPerTask-executor");
 }
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
@@ -156,6 +156,11 @@ public final class Tags {
     public static final StringTag THREAD_ID = new StringTag(23, "thread.id");
 
     /**
+     * CARRIER_THREAD represents the actual operating system thread that carries out the execution of the virtual thread.
+     */
+    public static final StringTag CARRIER_THREAD = new StringTag(24, "carrier.thread");
+
+    /**
      * Creates a {@code StringTag} with the given key and cache it, if it's created before, simply return it without
      * creating a new one.
      *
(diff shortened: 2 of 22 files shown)


apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -260,4 +260,6 @@ public class ComponentsDefine {
     public static final OfficialComponent SOLON_MVC = new OfficialComponent(158, "SolonMVC");
 
     public static final OfficialComponent CAFFEINE = new OfficialComponent(160, "Caffeine");
+
+    public static final OfficialComponent THREAD_PER_TASK_EXECUTOR = new OfficialComponent(161, "ThreadPerTask-executor");
```
wu-sheng, line 264: ID submitted through https://github.com/apache/skywalking/pull/13090

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/context/tag/Tags.java
```diff
@@ -155,6 +155,11 @@ public static final class HTTP {
      */
     public static final StringTag THREAD_ID = new StringTag(23, "thread.id");
 
+    /**
+     * CARRIER_THREAD represents the actual operating system thread that carries out the execution of the virtual thread.
+     */
+    public static final StringTag CARRIER_THREAD = new StringTag(24, "carrier.thread");
```
wu-sheng, line 161: To follow the naming style, I think the `thread.carrier` is better?

peachisai: > To follow the naming style, I think the `thread.carrier` is better?

Yes, better, done.

wu-sheng:
Could you try to fix the deadlink? I think the previous link is pointing to very old repo, which should be changed to skywalking asf main repo.

peachisai:
> Could you try to fix the deadlink? I think the previous link is pointing to very old repo, which should be changed to skywalking asf main repo.



> Could you try to fix the deadlink? I think the previous link is pointing to very old repo, which should be changed to skywalking asf main repo.



> Could you try to fix the deadlink? I think the previous link is pointing to very old repo, which should be changed to skywalking asf main repo.

Done.

peachisai:
> Could you try to fix the deadlink? I think the previous link is pointing to very old repo, which should be changed to skywalking asf main repo.

Done.

apm-sniffer/bootstrap-plugins/jdk-virtual-thread-executor-plugin/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+    </properties>
+
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-deploy-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
```
wu-sheng, line 44: Could you check whether your new module could generate javadoc jar? Another plugin pom.xml has relative configurations to generate that file.

wu-sheng (approved):
LGTM

apm-sniffer/bootstrap-plugins/jdk-virtual-thread-executor-plugin/pom.xml
peachisai: When I added the maven-javadoc-plugin, I could generate the javadoc.jar

peachisai: > Could you check whether your new module could generate javadoc jar? Another plugin pom.xml has relative configurations to generate that file.

When I added the maven-javadoc-plugin, I could generate the javadoc.jar

wu-sheng: Please add that plugin, otherwise, this will fail the release process.

PR #754, opened 2025-05-07
To prevent NPE use pool name for metrics if JDBC URL is not set.

john9x:
### Fix NPE 
```
java.lang.NullPointerException: Cannot invoke "String.toLowerCase()" because "url" is null
        at org.apache.skywalking.apm.plugin.jdbc.connectionurl.parser.URLParser.parser(URLParser.java:47)
        at org.apache.skywalking.apm.plugin.hikaricp.PoolingSealInterceptor.afterMethod(PoolingSealInterceptor.java:52)
        at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:110)
        at com.zaxxer.hikari.HikariDataSource.seal(HikariDataSource.java)
        at com.zaxxer.hikari.HikariDataSource.$sw$original$getConnection$08spks0(HikariDataSource.java:112)
        at com.zaxxer.hikari.HikariDataSource.$sw$original$getConnection$08spks0$accessor$$sw$6kb3bn3(HikariDataSource.java)
        at com.zaxxer.hikari.HikariDataSource$$sw$auxiliary$93993q0.call(Unknown Source)
        at org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstMethodsInter.intercept(InstMethodsInter.java:95)
```
The pool can be configured with data source options but not JDBC URL so in this case we can use pool name for metrics.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ Release Notes.
 
 * Add the virtual thread executor plugin
 * Fix Conflicts apm-jdk-threadpool-plugin conflicts with apm-jdk-forkjoinpool-plugin
+* Fix NPE in hikaricp-plugin if JDBC URL is not set
 
 All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/236?closed=1)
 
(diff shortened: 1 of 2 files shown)


wu-sheng:
Please provide the demo codes to represent two ways of using.
And changes.md should be updated. Please follow PR template.

john9x:
> Please provide the demo codes to represent two ways of using.

@wu-sheng How should I do this?

https://github.com/brettwooldridge/HikariCP/?tab=readme-ov-file#rocket-initialization
HikariCP pool can be configured using JDBC URL:
```
HikariConfig config = new HikariConfig();
config.setJdbcUrl("jdbc:mysql://localhost:3306/simpsons");
```
or using data source properties
```
Properties props = new Properties();
props.setProperty("dataSourceClassName", "org.postgresql.ds.PGSimpleDataSource");
props.setProperty("dataSource.serverName", "localhost");
...
HikariConfig config = new HikariConfig(props);
HikariDataSource ds = new HikariDataSource(config)
```

wu-sheng (approved):

PR #755, opened 2025-05-13
Fix MeterService NPE as not initialized in so11y plugin.

wu-sheng:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixed things mentioned in https://github.com/apache/skywalking/issues/13240 and https://github.com/apache/skywalking/discussions/13241.

* Agent kernel services could be not-booted-yet as ServiceManager#INSTANCE#boot executed after agent transfer
  initialization. Delay so11y metrics#build when the services are not ready to avoid MeterService status is not
  initialized.


kezhenxu94 (approved):

pan3793:
Does the community have a schedule for the next release? I hit this issue ...

wu-sheng:
We don't have a release date yet.
Usually, we release agent im every 5-7 months, as it is stable.

PR #757, opened 2025-05-27
fix parent class enhance issue

tjuwangning:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes #<756>.

### DIFF
--- a/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
+++ b/apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
@@ -19,6 +19,7 @@
 package org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance;
 
 import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.modifier.Visibility;
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.dynamic.DynamicType;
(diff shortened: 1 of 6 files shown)


lujiajing1126 (approved):
LGTM. Pls update CHANGES.md

lujiajing1126:
@kylixs Could you pls check this PR?

apm-sniffer/apm-agent-core/src/main/java/org/apache/skywalking/apm/agent/core/plugin/interceptor/enhance/ClassEnhancePluginDefine.java
```diff
@@ -102,6 +103,9 @@ protected DynamicType.Builder<?> enhanceInstance(TypeDescription typeDescription
                 newClassBuilder = newClassBuilder.defineField(
                     CONTEXT_ATTR_NAME, Object.class, ACC_PRIVATE | ACC_VOLATILE)
                                                  .implement(EnhancedInstance.class)
+                                                 .defineMethod("getSkyWalkingDynamicField", Object.class, Visibility.PUBLIC)
+                                                 .intercept(FieldAccessor.ofField(CONTEXT_ATTR_NAME))
+                                                 .defineMethod("setSkyWalkingDynamicField", void.class, Visibility.PUBLIC).withParameters(Object.class)
```
wu-sheng, line 108: These two method name strings should be put in the parent class as static string fields.

tjuwangning: Okay, I've updated it.

wu-sheng (approved):

lujiajing1126 (approved):

PR #758, opened 2025-06-09
Add support for dameng(DM) JDBC URL format in URLParser

saber-wang:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@ Release Notes.
   initialization. Delay so11y metrics#build when the services are not ready to avoid MeterService status is not
   initialized.
 * Fix retransform failure when enhancing both parent and child classes.
-* Add support for `dameng(DM)` jdbc url format in `URLParser`.
+* Add support for `dameng(DM)` JDBC url format in `URLParser`.
 
 All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/236?closed=1)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Release Notes.
   initialization. Delay so11y metrics#build when the services are not ready to avoid MeterService status is not
   initialized.
 * Fix retransform failure when enhancing both parent and child classes.
+* Add support for `dameng(DM)` jdbc url format in `URLParser`.
 
 All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/236?closed=1)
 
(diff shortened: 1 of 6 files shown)


wu-sheng:
I think that although you added this parser, the DB connection must be managed by supported JDBC pooling. Right?

apm-protocol/apm-network/src/main/java/org/apache/skywalking/apm/network/trace/component/ComponentsDefine.java
```diff
@@ -262,4 +262,7 @@ public class ComponentsDefine {
     public static final OfficialComponent CAFFEINE = new OfficialComponent(160, "Caffeine");
 
     public static final OfficialComponent THREAD_PER_TASK_EXECUTOR = new OfficialComponent(161, "ThreadPerTask-executor");
+
+    public static final OfficialComponent DMDB_JDBC_DRIVER = new OfficialComponent(162, "Dmdb-jdbc-driver");
```
wu-sheng: You need to register this ID in the main repo first. Please follow plugin contribution doc.

wu-sheng: And considering the license issue, we would not require adding dameng DB logo into the UI repo.

saber-wang:
> I think that although you added this parser, the DB connection must be managed by supported JDBC pooling. Right?

Yes, because this is a standard JDBC driver, so using something like HikariCP is absolutely fine.

wu-sheng:
I don mean removing from doc. I mean in other PR, you could see there is a PR in UI repo.
For this, we would require that.

saber-wang:
@wu-sheng  
I'm sorry, I've gone through the related PR and guidelines once more, but I'm still not entirely clear on what UI this is referring to.

docs/en/setup/service-agent/java-agent/Supported-list.md
```diff
@@ -185,6 +185,7 @@ The meter plugin provides the advanced metrics collections, which are not a part
     * [DB2](https://www.ibm.com/products/db2/database)
     * Sybase
     * [OceanBase](https://www.oceanbase.com/)
+    * DM
```
wu-sheng: Please revert this to have a link.

wu-sheng:
I mean you don't need this. https://github.com/apache/skywalking-booster-ui/pull/403/files

saber-wang:
Thanks a lot for the help 🙏!

CHANGES.md
```diff
@@ -12,6 +12,7 @@ Release Notes.
   initialization. Delay so11y metrics#build when the services are not ready to avoid MeterService status is not
   initialized.
 * Fix retransform failure when enhancing both parent and child classes.
+* Add support for `dameng(DM)` jdbc url format in `URLParser`.
```
wu-sheng:
```suggestion
* Add support for `dameng(DM)` JDBC url format in `URLParser`.
```

wu-sheng (approved):

PR #762, opened 2025-07-10
Fix enhanced instance causes Spring Application startup failure

tjuwangning:
### Fix <https://github.com/apache/skywalking/issues/13221>

After Spring Framework 4.0.2.RELEASE, if we use a Spring Advisor to add an interface to a class and that same class is also enhanced by SkyWalking, Spring will fail to start.

Reproduce (using spring-retry as an example):
* Enable Spring Retry by annotating a configuration with `@EnableRetry` and the method with `@Retryable`, the Spring Retry's annotations cause the bean class to implement the `Retryable` interface at runtime.
* If SkyWalking has also instrumented the same class, Spring's startup will abort with an exception.
exception

Root cause:
SkyWalking's enhancement changes how Spring decides which proxy type to use. As mentioned in [#13221](https://github.com/apache/skywalking/issues/13221#issuecomment-2854170165), SkyWalking modifies Spring's `hasNoUserSuppliedProxyInterfaces()` to avoid affecting the choice of proxy type. But once Spring Retry has added `Retryable` to the list of interfaces, `hasNoUserSuppliedProxyInterfaces()` will find a user-supplied interface and return false, which changes the proxy type and leads to a startup failure.

Starting in 4.0.2.RELEASE, Spring adds a new hook method `evaluateProxyInterfaces()`, which allows proxy-related flags (like proxyTargetClass) to be determined earlier in the lifecycle. We also need to enhance `evaluateProxyInterfaces()` (and its related callbacks) so that SkyWalking's EnhancedInstance interface is always ignored.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/13221


wu-sheng:
I am on vacation, can't review this until 26th.
Please make CI passed and polish codes as much as you can.

wu-sheng:
@lujiajing1126 Is this a case you are familiar with?

wu-sheng:
Please update supported-list.d doc to add Spring retry support?
局部截取_20250727_103322

test/plugin/scenarios/spring-retry-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,48 @@
...
+            spanId: 1
+            spanLayer: Unknown
+            startTime: not null
+            endTime: not null
+            componentId: not null
+            isError: false
+            spanType: Local
+            peer: ''
+            skipAnalysis: false
+          - operationName: HEAD:/case/healthCheck
```
wu-sheng: This is not required from the expectation file.

test/plugin/scenarios/spring-retry-scenario/config/expectedData.yaml
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+segmentItems:
+  - serviceName: spring-retry-scenario
+    segmentSize: nq 0
+    segments:
+      - segmentId: not null
+        spans:
+          - operationName: test.apache.skywalking.apm.testcase.spring.retry.service.CaseService.handle()
```
wu-sheng: There is no client span and I can't see the retry working in the expectation file.

I think you should enhance your test scenario.

lujiajing1126:
> @lujiajing1126 Is this a case you are familiar with?

@tjuwangning Ning Wang also works in our team. We've already discussed this patch offline.

I suppose he has to add some more detailed technical comments for this PR and explain better how the test case works.

wu-sheng:
Yes, please. I am a little confused about reviewing this test scenario.

lujiajing1126:
> Please update supported-list.d doc to add Spring retry support? 局部截取_20250727_103322

Actually, the original issue can be reproduced in any case that an additional interface is added for the original class. In the `spring-retry` case, it is the marker interface [`Retryable`](https://docs.spring.io/spring-retry/docs/api/current/org/springframework/retry/interceptor/Retryable.html)

wu-sheng:
> Actually, the original issue can be reproduced in any case that an additional interface is added for the original class. In the `spring-retry` case, it is the marker interface [`Retryable`](https://docs.spring.io/spring-retry/docs/api/current/org/springframework/retry/interceptor/Retryable.html)

Is this error caused by adding retry annotation, or retry actually happened?

tjuwangning:
> > Actually, the original issue can be reproduced in any case that an additional interface is added for the original class. In the `spring-retry` case, it is the marker interface [`Retryable`](https://docs.spring.io/spring-retry/docs/api/current/org/springframework/retry/interceptor/Retryable.html)
> 
> Is this error caused by adding retry annotation, or retry actually happened?

Spring cannot startup when add the retry annotation.
I hava revised the PR description to explain the issue more clearly. Furthermore, the current name of the test scenario is somewhat ambiguous, I will rename it later.

wu-sheng:
Then, I think you just need to add this retry annotation to existing scenario.

apm-sniffer/apm-sdk-plugin/spring-plugins/core-patch/src/main/java/org/apache/skywalking/apm/plugin/spring/patch/AutoProxyCreatorInterceptor.java
```diff
@@ -0,0 +1,53 @@
...
+package org.apache.skywalking.apm.plugin.spring.patch;
+
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+
+import java.lang.reflect.Method;
+
+/**
+ * <code>AutoProxyCreatorInterceptor</code> check that the bean has been implement {@link EnhancedInstance}.
```
wu-sheng: I am clear about your new description, but a little confused about this comment word.
I think the original method is returning true/false, and we are just checking `EnhancedInstance` implementation when the original return is false. 
When the original return is false? We should write this comment according to `#isConfigurationCallbackInterface` purpose, right?

wu-sheng:
And please update changes.md

apm-sniffer/apm-sdk-plugin/spring-plugins/core-patch/src/main/java/org/apache/skywalking/apm/plugin/spring/patch/AutoProxyCreatorInterceptor.java
tjuwangning: Yes, I have updated it.

test/plugin/scenarios/spring-4.1.x-scenario/src/main/java/test/apache/skywalking/apm/testcase/spring3/config/RetryConfig.java
```diff
@@ -0,0 +1,9 @@
+package test.apache.skywalking.apm.testcase.spring3.config;
```
wu-sheng, line 19: You missed the APLv2 header.

PR #771, opened 2025-08-21
Bump up apache parent pom to v35.

wu-sheng:
* Bump up apache parent pom to v35.
* Update Maven to 3.6.3 in mvnw.

### DIFF
--- a/apm-sniffer/bootstrap-plugins/jdk-http-plugin/pom.xml
+++ b/apm-sniffer/bootstrap-plugins/jdk-http-plugin/pom.xml
@@ -32,17 +32,49 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>8</maven.compiler.target>
+        <!-- Reset this to the blank value to avoid using -release in JDK compiling -->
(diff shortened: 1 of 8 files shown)


apm-sniffer/bootstrap-plugins/jdk-http-plugin/pom.xml
```diff
@@ -32,6 +32,8 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <!-- Reset this to the blank value to avoid using -release in JDK compiling -->
+        <maven.compiler.release></maven.compiler.release>
```
wu-sheng, line 36: This is intentional. Refer to Maven doc,


> Starting with JDK 9, the javac executable can accept the --release option to specify against which Java SE release you want to build the project. For example, you have JDK 11 installed and used by Maven, but you want to build the project against Java 8. The --release option ensures that the code is compiled following the rules of the programming language of the specified release, and that generated classes target the release **as well as the public API of that release**. This means that, unlike the old [-source and -target options](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html), the compiler will detect and generate an error when using APIs that don't exist in previous releases of Java SE.

- https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

PR #772, opened 2025-09-03
Fix CLASS_LOADER_TYPE_CACHE OOM issue with WeakHashMap

kylixs:
### Problem
  The CLASS_LOADER_TYPE_CACHE in SWDescriptionStrategy uses a two-level cache structure with ClassLoader.hashCode() as the first-level key and type
  name as the second-level key. In scenarios involving rule engines like Drools that dynamically generate large numbers of classes, this cache grows
  indefinitely and causes OutOfMemoryError during long-running applications.

Original Discussion, https://github.com/apache/skywalking/discussions/12627

  ### Root Cause
  1. No cache cleanup mechanism: The cache retains entries permanently, even after ClassLoaders are garbage collected
  2. HashCode collision risk: Different ClassLoaders may have the same hashCode, causing cache conflicts
  3. Memory leak: Dynamic classes are typically used once but their cache entries persist forever

  ### Solution
  Replace the hashCode-based cache with a WeakHashMap that automatically cleans up entries when ClassLoaders are garbage collected:

```java
  // Before: Memory leak prone
  private static final Map<Integer, Map<String, TypeCache>> CLASS_LOADER_TYPE_CACHE = new ConcurrentHashMap<>();

  // After: Automatic cleanup
  private static final Map<ClassLoader, Map<String, TypeCache>> CLASS_LOADER_TYPE_CACHE =
      Collections.synchronizedMap(new WeakHashMap<>());
```

  ### Key Changes
  1. WeakHashMap Implementation: Automatically removes cache entries when ClassLoader is GC'd
  2. Direct ClassLoader Reference: Use ClassLoader object as key instead of hashCode to avoid collisions
  3. Bootstrap ClassLoader Handling: Separate cache for null ClassLoader since WeakHashMap doesn't support null keys

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ Release Notes.
 * Bump up apache parent pom to v35.
 * Update Maven to 3.6.3 in mvnw.
 * Fix OOM due to too many span logs.
+* Fix ClassLoader cache OOM issue with WeakHashMap.
 
 All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/242?closed=1)
 


wu-sheng:
@gaojingyuan001 @hyhyf This one.

apm-sniffer/bytebuddy-patch/src/test/java/net/bytebuddy/agent/builder/SWDescriptionStrategyCacheTest.java
```diff
@@ -0,0 +1,239 @@
...
+import java.util.Map;
+
+/**
+ * Tests the behavior of the WeakHashMap caching mechanism in SWDescriptionStrategy.
+ */
+public class SWDescriptionStrategyCacheTest {
+
+    @Test
+    public void testWeakHashMapCacheCleanup() throws Exception {
+        // 获取静态缓存字段
```
wu-sheng: Please fix all Chinese in comments.

wu-sheng:
@gaojingyuan001 @hyhyf Any of you tried this one? This PR was raised due to your use cases.

wu-sheng:
@kylixs As this is an enhancement and it seems no harm to the whole system, could you update the changes and we consider to merge this if there is no further feedback?

Ax1an:
After applying this update for a week, the service using Drools no longer experienced memory leaks. I think it can be merged. @wu-sheng @kylixs

wu-sheng:
> After applying this update for a week, the service using Drools no longer experienced memory leaks. I think it can be merged. @wu-sheng @kylixs 

Thanks for the feedback.

@kylixs waiting for you to update change logs.

kylixs:
> > After applying this update for a week, the service using Drools no longer experienced memory leaks. I think it can be merged. @wu-sheng @kylixs
> 
> Thanks for the feedback.
> 
> @kylixs waiting for you to update change logs.

done

wu-sheng (approved):

PR #774, opened 2025-09-06
Support tagging cache keys for byte[] type in Jedis commands in jedis-2.x  and jedis-3.x

youjie23:
fix  https://github.com/apache/skywalking/issues/13478

Currently, the jedis-2.x-3.x-plugin only supports collecting the `cache.key` tag for keys of `String` type. However, in some application codes, methods corresponding to the `byte[]` key type are invoked due to various reasons . This limitation prevents the cache.keytag from being collected for these byte[]-type keys, potentially leading to incomplete monitoring or tracing data.

It is necessary to extend the plugin's capability to also collect the cache.keytag for keys of byte[]type, ensuring comprehensive coverage and consistent monitoring for both key types.

### Support tagging cache keys for `byte[]` type in Jedis commands in jedis-2.x  and jedis-3.x
- [X] Tests(including UT, IT, E2E) are added to verify the new feature.
- [X] Closes 13478.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,7 +14,7 @@ Release Notes.
 * Bump up cli to the 0.15.0-dev.latest(77b4c49e89c9c000278f44e62729d534f2ec842e) in e2e.
 * Bump up apache parent pom to v35.
 * Update Maven to 3.6.3 in mvnw.
-
+* Support tagging cache keys for `byte[]` type in Jedis commands in jedis-2.x and jedis-3.x
 All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/242?closed=1)
 
 ------------------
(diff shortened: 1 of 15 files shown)


wu-sheng:
As a byte array, how can you be sure it is a utf8 encoded string? Is there any redis doc to guarantee that?

wu-sheng:
byte array could represent image, even video from general usage. It doesn't have to be an encoded string.

wu-sheng:
This was discussed back to the plugin was being added/reviewed.
I am not sure there is a proper way to deal with byte array, unless this is an internal usage.

youjie23:
> This was discussed back to the plugin was being added/reviewed. I am not sure there is a proper way to deal with byte array, unless this is an internal usage.

Thank you for the review.
Yes, you are rigorous. Redis will directly store the passed byte[] regardless of its content encoding and type. It is imprudent to directly assume that it can be converted into a string using UTF-8 (this is only for our internal usage scenario).
Should we close this PR directly? Or should we attempt a conversion, like the code below?
```
private Optional<String> getKey(Object[] allArguments) {
         //code
        if (argument instanceof String) {
            return Optional.of(StringUtil.cut((String) argument, JedisPluginConfig.Plugin.Jedis.REDIS_PARAMETER_MAX_LENGTH));
        }
        if (argument instanceof byte[]) {
           try {
                 String key = new String((byte[]) argument, StandardCharsets.UTF_8);
                 return Optional.of(StringUtil.cut(key, JedisPluginConfig.Plugin.Jedis.REDIS_PARAMETER_MAX_LENGTH));
           } 
          catch{
          }
        }
        return Optional.empty();
    }
```

wu-sheng:
I don't think this is a proper way, especially since it increases the volume cost when a large value object.
I am going to close this PR and the original issue.

PR #816, opened 2026-07-29
Add tracing support for ThreadPoolExecutor invokeAll and invokeAny

### LINKED ISSUE #13945: [Feature] Add tracing support for ThreadPoolExecutor.invokeAll and invokeAny in the JDK thread pool plugin
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

Add tracing support for `ThreadPoolExecutor.invokeAll` and `ThreadPoolExecutor.invokeAny` in the JDK thread pool plugin.

The plugin currently propagates tracing context for tasks submitted through `ThreadPoolExecutor.execute` and `ThreadPoolExecutor.submit`, but the following inherited batch task submission methods are not covered:

```java
<T> List<Future<T>> invokeAll(
    Collection<? extends Callable<T>> tasks
)

<T> List<Future<T>> invokeAll(
    Collection<? extends Callable<T>> tasks,
    long timeout,
    TimeUnit unit
)

<T> T invokeAny(
    Collection<? extends Callable<T>> tasks
)

<T> T invokeAny(
    Collection<? extends Callable<T>> tasks,
    long timeout,
    TimeUnit unit
)
```
ThreadPoolExecutor inherits these methods from AbstractExecutorService.

Internally, invokeAll and invokeAny convert or wrap the original Callable tasks as RunnableFuture-based tasks before passing them to execute. The current execute interceptor skips RunnableFuture instances, while the original Callable tasks have not captured the active tracing context.

As a result, tasks executed through ThreadPoolExecutor.invokeAll or ThreadPoolExecutor.invokeAny do not continue the caller's trace and have no CrossThread reference to the parent trace segment.

Supporting these methods would make tracing context propagation consistent across the individual and batch task submission APIs of ThreadPoolExecutor.

### Use case

```java
ThreadPoolExecutor executor = new ThreadPoolExecutor(
    2,
    2,
    0L,
    TimeUnit.MILLISECONDS,
    new LinkedBlockingQueue<Runnable>()
);

List<Callable<String>> tasks = Arrays.asList(
    () -> callDownstreamService("task-1"),
    () -> callDownstreamService("task-2")
);

List<Future<String>> results = executor.invokeAll(tasks);

String firstResult = executor.invokeAny(tasks);
```

When these methods are called within an active SkyWalking trace, every Callable that starts execution should continue the caller's tracing context. Downstream HTTP, database, or RPC operations should be connected to the same trace through CrossThread references.

Currently, work submitted through ThreadPoolExecutor.invokeAll and ThreadPoolExecutor.invokeAny may appear as disconnected traces or may be missing from the parent trace.

The expected behavior is:
1. Support both overloads of ThreadPoolExecutor.invokeAll.
2. Support both overloads of ThreadPoolExecutor.invokeAny.
3. Each Callable that starts execution continues the active tracing context.
4. Each worker-thread trace segment contains a CrossThread reference to the caller.
5. Existing timeout, cancellation, exception, and result-returning semantics remain unchanged.
6. Existing behavior for execute, submit, and already wrapped tasks remains unchanged.
7. Plugin scenario tests verify tracing context propagation for all four overloads.

I implemented a local prototype for `ThreadPoolExecutor.invokeAll` to validate the feasibility of this feature.

After adding tracing context propagation, the `Callable` tasks submitted through `invokeAll` are connected to the caller's trace through `CrossThread` references, as shown below:

Image

### Related issues

_No response_

### Are you willing to submit a pull request to implement this on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

xuzhiguang:
### Support tracing for `ThreadPoolExecutor.invokeAll` and `invokeAny`
- [x] Closes apache/skywalking#13945.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Release Notes.
 * Add a Jetty 12 server plugin (`jetty-server-12.x`). Jetty 12 removed the `HttpChannel` handle target and moved request handling to the async `Server#handle(Request, Response, Callback)` core API, so it needs a separate plugin from the merged `jetty-server`.
 * Add a Struts 7 plugin (`struts2-7.x`) for Jakarta Struts, whose `DefaultActionInvocation` moved to `org.apache.struts2`.
 * Added support for Lettuce reactive Redis commands.
+* Add tracing support for `invokeAll` and `invokeAny` in the JDK thread pool plugin.
 * Add Spring AI 1.x plugin and GenAI layer.
 * Fix httpclient-5.x plugin injecting sw8 propagation headers into ClickHouse HTTP requests (port 8123), causing HTTP 400. Add `PROPAGATION_EXCLUDE_PORTS` config to skip tracing (including header injection) for specified ports in the classic client interceptor.
 * Add Spring RabbitMQ 2.x - 4.x plugin.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Release Notes.
 * Add a Jetty 12 server plugin (`jetty-server-12.x`). Jetty 12 removed the `HttpChannel` handle target and moved request handling to the async `Server#handle(Request, Response, Callback)` core API, so it needs a separate plugin from the merged `jetty-server`.
 * Add a Struts 7 plugin (`struts2-7.x`) for Jakarta Struts, whose `DefaultActionInvocation` moved to `org.apache.struts2`.
 * Added support for Lettuce reactive Redis commands.
-* Add tracing support for `invokeAll` and `invokeAny` in the JDK thread pool plugin.
+* Add tracing support for `invokeAll` and `invokeAny` in the JDK thread pool plugin (`jdk-threadpool-plugin`).
 * Add Spring AI 1.x plugin and GenAI layer.
 * Fix httpclient-5.x plugin injecting sw8 propagation headers into ClickHouse HTTP requests (port 8123), causing HTTP 400. Add `PROPAGATION_EXCLUDE_PORTS` config to skip tracing (including header injection) for specified ports in the classic client interceptor.
 * Add Spring RabbitMQ 2.x - 4.x plugin.
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolInvokeMethodInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolInvokeMethodInterceptor.java
@@ -27,24 +27,25 @@ import org.apache.skywalking.apm.agent.core.context.ContextManager;
 import org.apache.skywalking.apm.agent.core.context.ContextSnapshot;
 import org.apache.skywalking.apm.agent.core.context.trace.AbstractSpan;
 import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.EnhancedInstance;
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.InstanceMethodsAroundInterceptor;
-import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.MethodInterceptResult;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.InstanceMethodsAroundInterceptorV2;
+import org.apache.skywalking.apm.agent.core.plugin.interceptor.enhance.v2.MethodInvocationContext;
 import org.apache.skywalking.apm.network.trace.component.ComponentsDefine;
 import org.apache.skywalking.apm.plugin.wrapper.SwCallableWrapper;
 
-public class ThreadPoolInvokeMethodInterceptor implements InstanceMethodsAroundInterceptor {
+public class ThreadPoolInvokeMethodInterceptor implements InstanceMethodsAroundInterceptorV2 {
 
     private static final String OPERATION_NAME_PREFIX = "ThreadPoolExecutor/";
 
     @Override
     public void beforeMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
-                             MethodInterceptResult result) throws Throwable {
+                             MethodInvocationContext context) throws Throwable {
         if (!shouldEnhance(allArguments)) {
             return;
         }
 
         AbstractSpan span = ContextManager.createLocalSpan(OPERATION_NAME_PREFIX + method.getName());
         span.setComponent(ComponentsDefine.JDK_THREADING);
+        context.setContext(span);
 
         ContextSnapshot contextSnapshot = ContextManager.capture();
         Collection<?> callables = (Collection<?>) allArguments[0];
@@ -57,18 +58,20 @@ public class ThreadPoolInvokeMethodInterceptor implements InstanceMethodsAroundI
 
     @Override
     public Object afterMethod(EnhancedInstance objInst, Method method, Object[] allArguments, Class<?>[] argumentsTypes,
-                              Object ret) throws Throwable {
-        if (shouldEnhance(allArguments)) {
-            ContextManager.stopSpan();
+                              Object ret, MethodInvocationContext context) throws Throwable {
+        AbstractSpan span = (AbstractSpan) context.getContext();
+        if (span != null) {
+            ContextManager.stopSpan(span);
         }
         return ret;
     }
 
     @Override
     public void handleMethodException(EnhancedInstance objInst, Method method, Object[] allArguments,
-                                      Class<?>[] argumentsTypes, Throwable t) {
-        if (shouldEnhance(allArguments)) {
-            ContextManager.activeSpan().log(t);
+                                      Class<?>[] argumentsTypes, Throwable t, MethodInvocationContext context) {
+        AbstractSpan span = (AbstractSpan) context.getContext();
+        if (span != null) {
+            span.log(t);
         }
     }
 
--- a/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolInvokeMethodInterceptor.java
+++ b/apm-sniffer/bootstrap-plugins/jdk-threadpool-plugin/src/main/java/org/apache/skywalking/apm/plugin/ThreadPoolInvokeMethodInterceptor.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.skywalking.apm.plugin;
+
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Collection;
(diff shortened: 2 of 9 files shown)


wu-sheng (commented):
Thanks for the PR. The premise is correct — `invokeAll`/`invokeAny` convert tasks to `FutureTask`/`QueueingFuture` before calling `execute`, and `ThreadPoolExecuteMethodInterceptor.wrap()` explicitly returns `null` for `RunnableFuture`, so today these four overloads propagate nothing. Wrapping the `Callable`s before the JDK converts them is the right hook, iteration order is preserved, the caller's collection is not mutated, and cancelled `invokeAny` losers simply never run.

There is one blocking problem, though.

## Duplicate nested spans on every `ThreadPoolExecutor` subclass

`ThreadPoolExecutorInstrumentation#enhanceClass()` matches `HierarchyMatch.byHierarchyMatch(...)` **or** `MultiClassNameMatch(...)`, and the new intercept point is a plain `InstanceMethodsInterceptPoint`. `ClassEnhancePluginDefine` only adds `isDeclaredBy(typeDescription)` for `DeclaredInstanceMethodsInterceptPoint`:

```java
ElementMatcher.Junction<MethodDescription> junction = not(isStatic()).and(instanceMethodsInterceptPoint.getMethodsMatcher());
if (instanceMethodsInterceptPoint instanceof DeclaredInstanceMethodsInterceptPoint) {
    junction = junction.and(ElementMatchers.<MethodDescription>isDeclaredBy(typeDescription));
}
```

So `named("invokeAll").or(named("invokeAny"))` matches the **inherited** `AbstractExecutorService` methods on `java.util.concurrent.ThreadPoolExecutor` *and* on every subclass, and ByteBuddy generates an override in each one. The subclass's generated override then super-calls the base class's *enhanced* method, so the interceptor runs twice.

### Reproduction on current `main`

This is not specific to your patch — `submit` is inherited in exactly the same way, so it can be demonstrated on `main` today. Given a subclass that declares nothing:

```java
public static class MyPool extends ThreadPoolExecutor {
    public MyPool() { super(2, 2, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>()); }
}
```

run with `-Dskywalking.plugin.mount=plugins,activations,bootstrap-plugins` and `SW_AGENT_OPEN_DEBUG=true`, then look at the dumped bytecode.

**1. The base class gets a generated override.** `ThreadPoolExecutor` never declared `submit`, but after enhancement:

```
== java.util.concurrent.ThreadPoolExecutor  ifaces=[EnhancedInstance]
   declared: public Future ThreadPoolExecutor.submit(Callable)        <-- added by ByteBuddy
   declared: public Future ThreadPoolExecutor.submit(Runnable)        <-- added
   declared: public Future ThreadPoolExecutor.submit(Runnable,Object) <-- added
```

```
ThreadPoolExecutor.submit(Callable):
  20: invokestatic  ThreadPoolSubmitMethodInterceptor_internal.intercept(...)

ThreadPoolExecutor.submit$accessor$$sw$g28bpd0(Callable):
   2: invokespecial java/util/concurrent/AbstractExecutorService.submit    <-- real JDK impl, OK
```

**2. The subclass gets one too**, even though it declares nothing:

```
== Probe$MyPool  ifaces=[EnhancedInstance]
   declared: public Future MyPool.submit(Callable)

MyPool.submit(Callable):
  20: invokestatic  ThreadPoolSubmitMethodInterceptor_internal.intercept(...)
```

**3. And its super-call lands on the *enhanced* base method**, not on `AbstractExecutorService`:

```
MyPool.submit$accessor$$sw$ssa2tc1(Callable):
   2: invokespecial java/util/concurrent/ThreadPoolExecutor.submit    <-- the enhanced one
```

So a single `myPool.submit(task)` runs:

```
MyPool.submit                 -> interceptor run #1 -> invokespecial ThreadPoolExecutor.submit
ThreadPoolExecutor.submit     -> interceptor run #2 -> invokespecial AbstractExecutorService.submit
AbstractExecutorService.submit -> actual JDK work
```

### Why this is harmless today but not with this PR

`ThreadPoolSubmitMethodInterceptor` and `ThreadPoolExecuteMethodInterceptor` create **no span**, and `wrap()` is idempotent (`instanceof SwCallableWrapper` -> `null`), so run #2 is a silent no-op.

`ThreadPoolInvokeMethodInterceptor#beforeMethod` opens with an unconditional `ContextManager.createLocalSpan(...)`. On any subclass that produces:

```
GET:/greet/{username}                    spanId 0
└── ThreadPoolExecutor/invokeAll         spanId 1   <- subclass level: wraps tasks, captures snapshot
    └── ThreadPoolExecutor/invokeAll     spanId 2   <- base level: duplicate, childless leaf
    SwCallableWrapper/...  (worker segments) --ref--> spanId 1
```

The inner span is a pure artifact — the outer call already wrapped everything, so the worker segments reference span 1 and span 2 has no children. A three-level hierarchy produces three.

Affected in practice: `java.util.concurrent.ScheduledThreadPoolExecutor` (it declares its own `submit`, which is why `submit` is safe there, but it does **not** declare `invokeAll`/`invokeAny`), Tomcat's `org.apache.tomcat.util.threads.ThreadPoolExecutor`, and any framework or user pool extending `ThreadPoolExecutor`. The scenario uses a bare `new ThreadPoolExecutor(...)`, which is the single case that does *not* double, so CI cannot catch this.

### Suggested fix

Only open the span when this invocation actually wrapped something. That is false on the nested super-call, because everything is already a `SwCallableWrapper`:

```java
Collection<?> callables = (Collection<?>) allArguments[0];
List<Object> wrapped = new ArrayList<>(callables.size());
boolean wrappedAny = false;
for (Object c : callables) {
    Object w = wrap(c, snapshot);
    wrappedAny |= (w != c);
    wrapped.add(w);
}
if (!wrappedAny) {
    return;   // nested super-call: no span, no re-wrap
}
```

Note this also has to drive `afterMethod`/`handleMethodException`, which currently recompute `shouldEnhance(allArguments)` independently — see the next point.

Please also add a `class NestedPool extends ThreadPoolExecutor {}` to `jdk-threadpool-scenario` and assert the segment still contains exactly one `ThreadPoolExecutor/invokeAll` span, so this stays covered.

## Other comments

**`afterMethod` recomputes the guard instead of remembering the decision.** `shouldEnhance(allArguments)` is evaluated independently in `beforeMethod`, `afterMethod` and `handleMethodException`. It happens to stay consistent today, but it is a fragile contract, and it breaks outright with any fix for the above that is not purely argument-derived.

**The unit tests do not cover the feature.** The three cases exercise `null`/empty/non-`Collection` arguments plus `shouldTraceAlternativeInvokeAllSignature`, which reflects on a synthetic private `invokeAll(Collection, String)` that exists on no executor. Nothing asserts that tasks become `SwCallableWrapper`, that iteration order is preserved, that the caller's collection is untouched, that already-wrapped tasks are skipped, or anything at all about `invokeAny`. Please use the real `invokeAll(Collection, long, TimeUnit)` overload instead of the synthetic signature, and add assertions on the wrapping itself.

**`expectedData.yaml` span order (nit).** The new spans are listed `4, 3, 2, 1, 0`, but spans are archived in finish order (`TraceSegment#archive`), i.e. `1, 2, 3, 4, 0`. It passes only because `SegmentAssert#spansEquals` falls back to matching by `spanId`. Listing them in real order makes future failures much easier to read.

**Design point for the maintainers.** apache/skywalking#13945 asks for context propagation; this PR additionally emits a caller-side `ThreadPoolExecutor/invokeAll` local span, which `execute`/`submit` deliberately do not. It is defensible — `invokeAll` blocks, so the span carries real timing — but it is an unconditional per-call span-count increase, and the operation name reports `ThreadPoolExecutor` even for subclasses. Worth a conscious decision before merging.

xuzhiguang:
@wu-sheng Thanks for the detailed review. My plan is to handle only the outermost enhanced method by checking:
``` java
objInst.getClass() == method.getDeclaringClass()
```
For a ThreadPoolExecutor subclass, this is true in the subclass-generated override, but false when its super call enters the generated ThreadPoolExecutor override.

I will also use a ThreadLocal<Boolean> to record whether the outermost invocation actually created a span. afterMethod and handleMethodException will use this recorded state rather than recomputing from the mutated arguments, and only the invocation that created the span will stop or log it.

Regarding the caller-side ThreadPoolExecutor/invokeAll and invokeAny local span, I believe it is useful and intentional. Unlike execute and submit, these methods block the caller: invokeAll waits for all tasks, while invokeAny waits until one task completes successfully (or times out/fails). The local span therefore represents the batch invocation's wall-clock waiting time.

It also provides a clear parent for the context snapshots captured for the submitted callables, so the worker-side segments are associated with the specific invokeAll / invokeAny operation rather than directly with the broader caller span.

I understand that this adds one caller-side span per invocation. If maintainers prefer this plugin to provide context propagation only, without an invocation span, I can adjust the design accordingly; otherwise I will keep the span with the duplicate-span protection described above.

wu-sheng:
We usually don't recommend threadlocal, unless you have to. It's easy to be polluted and stale. 
We are better to instrument stricted cases to work perfectly.

xuzhiguang:
I reconsidered the implementation. I’ll use an exact match for java.util.concurrent.ThreadPoolExecutor and won’t enhance its subclasses for invokeAll/invokeAny.

Subclasses that inherit these methods will still go through the enhanced ThreadPoolExecutor implementation, which avoids duplicate nested interception.

I’ll also switch the interceptor to InstanceMethodsAroundInterceptorV2 and use MethodInvocationContext to pass the span creation state and span reference between beforeMethod, afterMethod, and handleMethodException. This avoids both ThreadLocal and recomputing the decision from the arguments.

xuzhiguang:
@wu-sheng Updated as discussed. The invokeAll/invokeAny instrumentation now matches ThreadPoolExecutor exactly. Please take another look when you have time.

wu-sheng:
Overall this is in good shape — the design is correct and I could verify the key decisions:

- `NameMatch` (instead of the sibling's `HierarchyMatch + MultiClassNameMatch`) is the right call here: these interceptors **create spans**, so enhancing subclasses too would produce nested duplicate `ThreadPoolExecutor/invokeAll` spans through the `super.invokeAll()` chain. Subclasses are still traced via the inherited instrumented override — the scenario's `NestedThreadPoolExecutor` proves exactly that path.
- Span lifecycle is safe: `afterMethod` always runs in `finally`, and the span is stored in `MethodInvocationContext` before the wrap loop, so no leak even if wrapping fails.
- The internal `execute(FutureTask)` calls don't double-wrap because the existing execute interceptor skips `RunnableFuture` — the two instrumentations compose cleanly.
- Unit tests pass locally and the `jdk-threadpool-scenario` is green; the `ge 14 -> ge 20` segment math checks out.

A few suggested improvements before merging:

1. **`expectedData.yaml`** — the six new child-segment refs assert `parentSpanId: not null`, which also passes with `0`. The headline property of this PR (child segments parented to the invoke spans, not the entry span) is therefore never asserted: if `ContextManager.capture()` were later moved above `createLocalSpan()`, CI would stay green. Please use `parentSpanId: gt 0` — it keeps the six expected segments interchangeable (no flakiness) but catches that regression.
2. **Unit test for the inactive-context guard** — every test creates an entry span first, so the `ContextManager.isActive()` branch of `shouldEnhance` is never exercised. Add one test without an entry span asserting `arguments[0]` is unchanged, `context.getContext()` is null, and no segments are recorded.
3. **`@Mock EnhancedInstance` is never initialized** — `TracingSegmentRunner` doesn't process Mockito annotations and there is no `MockitoRule`, so the field is silently `null` (harmless today since the interceptor never touches `objInst`). Either add `@Rule public MockitoRule rule = MockitoJUnit.rule();` like other plugin tests, or drop `@Mock` and pass `null` explicitly.
4. **`CHANGES.md`** — please include the plugin id so users can map it to `exclude_plugins`, e.g. `...in the JDK thread pool plugin (`jdk-threadpool-plugin`)`.
5. **`docs/en/setup/service-agent/java-agent/Supported-list.md`** — the *Thread Schedule Framework* section lists `JRE Callable and Runnable` and `JRE ForkJoinPool` but never listed the thread pool plugin. Since this PR grows its coverage, add a `JRE ThreadPoolExecutor (Optional²)` entry there.
6. Nit: a one-line comment on `NestedThreadPoolExecutor` noting it deliberately exercises the inherited-dispatch path (the added override lives on `ThreadPoolExecutor` itself) would keep someone from "simplifying" it to a plain pool later.

xuzhiguang:
@wu-sheng Thanks for the review. I addressed the remaining comments.

wu-sheng (approved):
LGTM. If plugin tests passed, it is good to merge. Thanks.

PR #820, opened 2026-08-12
Prompt for the next milestone ID when resetting CHANGES.md

wu-sheng:
### Problem

When `cmd_prepare` resets `CHANGES.md` for the next development version, it writes a
literal placeholder that nothing ever fills in:

```
All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/xxx?closed=1)
```

The release flow has no step that resolves `xxx`, so unless the release manager
spots it and edits by hand, the placeholder rides the release PR straight into
`main` and the next version's change log ships pointing at a dead milestone link.

This was hit while preparing 9.7.0 and had to be corrected manually on the release
branch.

### Fix

Ask for the milestone ID and substitute it.

- **Asked up front**, next to the version confirmation, rather than after
  `release:prepare` — so the release doesn't stop for input in the middle of a
  multi-minute build.
- **Validated as numeric**, and cross-checked against the milestone's title on
  `apache/skywalking`. Entering last release's ID by mistake is reported before
  anything is committed:
  ```
  [WARN]   Milestone 249 is 'Java - 9.7.0', expected 'Java - 9.8.0'. Double-check it.
  ```
- **`NEXT_MILESTONE=<id>`** answers non-interactively for scripted runs.
- **A blank answer** keeps the previous placeholder behaviour, but now warns rather
  than doing it silently.

`gh api` writes its error body to *stdout* on failure, so the title lookup gates on
gh's exit status — otherwise a 404 JSON payload would be reported as the milestone
name.

### Testing

`bash -n` passes. All five input paths exercised against the live GitHub API:

| Input | Result |
|---|---|
| `263` (correct) | `[INFO] Next milestone: 263 (Java - 9.8.0)` |
| `249` (last release's) | `[WARN] Milestone 249 is 'Java - 9.7.0', expected 'Java - 9.8.0'` |
| `999999` (nonexistent) | `[WARN] Could not verify milestone 999999; using it as given` |
| `26x3` (non-numeric) | `[ERROR] Milestone ID must be a number` — aborts |
| blank | `[WARN] ... will keep 'milestone/xxx'` |

The generated `CHANGES.md` is byte-identical to the hand-written correction made on
the 9.7.0 release branch (`diff` clean).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/en/contribution/release-java-agent.md
+++ b/docs/en/contribution/release-java-agent.md
@@ -76,6 +76,14 @@ Then run `gpgconf --kill gpg-agent` and `gpg --sign /dev/null` to cache it.
 4. **upload** — upload to Apache SVN `dist/dev` (prompts for SVN credentials)
 5. **email vote** — print vote email template with pre-filled version, commit ID, submodule commit, and checksums
 
+Before the long build starts, **prepare** asks for the GitHub milestone ID of the next
+development version, which it writes into the reset `CHANGES.md`. Look up the
+`Java - <next_version>` milestone at https://github.com/apache/skywalking/milestones and
+enter its number. The ID is checked against that milestone's title, and you are warned if
+they disagree. Set `NEXT_MILESTONE=<id>` to answer non-interactively; leave the prompt
+blank to keep the `milestone/xxx` placeholder and edit it by hand before merging the
+release PR.
+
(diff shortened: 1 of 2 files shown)


wu-sheng:
Folded into #821 along with the tag-based version fix, so the release-tooling changes land together. The commit is preserved there via cherry-pick.

PR #824, opened 2026-08-20
Add an agent plugin to support Spring LDAP

### LINKED ISSUE #13981: [Feature] Add tracing support for Spring LDAP in the Java Agent
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

SkyWalking Java Agent currently has no dedicated tracing support for Spring LDAP. As a result, LDAP calls made through `LdapTemplate` or `LdapClient` are not represented as meaningful LDAP client spans.

I would like to add a Spring LDAP agent plugin.

### Use case

Applications commonly use LDAP for authentication and directory queries. With this plugin, operators can identify LDAP latency and failures in end-to-end traces without exposing authentication credentials or directory data.

- Which logical LDAP operation was executed.
- The latency and error status of LDAP access.
- Which LDAP server was accessed.

### Related issues

_No response_

### Are you willing to submit a pull request to implement this on your own?

- [x] Yes I am willing to submit a pull request on my own!

### Code of Conduct

weilhuang:
### Add an agent plugin to support Spring LDAP

Add tracing support for non-streaming synchronous operations invoked through `LdapTemplate` and `LdapClient`.

The plugin:

- Creates database exit spans named `SpringLDAP/{operation}`.
- Reports the Spring LDAP component as ID 179.
- Records the logical LDAP operation and the resolved LDAP peer.
- Marks exceptions and failed authentication results as errors.
- Supports wrapped context sources, including `ObservationContextSource`.
- Does not collect LDAP DNs, filters, attributes, or credentials.
- Supports Spring LDAP 3.3.8, 4.0.4, and 4.1.0.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes  apache/skywalking#13981 .


wu-sheng (changes requested):
I found three issues that need addressing:

1. **P1 — Do not collect raw LDAP exception details by default**

`SpringLdapOperationInterceptor.handleMethodException` currently calls `span.log(t)`, which exports the exception message and stack to OAP. Spring LDAP exceptions can contain directory data. For example:

```text
event: error
error.kind: IncorrectResultSizeDataAccessException
message: base: 'ou=people,dc=example,dc=org'; filter: '(mail=alice@example.org)'.
stack: ...the same DN and filter...
```

This conflicts with the requirement not to expose LDAP DNs, filters, credentials, or directory data.

The default behavior should only call `errorOccurred()`, with no message or stack. Exception-detail collection may be provided through an explicit opt-in configuration, such as:

```properties
plugin.spring_ldap.collect_exception_details=false
```

The option must default to `false` and document that exception messages and stacks may contain sensitive LDAP data. Please add tests for both the privacy-safe default and the explicitly enabled behavior.

2. **P2 — A successful authentication mapper can be marked as failed**

The current `Boolean.FALSE` check applies to every operation identified as `authenticate`. However, both `LdapTemplate.authenticate(..., mapper)` and `LdapClient.AuthenticateSpec.execute(mapper)` may successfully authenticate and then legitimately return `Boolean.FALSE` from the mapper.

Please restrict false-result handling to primitive-boolean status methods:

```java
if (AUTHENTICATE_OPERATION.equals(state.operation)
    && method.getReturnType() == boolean.class
    && Boolean.FALSE.equals(ret)) {
    state.span.errorOccurred();
}
```

Please test these cases:

- Primitive `boolean authenticate(...)` returning `false` is an error.
- A successful `LdapTemplate` mapper returning `Boolean.FALSE` is not an error.
- A successful `LdapClient` mapper returning `Boolean.FALSE` is not an error.

3. **P3 — A failed context-source update overwrites the valid peer**

The V2 bridge invokes `afterMethod` even when the original method throws. For example, Spring LDAP 4.1 rejects `setContextSource(null)` before changing its existing context source, but the interceptor still resolves `null` and overwrites the previously valid peer with `unknown`.

Please record the failure in `handleMethodException` and skip the dynamic-field update from `afterMethod` when the setter failed. Add a regression test confirming that a rejected setter call preserves the previous peer.

P1 is the critical privacy issue. P2 and P3 are localized correctness issues.

weilhuang:
> I found three issues that need addressing:
> 
> 1. **P1 — Do not collect raw LDAP exception details by default**
> 
> `SpringLdapOperationInterceptor.handleMethodException` currently calls `span.log(t)`, which exports the exception message and stack to OAP. Spring LDAP exceptions can contain directory data. For example:
> 
> ```
> event: error
> error.kind: IncorrectResultSizeDataAccessException
> message: base: 'ou=people,dc=example,dc=org'; filter: '(mail=alice@example.org)'.
> stack: ...the same DN and filter...
> ```
> 
> This conflicts with the requirement not to expose LDAP DNs, filters, credentials, or directory data.
> 
> The default behavior should only call `errorOccurred()`, with no message or stack. Exception-detail collection may be provided through an explicit opt-in configuration, such as:
> 
> ```ini
> plugin.spring_ldap.collect_exception_details=false
> ```
> 
> The option must default to `false` and document that exception messages and stacks may contain sensitive LDAP data. Please add tests for both the privacy-safe default and the explicitly enabled behavior.
> 
> 2. **P2 — A successful authentication mapper can be marked as failed**
> 
> The current `Boolean.FALSE` check applies to every operation identified as `authenticate`. However, both `LdapTemplate.authenticate(..., mapper)` and `LdapClient.AuthenticateSpec.execute(mapper)` may successfully authenticate and then legitimately return `Boolean.FALSE` from the mapper.
> 
> Please restrict false-result handling to primitive-boolean status methods:
> 
> ```java
> if (AUTHENTICATE_OPERATION.equals(state.operation)
>     && method.getReturnType() == boolean.class
>     && Boolean.FALSE.equals(ret)) {
>     state.span.errorOccurred();
> }
> ```
> 
> Please test these cases:
> 
> * Primitive `boolean authenticate(...)` returning `false` is an error.
> * A successful `LdapTemplate` mapper returning `Boolean.FALSE` is not an error.
> * A successful `LdapClient` mapper returning `Boolean.FALSE` is not an error.
> 
> 3. **P3 — A failed context-source update overwrites the valid peer**
> 
> The V2 bridge invokes `afterMethod` even when the original method throws. For example, Spring LDAP 4.1 rejects `setContextSource(null)` before changing its existing context source, but the interceptor still resolves `null` and overwrites the previously valid peer with `unknown`.
> 
> Please record the failure in `handleMethodException` and skip the dynamic-field update from `afterMethod` when the setter failed. Add a regression test confirming that a rejected setter call preserves the previous peer.
> 
> P1 is the critical privacy issue. P2 and P3 are localized correctness issues.

Addressed. PTAL.

wu-sheng (approved):
Re-reviewed at 01f822c. All three requested changes are resolved: exception details are suppressed by default and require the documented explicit opt-in; false authentication is treated as failure only for primitive boolean methods, preserving successful mapper results; and rejected context-source setters no longer overwrite the existing peer. The new regression coverage exercises the privacy-safe default and opt-in, both mapper APIs, and setter failure. I also ran the focused module suite locally: 21 tests passed.

PR #825, opened 2026-08-25
fix: restore Log4j2Plugins.dat in apm-toolkit-log4j-2.x

### LINKED ISSUE #14006: [Bug] apm-toolkit-log4j-2.x jar has been missing Log4j2Plugins.dat since 9.5.0, breaking %traceId, %sw_ctx and GRPCLogClientAppender
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar issues.

### Apache SkyWalking Component

Java Agent (apache/skywalking-java)

### What happened

Since **9.5.0**, the released `apm-toolkit-log4j-2.x` jar no longer contains the Log4j2 plugin descriptor:

```
META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat
```

Without that descriptor Log4j2 cannot discover `TraceIdConverter`, so `%traceId` in a `PatternLayout` is greedily matched as the built-in `%t` (thread name) followed by the literal `raceId`:

```
ERROR [main] [mainraceId] Foo: message     <- actual
ERROR [main] [TID: N/A]   Foo: message     <- expected
```

Verified directly against Maven Central:

| Version | `Log4j2Plugins.dat` |
|---------|---------------------|
| 9.0.0 – 9.4.0 | present |
| 9.5.0 – 9.7.0 | **missing** |

Only `apm-toolkit-log4j-2.x` is affected. `apm-toolkit-log4j-1.x` and `apm-toolkit-logback-1.x` never shipped this file, as they do not use the Log4j2 plugin mechanism.

### Root cause

The toolkit module itself did not change:

- `apm-toolkit-log4j-2.x/pom.xml` is identical between 9.4.0 and 9.5.0 apart from the parent `<version>`.
- Every converter/appender class is byte-identical (sha1, first 12 chars):

| class | 9.4.0 | 9.6.0 |
|-------|-------|-------|
| `TraceIdConverter` | `76bf70b1d17c` | `76bf70b1d17c` |
| `SkyWalkingContextConverter` | `c5481fdab754` | `c5481fdab754` |
| `Log4j2OutputAppender` | `79bc1667fbac` | `79bc1667fbac` |
| `Log4j2SkyWalkingContextOutputAppender` | `cbe0b6076cee` | `cbe0b6076cee` |

The change is in the **`java-agent` root pom**. 9.5.0 added an `annotationProcessorPaths` block to `maven-compiler-plugin`:

```xml
<annotationProcessorPaths>
    <path>
        <groupId>org.projectlombok</groupId>
        <artifactId>lombok</artifactId>
        <version>${lombok.version}</version>
    </path>
</annotationProcessorPaths>
```

Once `annotationProcessorPaths` is declared explicitly, javac stops discovering annotation processors from the compile classpath. `apm-toolkit-log4j-2.x` declares `log4j-core` as `provided`, and it is Log4j2's `PluginProcessor` — previously picked up from that classpath — that generates `Log4j2Plugins.dat`. With the processor list narrowed to Lombok only, `PluginProcessor` never runs and the descriptor is silently dropped.

This matches the 9.5.0 changelog entry:

> Centralized plugin version management in the root POM and remove redundant declarations

so this looks like an unintended side effect of a build refactor rather than a deliberate removal.

### What you expected to happen

The released jar should contain `Log4j2Plugins.dat`, so that `%traceId` and `%sw_ctx` resolve automatically as documented, without users having to declare `packages` in `log4j2.xml`.

### How to reproduce

Confirm the descriptor is missing:

```bash
for v in 9.4.0 9.5.0 9.6.0 9.7.0; do
  curl -sfLO "https://repo1.maven.org/maven2/org/apache/skywalking/apm-toolkit-log4j-2.x/$v/apm-toolkit-log4j-2.x-$v.jar"
  printf "%-8s dat=%s\n" "$v" "$(unzip -l apm-toolkit-log4j-2.x-$v.jar | grep -c Log4j2Plugins.dat)"
done
```

Output:

```
9.4.0    dat=1
9.5.0    dat=0
9.6.0    dat=0
9.7.0    dat=0
```

End-to-end, with `apm-toolkit-log4j-2.x:9.6.0` and `log4j-core:2.24.3` on the classpath and this `log4j2.xml`:

```xml
<Configuration status="warn">
  <Appenders>
    <Console name="Console" target="SYSTEM_OUT">
      <PatternLayout pattern="%p [%t] [%traceId] %c{1.}: %m%n"/>
    </Console>
  </Appenders>
  <Loggers>
    <Root level="DEBUG"><AppenderRef ref="Console"/></Root>
  </Loggers>
</Configuration>
```

a single `log.error("MARKER")` prints `ERROR [main] [mainraceId] ...`. Swapping the toolkit to 9.4.0, with no other change, prints `ERROR [main] [TID: N/A] ...`.

### Anything else

**Impact.** Every project on 9.5.0+ that uses `%traceId` or `%sw_ctx` in `log4j2.xml`. The failure is silent — logging keeps working and only the trace id is replaced by a thread name, so it is easy to ship without noticing. This has been broken across three releases (9.5.0, 9.6.0, 9.7.0).

**Workarounds**, both verified locally:

1. Declare the package explicitly. Note `packages` is deprecated in Log4j2 and removed in Log4j 3:
   ```xml
   <Configuration packages="org.apache.skywalking.apm.toolkit.log.log4j.v2.x">
   ```
2. Pin the toolkit back to 9.4.0. Since all classes are byte-identical, this is functionally equivalent to 9.6.0 plus the missing descriptor.

**Suggested fix.** Add `log4j-core` to the processor paths for this module. Because `annotationProcessorPaths` replaces rather than appends when overriding the parent, the Lombok entry has to be repeated:

```xml
<!-- apm-toolkit-log4j-2.x/pom.xml -->
<build>
  <plugins>
    <plugin>
      <artifactId>maven-compiler-plugin</artifactId>
      <configuration>
        <annotationProcessorPaths>
          <path>
            <groupId>org.projectlombok</groupId>
            <artifactId>lombok</artifactId>
            <version>${lombok.version}</version>
          </path>
          <path>
            <groupId>org.apache.logging.log4j</groupId>
            <artifactId>log4j-core</artifactId>
            <version>${log4j-core.version}</version>
          </path>
        </annotationProcessorPaths>
      </configuration>
    </plugin>
  </plugins>
</build>
```

It would also be worth adding a build-time check that the descriptor exists in the packaged jar, so a future refactor cannot drop it silently again.

**Disclosure.** The root-cause analysis above was done with the help of an AI assistant (Claude). Every claim in it — the version boundary, the class checksums, the pom diff, and both workarounds — was verified against artifacts published on Maven Central and can be reproduced with the commands above.

### Are you willing to submit a pull request to fix on your own?

- [ ] Yes I am willing to submit a pull request on my own!

### Code of Conduct

waterWang:
### Problem

Since **9.5.0**, the released `apm-toolkit-log4j-2.x` jar no longer contains the Log4j2 plugin descriptor:

```
META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat
```

Without that descriptor Log4j2 cannot discover `TraceIdConverter`, so `%traceId` and `%sw_ctx` in a `PatternLayout` silently break — they are greedily matched as the built-in `%t` (thread name) followed by literal text.

Verified against Maven Central:

| Version | `Log4j2Plugins.dat` |
|---------|---------------------|
| 9.0.0 – 9.4.0 | present |
| 9.5.0 – 9.7.0 | **missing** |

### Root Cause

9.5.0 added an `annotationProcessorPaths` block to `maven-compiler-plugin` in the root `pom.xml`, listing only Lombok. Once `annotationProcessorPaths` is declared explicitly, javac stops discovering annotation processors from the compile classpath. The `log4j-core` jar is declared as `provided` in this module, and Log4j2's `PluginProcessor` — previously picked up from that classpath — is what generates `Log4j2Plugins.dat`. With the processor list narrowed to Lombok only, `PluginProcessor` never runs and the descriptor is silently dropped.

### Fix

Declare the compiler plugin in this module with `annotationProcessorPaths` containing both Lombok (preserving parent behaviour) and `log4j-core` (restoring the `PluginProcessor`).

### Verification

The module's Java source files (`TraceIdConverter`, `SkyWalkingContextConverter`, `Log4j2OutputAppender`, `Log4j2SkyWalkingContextOutputAppender`) do not use Lombok annotations, so adding it to the path is safe but optional.

### References

- Reported upstream: apache/skywalking#14006

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Release Notes.
 ------------------
 
 * Add Spring LDAP 3.3.x-4.x plugin.
+* Fix the Log4j2 plugin descriptor (`Log4j2Plugins.dat`) missing from the `apm-toolkit-log4j-2.x` jar since 9.5.0, which broke `%traceId` and `%sw_ctx` resolution in Log4j2 `PatternLayout` (apache/skywalking#14006).
 
 All issues and pull requests are [here](https://github.com/apache/skywalking/milestone/263?closed=1)
 
--- a/apm-application-toolkit/apm-toolkit-log4j-2.x/pom.xml
+++ b/apm-application-toolkit/apm-toolkit-log4j-2.x/pom.xml
@@ -39,4 +39,34 @@
             <scope>provided</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <!--
+                      When annotationProcessorPaths is declared explicitly,
+                      javac stops discovering processors from the classpath.
+                      log4j-core's PluginProcessor (which generates
+                      Log4j2Plugins.dat) must be listed here explicitly,
+                      otherwise %traceId / %sw_ctx resolve to built-in patterns.
+                    -->
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.projectlombok</groupId>
+                            <artifactId>lombok</artifactId>
+                            <version>${lombok.version}</version>
+                        </path>
+                        <path>
+                            <groupId>org.apache.logging.log4j</groupId>
+                            <artifactId>log4j-core</artifactId>
+                            <version>${log4j-core.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>
--- a/apm-application-toolkit/apm-toolkit-log4j-2.x/pom.xml
+++ b/apm-application-toolkit/apm-toolkit-log4j-2.x/pom.xml
@@ -47,18 +47,15 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
                     <!--
-                      When annotationProcessorPaths is declared explicitly,
-                      javac stops discovering processors from the classpath.
-                      log4j-core's PluginProcessor (which generates
-                      Log4j2Plugins.dat) must be listed here explicitly,
-                      otherwise %traceId / %sw_ctx resolve to built-in patterns.
+                      The root POM declares annotationProcessorPaths, which makes
+                      javac stop discovering processors from the classpath.
+                      Append log4j-core's PluginProcessor (which generates
+                      Log4j2Plugins.dat) so that %traceId / %sw_ctx resolve to the
+                      toolkit converters instead of built-in Log4j2 patterns.
+                      combine.children="append" keeps the Lombok entry inherited
+                      from the root POM.
                     -->
-                    <annotationProcessorPaths>
-                        <path>
-                            <groupId>org.projectlombok</groupId>
-                            <artifactId>lombok</artifactId>
-                            <version>${lombok.version}</version>
-                        </path>
+                    <annotationProcessorPaths combine.children="append">
                         <path>
                             <groupId>org.apache.logging.log4j</groupId>
                             <artifactId>log4j-core</artifactId>


wu-sheng:
Could you update the changes.md about this?

wu-sheng:
The fix is in the right place — module POM rather than root. Two notes on the form.

### 1. Prefer `combine.children="append"` over repeating the Lombok entry

`annotationProcessorPaths` inherits with **replace** semantics, so a child list wins entirely — which is exactly why Lombok has to be repeated here. Maven's `combine.children="append"` avoids the duplication: the module declares only what it adds, and the parent's entries are inherited.

```xml
<annotationProcessorPaths combine.children="append">
    <path>
        <groupId>org.apache.logging.log4j</groupId>
        <artifactId>log4j-core</artifactId>
        <version>${log4j-core.version}</version>
    </path>
</annotationProcessorPaths>
```

Verified locally — the resulting javac invocation carries both entries, and the descriptor comes back:

```
-processorpath lombok-1.18.42.jar     <- inherited from the root POM
               log4j-core-2.7.jar     <- declared here
→ META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat (380 bytes)
```

Its content is identical to the descriptor shipped in 9.4.0 and earlier: `TraceIdConverter`, `SkyWalkingContextConverter`, `GRPCLogClientAppender`.

The reason to prefer this is not brevity — it is that the replace form reproduces this very bug one level up. If the root POM ever gains another global processor, every module that declared its own `annotationProcessorPaths` would silently stop receiving it. That is the same silent failure mode being fixed here.

### 2. On "safe but optional"

> The module's Java source files ... do not use Lombok annotations, so adding it to the path is safe but optional.

Accurate today, but the wording invites someone to drop that entry later, and the resulting failure would be precisely the build error #767 fixed — `cannot find symbol` on a Lombok-generated accessor, with nothing in the diff pointing at the cause. With `combine.children="append"` the question does not arise at all.

### Either way, a guard is still needed

Neither form is self-defending: if the configuration is dropped, reordered, or overridden, the descriptor silently disappears again — no warning, no build failure, just a jar that is missing 380 bytes. Worth adding the build-time assertion the issue suggests, so the next occurrence fails the build instead of reaching three releases. I can send that as a follow-up.

waterWang:
Thanks for the review @wu-sheng. I've updated the PR per both points:

1. **CHANGES.md** now includes an entry for this fix (referencing apache/skywalking#14006).
2. **pom.xml** now uses `combine.children="append"` on `annotationProcessorPaths`, so the module declares only the log4j-core processor it adds while keeping the Lombok entry inherited from the root POM — no duplicated entry, and no risk that a future global processor silently stops being applied.

I also verified the mechanism locally: with only Lombok in the processor path, `Log4j2Plugins.dat` is not generated; appending log4j-core (which carries `org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor`) brings the descriptor back, and its content contains `TraceIdConverter`, `SkyWalkingContextConverter` and the appenders — identical to the 9.4.0 release.

On the build-time guard you mentioned: happy to add it as a follow-up if you'd like, so any future regression fails the build instead of shipping silently.

wu-sheng (approved):

